PR #6, opened 2017-02-09
add connection mode for udp sockets (work in progress)

little-dude:
Hi,

This PR adds a connection mode for UDP sockets, which correspond to [std::net::UdpSocket::connect()](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.connect).

This behavior is also described in [2008 POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/connect.html#top).

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -95,7 +95,7 @@ fn main() {
         // udp:6969: respond "yo dawg"
         {
             let socket: &mut UdpSocket = sockets.get_mut(udp_handle).as_socket();
-            if socket.endpoint().is_unspecified() {
+            if socket.local_endpoint().is_unspecified() {
                 socket.bind(6969)
             }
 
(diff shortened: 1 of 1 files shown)


little-dude:
This needs tests obviously, but I'm opening the PR to get early feedback.

whitequark:
> This PR adds a connection mode for UDP sockets, which correspond to std::net::UdpSocket::connect().
> 
> This behavior is also described in 2008 POSIX standard.

Note that smoltcp, by design, makes no effort to conform to POSIX or behave like std::net. In particular the TCP sockets are nothing like TcpListener/TcpStream; that is the job of a higher-level abstraction to present them like that. They are rather like TCBs (transmission control blocks) from RFC 793.

Anyway. I am partial to the idea behind this socket, but I strongly dislike the added branching in the API. Not sure what to make of it.

little-dude:
I see, thanks for the explanation. I think in this specific case, I can indeed provide this abstraction at a higher level (for the context I'm trying to use smoltcp as the network stack for redox).

whitequark:
@little-dude It's virtually certain that you will need to provide a higher-level abstraction, at least for TCP and to account for blocking. Therefore it seems alright to also do that for UDP sockets.

By the way, I am rather happy about the use of smoltcp in redox--the existing redox network stack wasn't exactly inspiring...

PR #15, opened 2017-06-14
Factor out the "raw_socket" and "tap_interface" features

batonius:
**Problem**: I'm investigating the possibility of using smoltcp in Redox (yeah, I know, I'm not the first one). The unusual thing about Redox is that while it has a version of Rust's `libstd`, it doesn't have libc and the `libc` crate doesn't support it, so while it's possible to use `std::Box` and `std::Vect`, raw sockets are unavailable. As a result, it's impossible to build smoltcp for Redox with the `std` feature on, which leaves many features disabled for no reason.

**Solution**: The PR factors out the `raw_socket` and `tap_interfaces` features, making the `std` feature not dependent on the `libc` crate. `raw_socket` and `tap_interfaces` are still default features. With this PR it's possible to build smoltcp for Redox with `--no-default-features --features="collections alloc std"`. 

**Other**: I was going to create a long-term fork, but turns out the required changes are minimal, so if it's ok with you, it would be great to depend directly on the original.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,13 @@ language: rust
 matrix:
   include:
     - rust: stable
-      env: FEATURES='std' MODE='test'
+      env: FEATURES='std raw_socket tap_interface' MODE='test'
     - rust: beta
-      env: FEATURES='std' MODE='test'
+      env: FEATURES='std raw_socket tap_interface' MODE='test'
     - rust: nightly
-      env: FEATURES='std' MODE='test'
+      env: FEATURES='std raw_socket tap_interface' MODE='test'
     - rust: nightly
-      env: FEATURES='std log' MODE='test'
+      env: FEATURES='std raw_socket tap_interface log' MODE='test'
     - rust: nightly
       env: FEATURES='alloc collections' MODE='build'
     - rust: nightly


whitequark:
The Travis build has failed.

batonius:
Fixed.

whitequark:
> I was going to create a long-term fork, but turns out the required changes are minimal, so if it's ok with you, it would be great to depend directly on the original.

Oh and any chances needed for redox (e.g. a DHCP packet parser/serializer, as discussed in the other PR) are rather welcome in smoltcp. I would prefer to not have redox `#[cfg]`s because it indicates some unsolved underlying problem but otherwise it's fine.

batonius:
Good to know. 

By the way, how do you feel about raw sockets? We need them for `ping`, obviously, but I was also thinking about using them to migrate the existing network stack piecemeal, i.e. to provide level 3 abstraction to the existing `tcpd`, `udpd` and `icmpd` daemons at first. This means raw sockets should have a priority over the build-in handlers, which looks like the standard behavior, http://man7.org/linux/man-pages/man7/raw.7.html :
> When a packet is received, it is passed to any raw sockets which have been bound to its protocol before it is passed to other protocol handlers (e.g., kernel protocol modules).

It looks like a pretty straightforward thing to do, being similar to UDP. So, if I add them, will you consider merging?

whitequark:
Naturally.

whitequark:
Oh regarding `ping`, I would rather like to have an ICMP socket option, the equivalent of `socket(PF_INET, SOCK_DGRAM, IPPROTO_ICMP);` on Linux.

batonius:
Interesting, I didn't know SOCK_DGRAM+IPPROTO_ICMP is a thing, I guess I'll look into adding ICMP sockets as well.

PR #18, opened 2017-06-16
Add raw sockets

batonius:
**Problem**: smoltcp doesn't support raw sockets, low-level utilities like `ping` and custom IP protocols are impossible to implement.

**Solution**: implement raw sockets, provide an example of their usage.

- `socket::udp::SocketBuffer` has been factored out as generic `common::RingBuffer`.
- `socket::raw` module has been implemented.
- `iface::EthernetInterface` has been changed to support raw sockets, giving them priority over the built-in handlers.
- The `ping` example has been implemented.

**Status**: the `ping` example works as expected.

**TODOs**:
- Currently, smoltcp uses iteration over the list of sockets for packet dispatch, which is highly inefficient on the platforms where standard containers like maps and hash tables are available. The existing `SocketSet` container should be complimented with lookup tables for raw and tcp/udp sockets.

**Other**:
- I'm using the current version of rustfmt for my code, sometimes it looks weird.
- I've used some recent stable Rust features like field shorthands and pub(crate), so I've bumped Rust version in Readme.md to 1.18.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ include complicated compile-time computations, such as macro or type tricks, eve
 at cost of performance degradation.
 
 _smoltcp_ does not need heap allocation *at all*, is [extensively documented][docs],
-and compiles on stable Rust 1.15 and later.
+and compiles on stable Rust 1.18 and later.
 
 [docs]: https://docs.rs/smoltcp/
 
@@ -213,6 +213,23 @@ cargo run --example client -- tap0 ADDRESS PORT
 
 It connects to the given address (not a hostname) and port (e.g. `socat stdio tcp4-listen 1234`),
 and will respond with reversed chunks of the input indefinitely.
+
+### examples/ping.rs
+
+_examples/ping.rs_ implements a minimal version of the `ping` utility using raw sockets.
+
+The host is assigned the hardware address `02-00-00-00-00-02` and IPv4 address `192.168.69.2`.
+
+Read its [source code](/examples/ping.rs), then run it as:
+
+```sh
+cargo run --example ping -- tap0 ADDRESS
+```
+
+It sends a series of 4 ICMP ECHO\_REQUEST packets to the given address with interval 1s and
+prints out a status line on each valid ECHO\_RESPONSE received. The first ECHO\_REQUEST packet is expected
+to be lost since arp\_cache is empty at the moment and _smoltcp_ sends an ARP request instead. The best
+target address to test would be the address of the `tap0` interface itself.
 
 License
 -------
--- a/examples/ping.rs
+++ b/examples/ping.rs
@@ -1,0 +1,156 @@
+#[macro_use]
+extern crate log;
+extern crate env_logger;
+extern crate getopts;
+extern crate smoltcp;
+extern crate byteorder;
+
+mod utils;
+
+use std::str::{self, FromStr};
+use std::time::{Duration, Instant};
+use smoltcp::Error;
+use smoltcp::wire::{EthernetAddress, IpAddress, IpProtocol, Icmpv4Repr, Icmpv4Packet, Ipv4Repr,
+                    Ipv4Packet, Ipv4Address};
+use smoltcp::iface::{ArpCache, SliceArpCache, EthernetInterface};
+use smoltcp::socket::{AsSocket, SocketSet};
+use smoltcp::socket::{RawSocket, RawSocketBuffer, RawPacketBuffer, RawPayloadType};
+use std::collections::HashMap;
+use byteorder::{ByteOrder, NetworkEndian};
+
+const PING_INTERVAL_S: u64 = 1;
+const PING_TIMEOUT_S: u64 = 5;
+const PINGS_TO_SEND: usize = 4;
+
+fn main() {
+    utils::setup_logging();
+
+    let (device, args) = utils::setup_device(&["ADDRESS"]);
+    let remote_addr = Ipv4Address::from_str(&args[0]).unwrap();
+    let local_addr = Ipv4Address::new(192, 168, 69, 2);
+
+    let raw_rx_buffer = RawSocketBuffer::new(vec![RawPacketBuffer::new(vec![0; 256])]);
+    let raw_tx_buffer = RawSocketBuffer::new(vec![RawPacketBuffer::new(vec![0; 256])]);
+    let raw_socket = RawSocket::new(
+        raw_rx_buffer,
+        raw_tx_buffer,
+        IpProtocol::Icmp,
+        RawPayloadType::Ipv4,
+    );
+
+    let hardware_addr = EthernetAddress([0x02, 0x00, 0x00, 0x00, 0x00, 0x02]);
+    let arp_cache = SliceArpCache::new(vec![Default::default(); 8]);
+    let mut iface = EthernetInterface::new(
+        Box::new(device),
+        Box::new(arp_cache) as Box<ArpCache>,
+        hardware_addr,
+        [IpAddress::Ipv4(local_addr)],
+    );
+
+    let mut sockets = SocketSet::new(vec![]);
+    let raw_handle = sockets.add(raw_socket);
+
+    let mut send_next = Duration::default();
+    let mut seq = 0;
+    let mut received = 0;
+    let mut echo_payload = [0xffu8; 40];
+    let mut waiting_queue = HashMap::new();
+
+    let startup_time = Instant::now();
+
+    loop {
+        {
+            let socket: &mut RawSocket = sockets.get_mut(raw_handle).as_socket();
+
+            let timestamp = Instant::now().duration_since(startup_time);
+            let timestamp_us = (timestamp.as_secs() * 1000000) +
+                (timestamp.subsec_nanos() / 1000) as u64;
+
+            if seq == PINGS_TO_SEND && waiting_queue.is_empty() {
+                break;
+            }
+
+            if socket.can_send() && seq < PINGS_TO_SEND && send_next <= timestamp {
+                NetworkEndian::write_u64(&mut echo_payload, timestamp_us);
+                let icmp_repr = Icmpv4Repr::EchoRequest {
+                    ident: 1,
+                    seq_no: seq as u16,
+                    data: &echo_payload,
+                };
+                let ipv4_repr = Ipv4Repr {
+                    src_addr: Ipv4Address::new(0, 0, 0, 0),
+                    dst_addr: remote_addr,
+                    protocol: IpProtocol::Icmp,
+                    payload_len: icmp_repr.buffer_len(),
+                };
+
+                let raw_payload = socket
+                    .send(ipv4_repr.buffer_len() + icmp_repr.buffer_len())
+                    .unwrap();
+
+                let mut ipv4_packet = Ipv4Packet::new(raw_payload).unwrap();
+                ipv4_repr.emit(&mut ipv4_packet);
+                let mut icmp_packet = Icmpv4Packet::new(ipv4_packet.payload_mut()).unwrap();
+                icmp_repr.emit(&mut icmp_packet);
+
+                waiting_queue.insert(seq, timestamp);
+                seq += 1;
+                send_next += Duration::new(PING_INTERVAL_S, 0);
+            }
+
+            if socket.can_recv() {
+                let payload = socket.recv().unwrap();
+                let ipv4_packet = Ipv4Packet::new(payload).unwrap();
+                let ipv4_repr = Ipv4Repr::parse(&ipv4_packet).unwrap();
+
+                if ipv4_repr.src_addr == remote_addr && ipv4_repr.dst_addr == local_addr {
+                    let packet = Icmpv4Packet::new(ipv4_packet.payload()).unwrap();
+                    if let Ok(Icmpv4Repr::EchoReply { seq_no, data, .. }) =
+                        Icmpv4Repr::parse(&packet)
+                    {
+                        let seq_no = seq_no as usize;
+                        if let Some(_) = waiting_queue.get(&seq_no) {
+                            let packet_timestamp_us = NetworkEndian::read_u64(data);
+                            println!(
+                                "{} bytes from {}: icmp_seq={}, time={}ms",
+                                data.len(),
+                                remote_addr,
+                                seq_no,
+                                (timestamp_us - packet_timestamp_us) as f32 / 1000.0f32
+                            );
+                            waiting_queue.remove(&seq_no);
+                            received += 1;
+                        }
+                    }
+                }
+            }
+
+            if !waiting_queue.is_empty() {
+                waiting_queue.retain(|seq, from| if (timestamp - *from).as_secs() <
+                    PING_TIMEOUT_S
+                {
+                    true
+                } else {
+                    println!("From {} icmp_seq={} timeout", remote_addr, seq);
+                    false
+                })
+            }
+        }
+
+        let timestamp = Instant::now().duration_since(startup_time);
+        let timestamp_ms = (timestamp.as_secs() * 1000) +
+            (timestamp.subsec_nanos() / 1000000) as u64;
+        match iface.poll(&mut sockets, timestamp_ms) {
+            Ok(()) | Err(Error::Exhausted) => (),
+            Err(e) => debug!("poll error: {}", e),
+        }
+    }
+
+    println!("--- {} ping statistics ---", remote_addr);
+    println!(
+        "{} packets transmitted, {} received, {}% packet loss",
+        seq,
+        received,
+        100 * (seq - received) / seq
+    );
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,3 +39,6 @@ name = "server"
 
 [[example]]
 name = "client"
+
+[[example]]
+name = "ping"


batonius:
Looks like the recent nightly broke the `collections` feature, master fails to build with it as well.

whitequark:
> Looks like the recent nightly broke the collections feature, master fails to build with it as well.

This... is actually really troublesome.

whitequark:
> I've used some recent stable Rust features like field shorthands and pub(crate), so I've bumped Rust version in Readme.md to 1.18.

This is 100% OK, I was going to start using those myself.

> I'm using the current version of rustfmt for my code, sometimes it looks weird.

I dislike rustfmt and hand-format all of my code; so please don't mind if I ask to fix some particularly egregious cases.

src/common/ring_buffer.rs
```diff
@@ -0,0 +1,118 @@
...
+pub trait Resetable {
+    fn reset(&mut self);
+}
+
+/// A ring buffer.
+#[derive(Debug)]
+pub struct RingBuffer<'a, T: 'a> {
+    storage: Managed<'a, [T]>,
+    read_at: usize,
+    length: usize,
```
whitequark: I would really like it if you at least kept the formatting as-is when moving code around. My preferences notwithstanding this makes git's copy/move tracking mechanisms work better.

src/iface/ethernet.rs
```diff
@@ -179,6 +179,13 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
                                         &eth_frame.src_addr());
                 }
 
+                for raw_socket in sockets.iter_mut().filter_map(Socket::as_raw) {
+                    if let Ok(()) = raw_socket.process(timestamp, &IpRepr::Ipv4(ipv4_repr),
+                                                       ipv4_packet.payload()) {
+                        return Ok(());
```
whitequark: That's not how raw sockets work on hosted systems. To quote Linux man pages:

> When a packet is received, it is passed to any raw sockets which have been bound to its protocol before it is passed to other protocol handlers (e.g., kernel protocol modules).

Note that it does *not* say that the packet is *not* passed to other protocol handlers. Indeed, that's why TCP sockets still work even if you're running tcpdump (or ping) concurrently.

src/socket/raw.rs
```diff
@@ -0,0 +1,272 @@
...
+/// A raw IP packet ring buffer.
+pub type SocketBuffer<'a, 'b: 'a> = RingBuffer<'a, PacketBuffer<'b>>;
+
+/// A Raw IP socket
+///
+/// A raw socket is bound to a specifict IP protocol and owns
+/// transmit and receive packet buffers.
+#[derive(Debug)]
+pub struct RawSocket<'a, 'b: 'a> {
+    ip_addr: IpAddress,
```
whitequark: A raw socket isn't bound to an IP address, only to a protocol.

batonius:
Done.

src/socket/raw.rs
```diff
@@ -0,0 +1,272 @@
...
+            self.ip_proto,
+            packet_buf.size
+        );
+        Ok(&mut packet_buf.as_mut()[..size])
+    }
+
+    /// Enqueue a packet to be sent to the given IP address, and fill it from a slice.
+    ///
+    /// See also [send](#method.send).
+    pub fn send_slice(&mut self, data: &[u8], ip_addr: IpAddress) -> Result<usize, ()> {
```
whitequark: There's no need to accept IpAddress in send, or provide it in recv; it is the responsibility of the client to dissect the packet.

examples/ping.rs
```diff
@@ -0,0 +1,139 @@
...
+    let mut send_next = Duration::default();
+    let mut seq = 0;
+    let mut received = 0;
+    let mut echo_payload = [0xffu8; 40];
+    let mut waiting_queue = HashMap::new();
+
+    let startup_time = Instant::now();
+
+    loop {
+        {
```
whitequark: This scope isn't necessary since ping only needs one socket.

examples/ping.rs
```diff
@@ -0,0 +1,139 @@
...
+                let repr = Icmpv4Repr::EchoRequest {
+                    ident: 1,
+                    seq_no: seq as u16,
+                    data: &echo_payload,
+                };
+                let raw_payload = socket.send(repr.buffer_len(), remote_addr).expect(
+                    "Can't allocate packet",
+                );
+                let mut packet =
+                    Icmpv4Packet::new(raw_payload).expect("Can't create ICMPv4 packet");
```
whitequark: Let's just `unwrap()`. If we're going to add more proper error handling to examples it should be done in sync (and use `error!`, not `expect()`).

examples/ping.rs
```diff
@@ -0,0 +1,139 @@
...
+
+            if socket.can_recv() {
+                let (payload, remote_ip) = socket.recv().expect("Can't receive ICMPv4 packet");
+                if remote_ip == remote_addr {
+                    let packet = Icmpv4Packet::new(payload).expect("Can't parse ICMPv4 packet");
+                    if let Ok(Icmpv4Repr::EchoReply { seq_no, data, .. }) =
+                        Icmpv4Repr::parse(&packet)
+                    {
+                        let seq_no = seq_no as usize;
+                        if let Some(_) = waiting_queue.get(&seq_no) {
```
whitequark: You can *drastically* simplify this by embedding the timestamp in the sent packet :) As a bonus the ping program will be possible to run with no_std code after this change.

batonius: I already do embed the timestamp, waiting_queue is used for timeout calculation.

examples/ping.rs
batonius: It's necessary because both `sockets.get_mut` and `iface.poll` take a mutable reference to `sockets`. It's the same as in the `client.rs` example.

examples/ping.rs
whitequark: Ah, I missed that, sorry. Hm... should be fine the way it is, then.

examples/ping.rs
whitequark: Oh, of course.

batonius:
I'd like to keep `IpAddress` in `RawSocket` for two reasons: first, since there's no routing yet, we can't determine source IP from destination IP alone, so we need a default source IP. Second, it works as an IPv4/IPv6 marker.

whitequark:
> first, since there's no routing yet, we can't determine source IP from destination IP alone

Take a closer look at EthernetInterface. It already selects a source IP as "first IP that matches destination IP protocol" in a few code paths.

> it works as an IPv4/IPv6 marker

Why is a marker necessary? The first nibble of the packet has that already.

batonius:
>It already selects a source IP as "first IP that matches destination IP protocol" in a few code paths.
Right, I see it now.
>Why is a marker necessary? 
A raw socket is bound to a version of IP on creation, by using either AF_INET or AF_INET6 as the `socket_family` argument of `socket`, this means an IPv4 raw socket should ignore IPv6 packets, and vice versa.

whitequark:
> A raw socket is bound to a version of IP on creation, by using either AF_INET or AF_INET6 as the socket_family argument of socket, this means an IPv4 raw socket should ignore IPv6 packets, and vice versa.

Let's add a `smoltcp::wire::IpVersion` enum for this and maybe other cases.

whitequark:
> Let's add a smoltcp::wire::IpVersion enum for this and maybe other cases.

Actually, hang on. It might be not so wise to tie it to IP. Raw sockets that receive, for example, Ethernet frames, 802.11 frames, 802.15.4 frames, are all still useful. Let's make a enum specifically listing what a raw socket can receive, instead.

README.md
```diff
@@ -214,6 +214,23 @@ cargo run --example client -- tap0 ADDRESS PORT
 It connects to the given address (not a hostname) and port (e.g. `socat stdio tcp4-listen 1234`),
 and will respond with reversed chunks of the input indefinitely.
 
+### examples/ping.rs
+
+_examples/ping.rs_ implements a minimail version of the `ping` utility using raw sockets.
```
jordens: minimal
(just passing by)

batonius:
Wait. `IpRepr::lower` only replaces src_ip for `IpRepr::Unspecified`, and I'm calling `emit` in `RawSocket::dispatch` with `IpRepr::Ipv4` I've just parsed it from the packet I've got from the tx buffer. So `EthernetInterface` doesn't replace src_ip in this case.

whitequark:
> Wait. IpRepr::lower only replaces src_ip for IpRepr::Unspecified, and I'm calling emit in RawSocket::dispatch with IpRepr::Ipv4 I've just parsed from the packet I've gotten from the tx buffer. So EthernetInterface doesn't replace src_ip in this case.

Of course, IP addresses [also have unspecified ranges](https://docs.rs/smoltcp/0.3.0/smoltcp/wire/struct.Ipv4Address.html#method.is_unspecified), so we can make EthernetInterface look at that too.

batonius:
So, I make an Unspecified repr from Ipv4Repr, turning all-zero src_addr into an Unspecified addr. It seems to work with all-zero src_addr in `ping`.

whitequark:
> So, I make an Unspecified repr from Ipv4Repr, turning all-zero src_addr into an Unspecified addr. It seems to work with all-zero src_addr in ping.

I think it's a better idea to change the condition in EthernetInterface to check for unspecified source address in any kind of IpRepr (via a query method on the latter); this will make any kind of sockets, such as ICMP sockets, work the same.

batonius:
Why is `Ipv4Addr::UNSPECIFIED` commented out? It looks like a useful constant to have.

whitequark:
> Why is Ipv4Addr::UNSPECIFIED commented out? It looks like a useful constant to have.

I'd like to have it as an associated constant but those aren't in stable yet.

whitequark:
They'll be stable soon, https://github.com/rust-lang/rust/issues/29646.

batonius:
I'm not sure if I should remove the address replacement logic from `IpRepr::lower`, it doesn't look like the right place for it, but it's being called from tcp.rs and I have no desire to touch it yet :)

whitequark:
Your `specify_src_addr` should be just a part of `lower`.

batonius:
Done. I'm just not sure if the src address replacement should be a part of lowering IpRepr to concrete representation, it looks like two separate actions to me. But if `specify_src_addr` isn't needed anywhere else, then it might make sense.

whitequark:
> I'm just not sure if the src address replacement should be a part of lowering IpRepr to concrete representation

Address replacement is *the whole point* of IpRepr::lower, the exact representation details are irrelevant.

And I'm not happy with your fix. There shouldn't be a `new_unspecified`, we'll have the associated const in 1.20. And splitting `lower` into two functions, one of which is a mutator and the other is not, is not helping with clarity here.

whitequark:
Thanks! Any chance you could update (or add?) a test for `IpRepr::lower`?

src/iface/ethernet.rs
```diff
@@ -179,6 +179,11 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
                                         &eth_frame.src_addr());
                 }
 
+                for raw_socket in sockets.iter_mut().filter_map(Socket::as_raw) {
+                    let _ = raw_socket.process(timestamp, &IpRepr::Ipv4(ipv4_repr),
```
whitequark: I'm worried about this `let _ =`. What errors can happen here besides `Exhausted`?

src/socket/mod.rs
```diff
@@ -98,6 +106,13 @@ impl<'a, 'b> Socket<'a, 'b> {
             where F: FnMut(&IpRepr, &IpPayload) -> Result<R, Error> {
         dispatch_socket!(self, |socket [mut]| socket.dispatch(timestamp, limits, emit))
     }
+
+    pub(crate) fn as_raw(&mut self) -> Option<&mut RawSocket<'a, 'b>> {
+        match *self {
+            Socket::Raw(ref mut raw) => Some(raw),
+            _ => None,
+        }
+    }
```
whitequark: I don't think single-use helper functions are necessary. Just use `as_socket`...

whitequark: Oh, this is a different issue, `as_socket` doesn't let us interrogate. Sure. Let's add `try_as_socket` in addition to `as_socket` to the `AsSocket` trait, that would be much cleaner.

src/socket/raw.rs
```diff
@@ -0,0 +1,241 @@
...
+pub type SocketBuffer<'a, 'b: 'a> = RingBuffer<'a, PacketBuffer<'b>>;
+
+/// A Raw IP socket
+///
+/// A raw socket is bound to a specifict IP protocol and owns
+/// transmit and receive packet buffers.
+#[derive(Debug)]
+pub struct RawSocket<'a, 'b: 'a> {
+    payload_type: PayloadType,
+    ip_proto: IpProtocol,
```
whitequark: There is no need to have separate `payload_type` and `ip_proto` fields because the latter is strictly derived from the former.

src/socket/raw.rs
```diff
@@ -0,0 +1,241 @@
...
+        where F: FnMut(&IpRepr, &IpPayload) -> Result<R, Error>,
+    {
+        let mut packet_buf = self.tx_buffer.dequeue_mut().map_err(|()| Error::Exhausted)?;
+        net_trace!("[{}] proto {} sending {} octets",
+                   self.debug_id, self.ip_proto, packet_buf.size);
+        match self.payload_type {
+            PayloadType::Ipv4 => {
+                let mut ipv4_packet = Ipv4Packet::new(packet_buf.as_mut())?;
+                ipv4_packet.fill_checksum();
+                let ipv4_packet = Ipv4Packet::new(&*ipv4_packet.into_inner())?;
```
whitequark: Why the reinitialization?

src/storage/ring_buffer.rs
```diff
@@ -0,0 +1,134 @@
+use managed::Managed;
+
+/// A trait for setting a value to the default state.
+///
+/// In-palace analog of Default.
+/// Used by RingBuffer for initializing a storage.
+pub trait Resetable {
```
whitequark: `Resettable`, and let's move it to `smoltcp::storage` since it is not really specific to ring buffers.

src/storage/ring_buffer.rs
```diff
@@ -0,0 +1,134 @@
+use managed::Managed;
+
+/// A trait for setting a value to the default state.
+///
+/// In-palace analog of Default.
```
whitequark: `In-place`

whitequark:
Okay, this is much better. After the last round of review I don't think there will be any further problems.

src/iface/ethernet.rs
batonius: Error::Rejected, and it definitely should be ignored.

src/socket/raw.rs
batonius: Why? `payload_type` could be Ipv4 or Ipv6 and ip_proto is independent of that.

src/iface/ethernet.rs
whitequark: Ah right, so only that. Then can you add a match that ignores this error and panics on any other? That will be robust in light of code evolution (or help when fuzzing).

src/socket/raw.rs
batonius: `Ipv4Repr::parse` requires non-mutable `Ipv4Packet`.

whitequark: `&mut T` to `&T` is a valid cast, no?

src/socket/raw.rs
whitequark: Oh, of course.

Okay, just rename it to `ip_protocol`.

src/iface/ethernet.rs
batonius: You mean `return Err(e)`, not panic? Becuase you do `return Err(e)` for socket errors.

src/socket/raw.rs
batonius: That what I thought, but it doesn't work for me for some reason:
```
   --> socket/raw.rs:185:49
    |
185 |                 let ipv4_repr = Ipv4Repr::parse(&ipv4_packet)?;
    |                                                 ^^^^^^^^^^^^ types differ in mutability
    |
    = note: expected type `&wire::ipv4::Packet<&_>`
               found type `&wire::ipv4::Packet<&mut [u8]>`
    = help: here are some functions which might fulfill your needs:
            - .into_inner()
```

src/iface/ethernet.rs
whitequark: Panics, through `unreachable!()`, since those branches should be.

src/socket/raw.rs
whitequark: Ohhh right, that's an artifact of how my `wire` impls are done, nevermind.

batonius:
I believe that's all with the change requests.

whitequark:
Wonderful, thanks! Do you think you can rebase the commits?

batonius:
Something like this?

whitequark:
Yup! I'll merge tomorrow, I think.

batonius:
Good. I have some ideas about packet dispatching but I think this time I'll open an issue to discuss it first.

whitequark:
Sure, go ahead!

whitequark:
Merged, with stylistic, documentation, and formatting changes. No functional changes.

Most notably I've replaced `socket::RawPayloadType` with a `wire::IpVersion` since our raw sockets are specifically **IP** raw sockets.

whitequark:
@batonius Incidentally--did you decide to do anything about ICMP sockets? I think (out of stuff that redox supports right now) ICMP sockets and DHCP support are definitely in scope of smoltcp, and DNS support almost certainly is. At the very least the `wire::dhcp` and `wire::dns` modules would be useful; DNS is stateless so there is perhaps not much more to be done, and a DHCP state machine should certainly live inside smoltcp, not outside.

batonius:
My current goal is to replace the two low-level network daemons, ethernetd and ipd, with a unified one based on smoltcp, while keeping the interface ipd provides to tcpd, udpd and icmpd, which is basically raw sockets. Eventually, I'd like to have ICMP sockets, DHCP and DNS all integrated into that unified daemon, but only after I integrate udpd and tcpd into it.

whitequark:
Sounds good to me!

PR #32, opened 2017-08-01
Fix build on i686

dcoffill:
Smoltcp doesn't build on 32-bit Linux (i686-unknown-linux-gnu) with default features.  This is because the  width of the `libc::timeval` `tv_usec` member is platform and architecture dependent.  To fix this, I simply changed the initialization to use the `libc::suseconds_t` type alias defined by `libc`, instead of specifying the exact bit width.


whitequark:
Thanks!

Issue #19, opened 2017-06-18
[RFC] Packet dispatch


batonius:
### Problem
Currently, sloltcp uses iteration to dispatch ingress packets to sockets and to poll sockets for egress packets. While acceptable on the platforms without `libstd` and memory allocation, this approach is very inefficient on systems where growable containers are available.
### Discussion
1. The problem is two-fold: ingress packets should be dispatched by lookup tables based on their destination, while only "dirty" sockets should be polled for egress packets.
2. After #18 is merged, there will be two kinds of sockets: l3 (raw) sockets should be dispatched to based on packet's IP version and IP protocol number, and l4 sockets (tcp and udp) should be dispatched to based on the dst port.
3. Sockets can be bound to a specific IP address. As a result, a solution should support several sockets of the same type listening on the same port but bound to different IP addresses.
4. A solution should keep the existing behavior when compiled without the `std` feature.
### Proposal
1. Dispatch packets to sockets based on the port/protocol number only, leaving the question of sockets being bound to IPs to the `Socket::proccess` method.
2. Expand `SocketSet` by (conditionally) adding three `std::HashMap`'s to it, for raw, tcp and udp sockets, and a `Vec` to keep dirty sockets' handlers.
3. Hash map for raw sockets should have a type of `std::HashMap<(RawSocketType, IpProtocol), Vec<Handle>>`, hash maps for udp and tcp should have a type of `std::HashMap<u16, Vec<Handle>>` . `Vec` is required to support the possibility of several sockets listening to the same port/protocol on different addresses.
4. Extend `SocketSet::add` to use socket's `endpoint` and `ip_protocol` methods to construct an index to insert socket's handler into the corresponding lookup table.
5. Extend `SocketSet::remove` to remove socket from a lookup table based on socket's `endpoint`.
6. Extend `Set::IterMut` to support construction from a mutable reference to `ManagedSlice` and an iterator of `Handles`, producing an iterator of Socket.
7. Replace `SocketSet::iter_mut` with `iter_raw(&mut self, ip_repr: &IpRepr) -> IterMut` and  `iter_l4(&mut self, ip_repr: &IpRepr) -> IterMut`, which should look up sockets based on ip_repr and return an iterator over the sockets found. With the `std` feature disabled, the functions should return iterators over the entire `SocketSet`. 
[...]

PR #34, opened 2017-08-28
Expand ring_buffer API

batonius:
**Problem**: the `ring_buffer` container lacks several useful methods.

**Solution**: add the missing methods.

- the `with_default`, `capacity`, `remove` and `expand_storage` methods.
- tests for the new methods.

**Other**:
This PR is part of #19 effort.

### DIFF
--- a/src/socket/raw.rs
+++ b/src/socket/raw.rs
@@ -43,6 +43,9 @@ impl<'a> PacketBuffer<'a> {
 
 impl<'a> Resettable for PacketBuffer<'a> {
     fn reset(&mut self) {
+        for i in 0..self.size {
+            self.payload[i] = 0u8;
+        }
         self.size = 0;
     }
 }
--- a/src/socket/udp.rs
+++ b/src/socket/udp.rs
@@ -46,6 +46,9 @@ impl<'a> PacketBuffer<'a> {
 impl<'a> Resettable for PacketBuffer<'a> {
     fn reset(&mut self) {
         self.endpoint = Default::default();
+        for i in 0..self.size {
+            self.payload[i] = 0u8;
+        }
         self.size = 0;
     }
 }


src/storage/ring_buffer.rs
```diff
@@ -30,6 +30,24 @@ impl<'a, T: 'a> RingBuffer<'a, T> {
         }
     }
 
+    /// Create a ring buffer with the given storage.
+    ///
+    /// During creation, every element in `storage` is set to Default::default().
+    pub fn with_default<S>(storage: S) -> RingBuffer<'a, T>
+        where S: Into<ManagedSlice<'a, T>>, T: Default,
```
whitequark: I'm uncomfortable about this ambiguity, having two separate paths for resetting and creating a default member. How about making `Resettable` imply `Default` instead?

src/storage/ring_buffer.rs
```diff
@@ -101,6 +119,67 @@ impl<'a, T: 'a> RingBuffer<'a, T> {
             Err(error) => Err(error)
         }
     }
+
+    /// Get capacity of the underlying storage.
```
whitequark: Weird to have `capacity()` but not `len()`. 

Style nit: these should go up after `full()` and `empty()`, where all the query methods live.

src/storage/ring_buffer.rs
```diff
@@ -101,6 +119,67 @@ impl<'a, T: 'a> RingBuffer<'a, T> {
...
+    pub fn capacity(&self) -> usize {
+        self.storage.len()
+    }
+
+    /// Remove the first element equal to `value` from the buffer,
+    /// reutrn `Err(Error::Exhausted)` if no such element was found.
+    pub fn remove(&mut self, value: &T) -> Result<()>
```
whitequark: This seems to go completely against the spirit of a ring buffer. Why do you need this?

src/storage/ring_buffer.rs
```diff
@@ -101,6 +119,67 @@ impl<'a, T: 'a> RingBuffer<'a, T> {
...
+            }
+            self.length -= 1;
+            Ok(())
+        } else {
+            Err(Error::Exhausted)
+        }
+    }
+
+    /// Try to expand the underlying storage, panic if the storage cannot be expanded.
+    #[cfg(not(any(feature = "std", feature = "collections")))]
```
whitequark: I'd prefer having a single method and optionally compiling in including the `ManagedSlice::Owned` branch.

src/storage/ring_buffer.rs
```diff
@@ -101,6 +119,67 @@ impl<'a, T: 'a> RingBuffer<'a, T> {
...
+    pub fn expand_storage(&mut self) {
+        panic!("Cannot expand a ring_buffer in not-owned memory")
+    }
+
+    #[cfg(any(feature = "std", feature = "collections"))]
+    pub fn expand_storage(&mut self)
+        where T: Copy + Default {
```
whitequark: Do you need `Copy`? I think `mem::swap` or even `slice::swap` is enough to implement the expansion.

whitequark: I guess that's twice as much work so the Copy bound is fine.

src/storage/ring_buffer.rs
batonius: I don't think we can do it, otherwise, I would have gone with `Default` when I refactored `RingBuffer` out. Currently, we're using `RingBuffer` to store `PacketBuffer` for UDP and Raw sockets, and these contain `Managed<'a, [u8]>`, meaning we can't just `Default::default()` them, so we use `Resettable` to reset the other fields. Yet for 90% of the potential use cases `Default::default()` would be enough and there's no need to require `Resettable`.

Now I think of it, maybe we should just get rid of `Resettable` and initialization of elements in the constructors altogether? I can't remember why I even need them, the container itself doesn't care about the values, and we already initialize them in the constructors before inserting.

whitequark: > Now I think of it, maybe we should just get rid of Resettable and initialization of elements in the constructors altogether?

I think you'll get infoleaks when reusing sockets then.

src/storage/ring_buffer.rs
batonius: I was using `RingBuffer` as a queue of dirty sockets and I needed to delete sockets from the queue once they're removed from the main container. I agree its usefulness isn't evident, but it could be handy to have.

whitequark: Ah. But as I've explained in [this comment](https://github.com/m-labs/smoltcp/issues/19#issuecomment-325299849) a queue won't work here, you'll need a BTreeMap as well, see my latest commit for more context as to why.

src/storage/ring_buffer.rs
batonius: We only reset elements in the `RingBuffer`'s constructor tho, `try_enqueue` explicitly takes a function to reset the new element, and `enqueue` doesn't do it at all.

whitequark: Exactly. Suppose application A put key material all over the socket. Then, the socket, without being reset, is passed to application B. This application can now call `.enqueue()` and read from the buffer what it should have never seen. Ideally we'd have something like `&write [u8]` but alas.

batonius: This is an interesting point I didn't think about, but I think we should just make `RingBuffer`'s methods pub(crate) since a user doesn't need to use them directly. It's impossible to get access to the old data via sockets' methods because they all limited by the length of the data in the buffer. Besides, I think memsetting buffers all the time would be too costly.

whitequark: Of course it's possible, just use `tcp_socket.send()`, which directly calls `enqueue()`...

src/storage/ring_buffer.rs
batonius: Ok then.

src/storage/ring_buffer.rs
batonius: `TcpSocket` doesn't use `RingBuffer` right now, but you're right, you can get access to the old data via `UdpSocket::send()`. So the current problem is twofold - we don't call `reset` on element reuse and we don't memset the buffer in `reset`. I guess I'll do it then. This means the `Resettable` bound is now container-wide and there's no point in `with_default` constructor.

Now I need to find a way to memset a buffer in such a way that the compiler won't optimize it out :)

whitequark: It's only legal for memset to be removed right before deallocation, and we don't have that, so it's not a concern.

batonius:
- Removed `remove` method and `with_default` constructor.
- Added `len` method.
- `RingBuffer` now calls `Resetable::reset()` when enqueuing an element.
- `UdpPacketBuffer` and `RawPacketBuffer` now memset their payload in `Resetable::reset()`.

whitequark:
Okay, so I think we have some confusion with this PR.

  * The new `expand_storage` method is fine but it seems more useful to call it `reserve_exact` to match the standard containers and let expanding by more than a single element. I'm not sure if it's ever useful to do that, a reasonable strategy is something like adding 50% more buffers once you're 75% full or something.
  * Let's not add any unsafe code, smoltcp has 0 grep hits for `unsafe` and I think this is an important advantage.
  * I think I miscommunicated what I talked about wrt resetting to you. It already works properly. It's not actually necessary to reset elements on every `enqueue()` call because every such call after the socket is open comes from the same application. The idea is that a TCP/IP stack implementation would have a pool of packet (and/or socket) buffers; when a new application asks it for a socket, it creates a new UdpSocket (cheap) and assigns it a socket buffer from a pool (cheap) or if none are available, creates a new one (expensive); in either case the UdpSocket constructor does the reset. After the application is done, the UDP socket is destroyed; if you look at the API, there isn't even a way to close an UDP socket once it's open.

I'm sorry for making you do all this work, I'll try to be more clear from the outset in the future...

batonius:
No, it's ok, I should have thought it through before making the PR.

I think I understand the resetting issue, but it took me some time to realize "socket buffer" doesn't mean `SocketBuffer` :) The UDP constructor doesn't actually reset anything, but it consumes instances of `SocketBuffer` which need to be created for each new socket from `ManagedSlice`s (from a pool?), and this is where the resetting happens.  I can't see how these buffers can be returned to the pool right now tho: since `ManagedSlice` owns a vector, `SocketBuffer` owns a slice and `UdpSocket` owns buffers, dropping a socket would free the vectors.

Anyway, I see we don't need resetting in `enqueue`, but I reckon we still need to actually erase old payload during the resetting.

whitequark:
Let's address buffer resetting in a coordinated way across the entire codebase when we'll make it possible to return buffers to the pool. The changes here are trivial and insufficient on their own.

Regarding expandable ring buffers, I think that was needed for egress handling, but egress is now going to use BTreeMap too, to sort sockets by the earliest time they should be pollwed, so it seems like it's not necessary for now. If it ever is, it's always possible to cherry-pick the commit from here.

Thus, closing.

PR #38, opened 2017-08-31
Refactor Socket::process into Socket::would_accept and Socket::process_accepted

batonius:
**Problem**: Currently the `process` methods both check whatever the packet can be processed by the socket and do the processing. While it makes sense at the moment, the eventual adoption of a packet dispatching mechanism will make the checks optional in some configurations, thus it's beneficial to split the methods in two preemptively. As a side effect, the refactoring will allow us to get rid of the `Error::Rejected` error, which doesn't represent a problem with a packet or a socket, but serves as a marker of a failed packet dispatching.

**Solution**: Split the `process` methods into `would_accept` and `process_accepted`.

- `TcpSocket::process`, `UdpSocket::process` and `RawSocket::process` have been refactored.
- Tests for new methods have been added.
- `Error::Rejected` has been removed as not needed anymore.

**Other**:
This PR is a part of #19 effort.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,10 +126,6 @@ pub enum Error {
     /// E.g. a TCP packet addressed to a socket that doesn't exist.
     Dropped,
 
-    // Implementation detail.
-    #[doc(hidden)]
-    Rejected,
-
     #[doc(hidden)]
     __Nonexhaustive
 }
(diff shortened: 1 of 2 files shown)


src/iface/ethernet.rs
```diff
@@ -285,13 +285,13 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
-                Ok(()) => handled_by_raw_socket = true,
-                // The packet isn't addressed to the socket, or cannot be accepted by it.
-                Err(Error::Rejected) | Err(Error::Exhausted) => (),
-                // Raw sockets either accept or reject packets, not parse them.
-                _ => unreachable!(),
+            if raw_socket.would_accept(&ip_repr) {
+                match raw_socket.process_accepted(&ip_repr, ip_payload) {
+                    // The packet is valid and handled by socket.
+                    Ok(()) => handled_by_raw_socket = true,
+                    // Raw sockets can't fail to process an accepted packet
```
whitequark: They can if they're out of space.

whitequark:
Manually merged with stylistic changes.

PR #42, opened 2017-09-12
Use smart pointers in SocketSet's API

batonius:
**Problem**: Currently `SocketSet` return sockets as `&mut Socket` which makes impossible to automatically track state changes, which, in turn, makes implementing packet dispatch problematic.

**Solution**: Implement `SocketRef` smart reference, which automatically calls bookkeeping code at the end of a usage session. Change `SocketSet` to return `SocketRef` instead of `&mut Socket`.

- the `SocketRef` struct has been added;
- the `SocketSession` trait for socket type-specific usage session cleanup has been added;
- `SocketSet` returns `SocketRef` instead of `&mut Socket`;
- the `AsSocket`trait has been refactored into `FromSocket`;

**Other**: Currently `SocketRef` holds only socket's handle and `&mut *Socket`, and all `SocketSession::finish` implementations are empty. 

This PR is a part of #19 effort.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -12,8 +12,7 @@ use std::os::unix::io::AsRawFd;
 use smoltcp::phy::wait as phy_wait;
 use smoltcp::wire::{EthernetAddress, Ipv4Address, IpAddress, IpCidr};
 use smoltcp::iface::{ArpCache, SliceArpCache, EthernetInterface};
-use smoltcp::socket::{AsSocket, SocketSet};
-use smoltcp::socket::{TcpSocket, TcpSocketBuffer};
+use smoltcp::socket::{SocketSet, TcpSocket, TcpSocketBuffer};
 
 fn main() {
     utils::setup_logging("");
@@ -50,14 +49,14 @@ fn main() {
     let tcp_handle = sockets.add(tcp_socket);
 
     {
-        let socket: &mut TcpSocket = sockets.get_mut(tcp_handle).as_socket();
+        let mut socket = sockets.get::<TcpSocket>(tcp_handle);
         socket.connect((address, port), 49500).unwrap();
     }
 
     let mut tcp_active = false;
     loop {
         {
-            let socket: &mut TcpSocket = sockets.get_mut(tcp_handle).as_socket();
+            let mut socket = sockets.get::<TcpSocket>(tcp_handle);
             if socket.is_active() && !tcp_active {
                 debug!("connected");
             } else if !socket.is_active() && tcp_active {
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -13,7 +13,7 @@ use wire::{Ipv4Packet, Ipv4Repr};
 use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
 #[cfg(feature = "socket-udp")] use wire::{UdpPacket, UdpRepr};
 #[cfg(feature = "socket-tcp")] use wire::{TcpPacket, TcpRepr, TcpControl};
-use socket::{Socket, SocketSet, AsSocket};
+use socket::{Socket, SocketSet, FromSocket};
 #[cfg(feature = "socket-raw")] use socket::RawSocket;
 #[cfg(feature = "socket-udp")] use socket::UdpSocket;
 #[cfg(feature = "socket-tcp")] use socket::TcpSocket;
@@ -184,29 +184,29 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
         let mut caps = self.device.capabilities();
         caps.max_transmission_unit -= EthernetFrame::<&[u8]>::header_len();
 
-        for socket in sockets.iter_mut() {
+        for mut socket in sockets.iter_mut() {
             let mut device_result = Ok(());
             let socket_result =
-                match socket {
+                match *socket {
                     #[cfg(feature = "socket-raw")]
-                    &mut Socket::Raw(ref mut socket) =>
+                    Socket::Raw(ref mut socket) =>
                         socket.dispatch(|response| {
                             device_result = self.dispatch(timestamp, Packet::Raw(response));
                             device_result
                         }, &caps.checksum),
                     #[cfg(feature = "socket-udp")]
-                    &mut Socket::Udp(ref mut socket) =>
+                    Socket::Udp(ref mut socket) =>
                         socket.dispatch(|response| {
                             device_result = self.dispatch(timestamp, Packet::Udp(response));
                             device_result
                         }),
                     #[cfg(feature = "socket-tcp")]
-                    &mut Socket::Tcp(ref mut socket) =>
+                    Socket::Tcp(ref mut socket) =>
                         socket.dispatch(timestamp, &caps, |response| {
                             device_result = self.dispatch(timestamp, Packet::Tcp(response));
                             device_result
                         }),
-                    &mut Socket::__Nonexhaustive(_) => unreachable!()
+                    Socket::__Nonexhaustive(_) => unreachable!()
                 };
             match (device_result, socket_result) {
                 (Err(Error::Unaddressable), _) => break, // no one to transmit to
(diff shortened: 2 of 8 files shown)


src/socket/socket_ref.rs
```diff
@@ -0,0 +1,83 @@
...
+            fused: true,
+        }
+    }
+}
+
+impl<'a, 'b, 'c, T: FromSocket<'b, 'c>> SocketRef<'a, T> {
+    pub(crate) fn from_socket_ref(mut socket_ref: SocketRef<'a, Socket<'b, 'c>>)
+                                  -> Option<SocketRef<'a, T>> {
+        if let Some(socket) = T::from_socket(socket_ref.socket) {
+            socket_ref.fused = false;
```
batonius: It's possible I'm missing something here: I want to prevent a spurious destructor call, but I can't use `mem::forget` since `socket_ref` is being borrowed from until the end of the method. It doesn't prevent the destructor from being called tho, so I have to do it by hand.

examples/client.rs
```diff
@@ -49,14 +48,14 @@ fn main() {
     let tcp_handle = sockets.add(tcp_socket);
 
     {
-        let socket: &mut TcpSocket = sockets.get_mut(tcp_handle).as_socket();
+        let mut socket = sockets.get::<TcpSocket>(tcp_handle).unwrap();
```
whitequark: I believe dispatching over the socket type is the rare case (it only happens inside smoltcp and in OS/firmware code that's generic over the socket, e.g. fd handling), so let's have a panicking .get() and a non-panicking get_opt(), seems cleaner than unwrapping everywhere.

src/iface/ethernet.rs
```diff
@@ -165,26 +165,26 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
         let mut limits = self.device.limits();
         limits.max_transmission_unit -= EthernetFrame::<&[u8]>::header_len();
 
-        for socket in sockets.iter_mut() {
+        for mut socket in sockets.iter_mut() {
             let mut device_result = Ok(());
             let socket_result =
-                match socket {
-                    &mut Socket::Raw(ref mut socket) =>
+                match *(&mut *socket) {
```
whitequark: *squint* what... is this thing

src/socket/mod.rs
```diff
@@ -16,6 +16,7 @@ mod raw;
 mod udp;
 mod tcp;
 mod set;
+mod socket_ref;
```
whitequark: `mod ref`, `use self::ref::Ref` and then reexport as a prefixed name; that's how everything else is done here.

src/socket/socket_ref.rs
```diff
@@ -0,0 +1,83 @@
...
+            Some(SocketRef::new(socket, socket_ref.handle))
+        } else {
+            None
+        }
+    }
+}
+
+impl<'a, T: SocketSession> Deref for SocketRef<'a, T> {
```
whitequark: We don't need Deref for SocketRef since all useful methods on sockets use &mut self anyway. It seems like this would lead to people thinking they can do something with &XxxSocket.

src/socket/socket_ref.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Allows implementation of custom on-drop logic.
+pub trait SocketSession {
+    fn finish(&mut self) {}
+}
+
+impl<'a, 'b> SocketSession for RawSocket<'a, 'b> {}
+impl<'a, 'b> SocketSession for UdpSocket<'a, 'b> {}
+impl<'a> SocketSession for TcpSocket<'a> {}
+
+impl<'a, 'b> SocketSession for Socket<'a, 'b> {
```
whitequark: We don't (and can't, Rust enums are not extensible) support new socket types defined in external crates, so the entire SocketSession logic can simply be folded into the SocketRef Drop impl.

src/socket/set.rs
```diff
@@ -164,9 +152,9 @@ impl<'a, 'b: 'a, 'c: 'a + 'b> Set<'a, 'b, 'c> {
         Iter { lower: self.sockets.iter() }
     }
 
-    /// Iterate every socket in this set, as mutable.
+    /// Iterate every socket in this set, as SocketRef.
     pub fn iter_mut<'d>(&'d mut self) -> IterMut<'d, 'b, 'c> {
-        IterMut { lower: self.sockets.iter_mut() }
+        IterMut { lower: self.sockets.iter_mut(), index: 0 }
```
whitequark: It would be more elegant to use `.iter_mut().enumerate()` here.

src/socket/mod.rs
```diff
@@ -32,19 +33,14 @@ pub use self::tcp::TcpSocket;
...
-/// the [AsSocket](trait.AsSocket.html) trait, and call e.g. `socket.as_socket::<UdpSocket<_>>()`.
-///
-/// The `process` and `dispatch` functions are fundamentally asymmetric and thus differ in
-/// their use of the [trait PacketRepr](trait.PacketRepr.html). When `process` is called,
-/// the packet length is already known and no allocation is required; on the other hand,
-/// `process` would have to downcast a `&PacketRepr` to e.g. an `&UdpRepr` through `Any`,
-/// which is rather inelegant. Conversely, when `dispatch` is called, the packet length is
-/// not yet known and the packet storage has to be allocated; but the `&PacketRepr` is sufficient
-/// since the lower layers treat the packet as an opaque octet sequence.
+/// the [FromSocket](trait.FromSocket.html) trait, and call e.g. `UdpSocket::from_socket(socket)`.
```
whitequark: Worth mentioning SocketSet here and that they're probably looking for `socket_set.get::<TcpSocket>(tcp_handle)` instead

src/socket/mod.rs
```diff
@@ -85,36 +81,29 @@ impl<'a, 'b> Socket<'a, 'b> {
...
 
 /// A conversion trait for network sockets.
-///
-/// This trait is used to concisely downcast [Socket](trait.Socket.html) values to their
-/// concrete types.
-pub trait AsSocket<T> {
-    fn as_socket(&mut self) -> &mut T;
-    fn try_as_socket(&mut self) -> Option<&mut T>;
+pub trait FromSocket<'a, 'b>: SocketSession + Sized {
+    fn from_socket<'c>(&'c mut Socket<'a, 'b>) -> Option<&'c mut Self>;
```
whitequark: `downcast_socket` maybe? `from_socket` is not really indicative of what this does.

src/socket/socket_ref.rs
whitequark: `Option<&mut T>` maybe? If I understand correctly, this Option would be deconstructed exactly three times, one in constructor, one in destructor and one in deref, so the one added jump is basically nothing. Then just take the option.

Not worth introducing unsafe code just for this IMO.

src/socket/socket_ref.rs
batonius: `DerefMut` requires `Deref`.

src/socket/set.rs
batonius: This is what I did initially, but `std::iter::Enumerate` requires `std`, so I have to do it manually.

src/iface/ethernet.rs
batonius: Looks like I've got carried away a little.

src/socket/socket_ref.rs
batonius: `SocketRef` is generic over socket type, we can't put socket-specific cleanup code there. Besides, I'm going to extend `SocketSession` with an associated type that holds socket's state, something like https://github.com/batonius/smoltcp/blob/packet_dispatch/src/socket/tracker.rs#L11 .

src/socket/set.rs
batonius: Apparently, I've missed `core::iter`.

src/socket/socket_ref.rs
batonius: `Option<&mut T>` in `SocketRef` would mean `DerefMut::deref_mut` could panic, and there's really no reason for it. I guess I live it as it is for now.

src/socket/mod.rs
batonius: `ref` is a keyword, so I can't use `mod ref;`, I'll rename `socket_ref::SocketRef` to `socket_ref::Ref`.

batonius:
Updated.

whitequark:
Merged manually with... lots of changes.
  * Some renames.
  * A lot of doc improvement.
  * Merged FromSocket::downcast and ::from_socket_ref functions into one, using Ref::map (like libcore's own BorrowMut).
  * Folded SocketHandle into sockets themselves since they basically already have that in form of debug_id.
  * Removed try_get since it's not used anywhere.
  * Made panics in SocketSet more informative.
Please study them and take note.

Issue #36, opened 2017-08-30
Hardware Checksum Calculation


steffengy:
I think the possibility to move checksum calculation to the hardware when possible,
would be neat.

E.g. With LWIP you can do

```c
#define CHECKSUM_BY_HARDWARE
#define CHECKSUM_GEN_IP 0
#define CHECKSUM_GEN_UDP 0
#define CHECKSUM_GEN_TCP 0
#define CHECKSUM_GEN_ICMP 0
#define CHECKSUM_CHECK_IP 0
#define CHECKSUM_CHECK_UDP 0
#define CHECKSUM_CHECK_TCP 0
#define CHECKSUM_CHECK_ICMP 0
```

and use `Checksum insertion control` of STM32 to let the hardware handle those checksums.
You'd simply not calculate the checksum in the software &
the SoC will insert it before transmitting/receiving the frame.

whitequark:
Sure. Do you have any particular hardware in mind? Show me the datasheet and I'll think about exposing this capability.

steffengy:
STM32F2xxx/F4xxx (F4 is just a superset with more functionality basically, ethernet is afaik the same):

http://www.st.com/content/ccc/resource/technical/document/reference_manual/3d/6d/5a/66/b4/99/40/d4/DM00031020.pdf/files/DM00031020.pdf/jcr:content/translations/en.DM00031020.pdf

Searching for "checksum" does already give good results.

steffengy:
To clarify, since it wasn't really clearly worded:

The only thing that smoltcp would have to support for this, 
would be disabling checksum generation (use zeros as checksum).

The hardware will generate & insert the checksum on its own before sending the frame.

As shown above LWIP (an ip stack) supports fine-grained control of checksums e.g. disabling them for ICMP, ...

Not sure how to best expose this in rust, using features would lead to a "feature overload".

whitequark:
Yeah, I understood. I haven't had time to flesh out the ideas I had, hence the silence.

whitequark:
OK, so the route to fixing this would be to:
  * introduce a `enum Checksum { Both, Rx, Tx, None }`;
  * add a parameter `verify_cksum: Checksum` to all `*Repr::{parse,emit}` methods;
  * rename `DeviceLimits` into `DeviceCaps` or something (for "capabilities");
  * add fields `soft_ethernet_cksum: Checksum, soft_ip_cksum: Checksum, ..` to `DeviceCaps`;
[...]

PR #43, opened 2017-09-15
hardware based checksum generation & validation

steffengy:
This introduces the changes discussed in #36.

open issues:
- ~~ICMPv4 checksums don't seem to be implemented? (unused checksum warning in icmpv4::parse)~~

### DIFF
--- a/examples/ping.rs
+++ b/examples/ping.rs
@@ -10,6 +10,7 @@ mod utils;
 use std::str::FromStr;
 use std::time::Instant;
 use std::os::unix::io::AsRawFd;
+use smoltcp::phy::Device;
 use smoltcp::phy::wait as phy_wait;
 use smoltcp::wire::{EthernetAddress, IpVersion, IpProtocol, IpAddress,
                     Ipv4Address, Ipv4Packet, Ipv4Repr,
@@ -56,6 +57,7 @@ fn main() {
                                     raw_rx_buffer, raw_tx_buffer);
 
     let hardware_addr  = EthernetAddress([0x02, 0x00, 0x00, 0x00, 0x00, 0x02]);
+    let caps = device.capabilities();
     let mut iface = EthernetInterface::new(
         Box::new(device), Box::new(arp_cache) as Box<ArpCache>,
         hardware_addr, [IpAddress::from(local_addr)]);
@@ -98,9 +100,9 @@ fn main() {
                     .unwrap();
 
                 let mut ipv4_packet = Ipv4Packet::new(raw_payload);
-                ipv4_repr.emit(&mut ipv4_packet);
+                ipv4_repr.emit(&mut ipv4_packet, &caps.checksum);
                 let mut icmp_packet = Icmpv4Packet::new(ipv4_packet.payload_mut());
-                icmp_repr.emit(&mut icmp_packet);
+                icmp_repr.emit(&mut icmp_packet, &caps.checksum);
 
                 waiting_queue.insert(seq_no, timestamp);
                 seq_no += 1;
@@ -110,11 +112,11 @@ fn main() {
             if socket.can_recv() {
                 let payload = socket.recv().unwrap();
                 let ipv4_packet = Ipv4Packet::new(payload);
-                let ipv4_repr = Ipv4Repr::parse(&ipv4_packet).unwrap();
+                let ipv4_repr = Ipv4Repr::parse(&ipv4_packet, &caps.checksum).unwrap();
 
                 if ipv4_repr.src_addr == remote_addr && ipv4_repr.dst_addr == local_addr {
                     let icmp_packet = Icmpv4Packet::new(ipv4_packet.payload());
-                    let icmp_repr = Icmpv4Repr::parse(&icmp_packet);
+                    let icmp_repr = Icmpv4Repr::parse(&icmp_packet, &caps.checksum);
 
                     if let Ok(Icmpv4Repr::EchoReply { seq_no, data, .. }) = icmp_repr {
                         if let Some(_) = waiting_queue.get(&seq_no) {
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -181,7 +181,7 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
                         socket.dispatch(|response| {
                             device_result = self.dispatch(timestamp, Packet::Raw(response));
                             device_result
-                        }),
+                        }, &caps.checksum),
                     #[cfg(feature = "socket-udp")]
                     &mut Socket::Udp(ref mut socket) =>
                         socket.dispatch(|response| {
@@ -190,7 +190,7 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
                         }),
                     #[cfg(feature = "socket-tcp")]
                     &mut Socket::Tcp(ref mut socket) =>
-                        socket.dispatch(timestamp, &limits, |response| {
+                        socket.dispatch(timestamp, &caps, |response| {
                             device_result = self.dispatch(timestamp, Packet::Tcp(response));
                             device_result
                         }),
@@ -278,7 +278,8 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
                     eth_frame: &EthernetFrame<&'frame T>) ->
                    Result<Packet<'frame>> {
         let ipv4_packet = Ipv4Packet::new_checked(eth_frame.payload())?;
-        let ipv4_repr = Ipv4Repr::parse(&ipv4_packet)?;
+        let checksum_caps = self.device.capabilities().checksum;
+        let ipv4_repr = Ipv4Repr::parse(&ipv4_packet, &checksum_caps)?;
 
         if !ipv4_repr.src_addr.is_unicast() {
             // Discard packets with non-unicast source addresses.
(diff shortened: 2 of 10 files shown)


steffengy:
@whitequark 
Currently this approach fails for some modified examples e.g. 
```
examples/ping.rs:104:32
104 |             icmp_repr.emit(&mut icmp_packet);
```

which would have to become
```
examples/ping.rs:104:32
104 |             icmp_repr.emit(&mut icmp_packet, iface.device.capabilities().checksum.icmpv4);
```

Does that seem fine to you or do you have a better solution in mind?

whitequark:
Could you please split your changes into several commits? Reviewing them as one large diff is pretty hard.

steffengy:
@whitequark 
done, be aware that only the latest commit will build though.

steffengy:
@whitequark 
Additionally I just fixed the remaining tests as I saw fit.

Tests now pass, let me know what else should be done.

steffengy:
@whitequark 
rebased again.

src/phy/mod.rs
```diff
@@ -135,12 +135,57 @@ pub use self::tap_interface::TapInterface;
...
+        match *self {
+            Checksum::Both | Checksum::Tx => true,
+            _ => false
+        }
+    }
+}
+
+/// Configuration of checksum capabilities for each applicable protocol
+#[derive(Debug, Clone, Default)]
+pub struct ChecksumCaps {
```
whitequark: `ChecksumCapabilities`, for consistence.

src/phy/mod.rs
```diff
@@ -135,12 +135,57 @@ pub use self::tap_interface::TapInterface;
...
+    pub ipv4: Checksum,
+    pub udp: Checksum,
```
whitequark: I can imagine that a network card could handle ipv4 checksum but not ipv6, so perhaps split these off into udpv4 and tcpv4?

src/wire/icmpv4.rs
```diff
@@ -441,7 +442,9 @@ impl<'a> Repr<'a> {
 
     /// Emit a high-level representation into an Internet Control Message Protocol version 4
     /// packet.
-    pub fn emit<T: AsRef<[u8]> + AsMut<[u8]> + ?Sized>(&self, packet: &mut Packet<&mut T>) {
+    pub fn emit<T: AsRef<[u8]> + AsMut<[u8]> + ?Sized>(&self, 
+                                                       packet: &mut Packet<&mut T>, 
+                                                       emit_checksum: Checksum) {
```
whitequark: I think it's worth passing a `&ChecksumCapabilities` around because all of the parse/emit methods already know what kind of packet are they working with, not individual `Checksum` fields, since then the resulting code is easier to read. `header.emit(&mut ip_packet, checksum_caps)` plus `if checksum_caps.ipv4.tx() { ... }` vs `header.emit(&mut ip_packet, emit_checksum)` plus `emit_checksum.tx()`, where `emit_checksum` actually conatins both emission and parsing capabilities in the latter case.

src/phy/fault_injector.rs
```diff
@@ -188,12 +188,12 @@ impl<D: Device> Device for FaultInjector<D>
     type RxBuffer = D::RxBuffer;
     type TxBuffer = TxBuffer<D::TxBuffer>;
 
-    fn limits(&self) -> DeviceLimits {
```
whitequark: Do you think you could finish the renaming in this commit? I'd like to keep history bisectable, and generally clean.

src/socket/raw.rs
```diff
@@ -196,7 +196,11 @@ impl<'a, 'b> RawSocket<'a, 'b> {
                 IpVersion::Ipv4 => {
                     let mut packet = Ipv4Packet::new_checked(buffer.as_mut())?;
                     if packet.protocol() != protocol { return Err(Error::Unaddressable) }
-                    packet.fill_checksum();
+                    if emit_checksum.ipv4.tx() {
+                        packet.fill_checksum();
+                    } else {
+                        packet.set_checksum(0);
```
whitequark: What's the rationale here?

examples/ping.rs
```diff
@@ -10,6 +10,7 @@ mod utils;
 use std::str::FromStr;
```
whitequark: Do you think you could fold this commit into previous ones? Same reasoning as before.

src/socket/raw.rs
steffengy: Ensuring the checksum field is surely zeroed in the resulting packet.
Not sure if the implementation has an issue for other cases,
but it did cause issues for ICMPV4 for STM32, as indicated by the reference manual:

```
Note that: for ICMP-over-IPv4 packets, the checksum field in the ICMP packet must
always be 0x0000 in both modes, because pseudo-headers are not defined for such
packets. If it does not equal 0x0000, an incorrect checksum may be inserted into the
packet.
```
http://www.st.com/content/ccc/resource/technical/document/reference_manual/51/f7/f3/06/cd/b6/46/ec/CD00225773.pdf/files/CD00225773.pdf/jcr:content/translations/en.CD00225773.pdf (page 846)

steffengy:
@whitequark 
Addressed your feedback, looks far better now.
I split the changes into 2 commits: First renaming and second the whole implementation
and both commits can be built on their own.

Let me know if that works for you :)

jordens:
Out of curiosity: Does someone have data on how much this speeds things up? How much time is spent on checksums in smoltcp (excluding the higher non-smoltcp protocol layers)?

src/socket/raw.rs
whitequark: Okay, this is worth a small comment in every call site.

src/wire/udp.rs
```diff
@@ -203,20 +204,24 @@ impl<'a> Repr<'a> {
...
-        if !packet.verify_checksum(src_addr, dst_addr) {
-            match (src_addr, dst_addr) {
-                (&IpAddress::Ipv4(_), &IpAddress::Ipv4(_))
-                        if packet.checksum() != 0 => {
-                    // ... except on UDP-over-IPv4, where it can be omitted.
-                    return Err(Error::Checksum)
-                },
-                _ => {
-                    return Err(Error::Checksum)
+        if checksum.udpv4.rx() {
```
whitequark: `if checksum.udpv4.rx() && !packet.verify_checksum(src_addr, dst_addr)` avoids excessive rightwards drift.

src/wire/tcp.rs
```diff
@@ -644,13 +645,17 @@ impl<'a> Repr<'a> {
...
+                            dst_addr: &IpAddress,
+                            checksum: &ChecksumCapabilities) -> Result<Repr<'a>>
             where T: AsRef<[u8]> {
         // Source and destination ports must be present.
         if packet.src_port() == 0 { return Err(Error::Malformed) }
         if packet.dst_port() == 0 { return Err(Error::Malformed) }
+        
         // Valid checksum is expected...
-        if !packet.verify_checksum(src_addr, dst_addr) { return Err(Error::Checksum) }
+        if checksum.tcpv4.rx() {
```
whitequark: Same as below on rightwards drift.

src/iface/ethernet.rs
```diff
@@ -352,10 +353,11 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
         }
     }
 
-    fn process_icmpv4<'frame>(ipv4_repr: Ipv4Repr, ip_payload: &'frame [u8]) ->
-                             Result<Packet<'frame>> {
+    fn process_icmpv4<'frame>(&self, ipv4_repr: Ipv4Repr, 
+                              ip_payload: &'frame [u8]) -> Result<Packet<'frame>> {
         let icmp_packet = Icmpv4Packet::new_checked(ip_payload)?;
-        let icmp_repr = Icmpv4Repr::parse(&icmp_packet)?;
+        let checksum = self.device.capabilities().checksum;
```
whitequark: `checksum_caps` would be more clear. I actually got confused on the type of this variable while reviewing.

steffengy:
@whitequark 
Ready for another round, I hope it's clear enough now :)

src/iface/ethernet.rs
```diff
@@ -486,12 +491,13 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
                     udp_repr.emit(&mut UdpPacket::new(payload),
-                                  &ip_repr.src_addr(), &ip_repr.dst_addr());
+                                  &ip_repr.src_addr(), &ip_repr.dst_addr(), 
+                                  &checksum_caps);
                 })
             }
             #[cfg(feature = "socket-tcp")]
             Packet::Tcp((ip_repr, mut tcp_repr)) => {
-                let caps = self.device.capabilities();
+                let limits = self.device.capabilities();
```
whitequark: Rebase issue?

steffengy: Oops. Fixed.

src/wire/icmpv4.rs
```diff
@@ -384,7 +385,7 @@ pub enum Repr<'a> {
 impl<'a> Repr<'a> {
     /// Parse an Internet Control Message Protocol version 4 packet and return
     /// a high-level representation.
-    pub fn parse<T: AsRef<[u8]> + ?Sized>(packet: &Packet<&'a T>) -> Result<Repr<'a>> {
+    pub fn parse<T: AsRef<[u8]> + ?Sized>(packet: &Packet<&'a T>, _checksum: &ChecksumCapabilities) -> Result<Repr<'a>> {
```
whitequark: Shouldn't we verify checksum in this method? Looks like I forgot to implement it.

steffengy: Yeah I'm not sure what the story is around there, that's what I meant with:
`ICMPv4 checksums don't seem to be implemented? (unused checksum warning in icmpv4::parse)`
in my first comment.

I'll include that in the next round:
```rust
if checksum_caps.icmpv4.rx() && !packet.verify_checksum() { 
            return Err(Error::Checksum) 
        }
```

steffengy:
@whitequark And another round, thanks for the patience!

whitequark:
Thanks! I'm looking forward to more great contributions from you.

PR #44, opened 2017-09-16
Add support for default gateway

batonius:
**Context**: I've actually started working on the Redox network daemon based on smoltcp I was talking about back in June and so far I was able to replace L2 and L3 daemons(ethernetd and ipd), leaving L4 daemons like udpd and tcpd working on top of the new daemon.

**Problem**: I'm collecting a wishlist in the process, but the only problem so far is the inability to route packets destined to an IP outside a local subnet to a default gateway, making everything outside the local subnet unaccessible.

**Solution**: I've implemented the bare minimum of the required functional and it appears to be working. The routing is optional and disabled by default. A proper implementation would involve actual routing tables, but I propose we adopt this or other limited solution as a stop-gap for now.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -9,7 +9,7 @@ use wire::{EthernetAddress, EthernetProtocol, EthernetFrame};
 use wire::{ArpPacket, ArpRepr, ArpOperation};
 use wire::{Ipv4Packet, Ipv4Repr};
 use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
-use wire::{IpAddress, IpProtocol, IpRepr};
+use wire::{IpAddress, IpProtocol, IpRepr, IpCidr};
 #[cfg(feature = "socket-udp")] use wire::{UdpPacket, UdpRepr};
 #[cfg(feature = "socket-tcp")] use wire::{TcpPacket, TcpRepr, TcpControl};
 use socket::{Socket, SocketSet, AsSocket};
@@ -27,7 +27,8 @@ pub struct Interface<'a, 'b, 'c, DeviceT: Device + 'a> {
     device:         Managed<'a, DeviceT>,
     arp_cache:      Managed<'b, ArpCache>,
     hardware_addr:  EthernetAddress,
-    protocol_addrs: ManagedSlice<'c, IpAddress>,
+    protocol_addrs: ManagedSlice<'c, IpCidr>,
+    gateway:        Option<IpAddress>,
 }
 
 enum Packet<'a> {
@@ -48,24 +49,29 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
     /// # Panics
     /// See the restrictions on [set_hardware_addr](#method.set_hardware_addr)
     /// and [set_protocol_addrs](#method.set_protocol_addrs) functions.
-    pub fn new<DeviceMT, ArpCacheMT, ProtocolAddrsMT>
+    pub fn new<DeviceMT, ArpCacheMT, ProtocolAddrsMT, GatewayAddressT, GatewayT>
               (device: DeviceMT, arp_cache: ArpCacheMT,
-               hardware_addr: EthernetAddress, protocol_addrs: ProtocolAddrsMT) ->
+               hardware_addr: EthernetAddress, protocol_addrs: ProtocolAddrsMT,
+               gateway: GatewayT) ->
               Interface<'a, 'b, 'c, DeviceT>
             where DeviceMT: Into<Managed<'a, DeviceT>>,
                   ArpCacheMT: Into<Managed<'b, ArpCache>>,
-                  ProtocolAddrsMT: Into<ManagedSlice<'c, IpAddress>>, {
+                  ProtocolAddrsMT: Into<ManagedSlice<'c, IpCidr>>,
+                  GatewayAddressT: Into<IpAddress>,
+                  GatewayT: Into<Option<GatewayAddressT>>, {
         let device = device.into();
         let arp_cache = arp_cache.into();
         let protocol_addrs = protocol_addrs.into();
+        let gateway = gateway.into().map(Into::into);
 
         Self::check_hardware_addr(&hardware_addr);
         Self::check_protocol_addrs(&protocol_addrs);
         Interface {
-            device:         device,
-            arp_cache:      arp_cache,
-            hardware_addr:  hardware_addr,
-            protocol_addrs: protocol_addrs,
+            device,
+            arp_cache,
+            hardware_addr,
+            protocol_addrs,
+            gateway,
         }
     }
 
@@ -89,16 +95,16 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
         Self::check_hardware_addr(&self.hardware_addr);
     }
 
-    fn check_protocol_addrs(addrs: &[IpAddress]) {
-        for addr in addrs {
-            if !addr.is_unicast() {
-                panic!("protocol address {} is not unicast", addr)
+    fn check_protocol_addrs(addrs: &[IpCidr]) {
+        for cidr in addrs {
+            if !cidr.address().is_unicast() {
+                panic!("protocol address {} is not unicast", cidr.address())
             }
         }
     }
 
     /// Get the protocol addresses of the interface.
-    pub fn protocol_addrs(&self) -> &[IpAddress] {
+    pub fn protocol_addrs(&self) -> &[IpCidr] {
         self.protocol_addrs.as_ref()
     }
 
@@ -106,7 +112,7 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
     ///
     /// # Panics
     /// This function panics if any of the addresses is not unicast.
-    pub fn update_protocol_addrs<F: FnOnce(&mut ManagedSlice<'c, IpAddress>)>(&mut self, f: F) {
+    pub fn update_protocol_addrs<F: FnOnce(&mut ManagedSlice<'c, IpCidr>)>(&mut self, f: F) {
         f(&mut self.protocol_addrs);
         Self::check_protocol_addrs(&self.protocol_addrs)
     }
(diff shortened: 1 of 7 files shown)


little-dude:
Sorry for polluting the discussion, but I'm interested in following (and potentially helping) your redox integration @batonius, but I can't find where you're pushing the code in your github feed. Is this work private for the moment? I tried, long ago to use smoltcp in redox but I gave up half-way through.

whitequark:
> A proper implementation would involve actual routing tables, but I propose we adopt this or other limited solution as a stop-gap for now.

I agree. The only thing I would like to do differently here is to add a `struct Ipv4Cidr` instead of handling a bare IP and bit count. I'll do this myself when merging this.

batonius:
@little-dude It's kinda messy PoC right now, so I haven't been pushing, here it is - https://github.com/batonius/netstack/tree/smolnetd . So far I've implemented the `ip:` scheme enough for `icmpd` and `udpd` to use it and for `dns` and `ping` to work on top of them. `tcpd` is useless for now because it expects the `ip:` scheme to calculate TCP checksum.

Right now it's far from reimplementing all of the `ipd`/`udpd` features - it doesn't support blocking sockets, timeouts for UDP sockets, socket options, it allocates a new buffer on every ingress packet instead of using a pool and it polls socket all the time.

whitequark:
Quick comment so this doesn't sit idle: can you please add a `struct Ipv4Cidr { addr, mask }` and `enum IpCidr { Ipv4(Ipv4Cidr) }`, just like it's done for `struct Ipv4Addr` and `enum IpAddr`? I'd really like smoltcp to gain these abstractions eventually, anyway.

whitequark:
Oh sorry I already wrote this comment and forgot about it! Let me take another look.

batonius:
My first thought was to implement some kind of `Cidr` data type, which is handy to have anyway, but I decided against it because it doesn't fit here conceptually. Specifically, 'mask length' is an attribute of an address assigned to an interface, not of the default route, which is a bare IP address, so it would be wrong to bundle them together in a single `Cidr` value. That means the right approach should probably involve changing `EthernetInterface::new` to accept a slice of `IpCidr`s instead of `IpAdddress`s and require an `IpAddress` of the default gateway as well, but I really didn't want to change public API.

Having said that, it's not critical for a stop-gap solution, so if you think it's ok I can add `IpCidr` and rephrase the PR using it.

whitequark:
> That means the right approach should probably involve changing EthernetInterface::new to accept a slice of IpCidrs instead of IpAdddresss and require an IpAddress of the default gateway as well, but I really didn't want to change public API.

Yeah sure that's what I meant. The public API needs to be changed, and I change it at will anyway, we're at 0.x.

batonius:
Done.

src/iface/ethernet.rs
```diff
@@ -41,24 +42,28 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
     /// # Panics
     /// See the restrictions on [set_hardware_addr](#method.set_hardware_addr)
     /// and [set_protocol_addrs](#method.set_protocol_addrs) functions.
-    pub fn new<DeviceMT, ArpCacheMT, ProtocolAddrsMT>
+    pub fn new<DeviceMT, ArpCacheMT, ProtocolAddrsMT, DefaultGWMT>
```
whitequark: Nit: I think `MT` used to mean something but now it can be just replaced with `T` everywhere here (so `GatewayT`).

src/iface/ethernet.rs
```diff
@@ -23,7 +23,8 @@ pub struct Interface<'a, 'b, 'c, DeviceT: Device + 'a> {
     device:         Managed<'a, DeviceT>,
     arp_cache:      Managed<'b, ArpCache>,
     hardware_addr:  EthernetAddress,
-    protocol_addrs: ManagedSlice<'c, IpAddress>,
+    protocol_addrs: ManagedSlice<'c, IpCidr>,
+    default_gw:     Option<IpAddress>,
```
whitequark: Nit: the phrase "default gateway" is a remnant of the times where a single LAN segment would conceivably have more than one gateway. At this point it's an anachronism (like classful networks!), let's call this just "gateway". Also applies elsewhere in this PR.

src/iface/ethernet.rs
```diff
@@ -82,32 +87,42 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+        self.protocol_addrs.iter().any(|probe| probe.addr() == addr)
+    }
+
+    /// Get the default gateway of the interface.
+    pub fn default_gateway(&self) -> Option<IpAddress> {
+        self.default_gw
+    }
+
+    /// Set the default gateway of the interface.
+    pub fn set_default_gateway<T: Into<Option<IpAddress>>>(&mut self, default_gw: T) {
```
whitequark: I assume there's an `impl Into<Option<T>> for Option<Into<T>>` or something like it?

src/wire/ip.rs
```diff
@@ -126,6 +127,55 @@ impl fmt::Display for Address {
...
 }
 
+/// A CIDR containing a prefix of an IP subnet.
+pub enum Cidr {
+    Ipv4(Ipv4Cidr),
+}
+
+impl Cidr {
+    /// Construct a new CIDR
+    pub fn new(addr: Address, mask_len: u8) -> Cidr {
```
whitequark: This function doesn't seem very useful as the valid ranges of prefix lengths depend on the address.

src/wire/ip.rs
```diff
@@ -126,6 +127,55 @@ impl fmt::Display for Address {
...
+    pub fn contains_ip(&self, addr: Address) -> bool {
+        match (self, addr) {
+            (&Cidr::Ipv4(ref ipv4_cidr), Address::Ipv4(ipv4_addr)) => {
+                ipv4_cidr.contains_ip(ipv4_addr)
+            },
+            _ => false,
+        }
+    }
+
+    /// Return the mask_len of the CIDR.
```
whitequark: Prefix length, not mask length; the mask is as long as the IP address, but has only as many zeroes as the prefix length.

src/wire/ipv4.rs
```diff
@@ -75,6 +75,36 @@ impl fmt::Display for Address {
     }
 }
 
+/// An IPv4 CIDR containing a prefix of a IPv4 subnet.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
+pub struct Cidr {
+    pub addr: Address,
```
whitequark: Nit: I think `address` sounds nicer, and see above on mask_len.

src/wire/ip.rs
```diff
@@ -126,6 +127,55 @@ impl fmt::Display for Address {
...
+        match addr {
+            Address::Ipv4(addr) => Cidr::Ipv4(Ipv4Cidr::new(addr, mask_len)),
+            Address::Unspecified => {
+                panic!("Can't convert an unspecified address to a CIDR")
+            }
+        }
+    }
+
+    /// Check whether the subnet described by the CIDR conains the address.
+    pub fn contains_ip(&self, addr: Address) -> bool {
```
whitequark: Nit: just `contains`.

src/wire/ip.rs
```diff
@@ -126,6 +127,55 @@ impl fmt::Display for Address {
     }
 }
 
+/// A CIDR containing a prefix of an IP subnet.
```
whitequark: I've been uncomfortable calling this data structure "CIDR" for a while but I just realized what the proper name for it would be: `IpSubnet`! That makes much more sense. I'm sorry for having you change everything.

src/wire/ip.rs
```diff
@@ -126,6 +127,55 @@ impl fmt::Display for Address {
...
+
+    /// Return the mask_len of the CIDR.
+    pub fn mask_len(&self) -> u8 {
+        match self {
+            &Cidr::Ipv4(Ipv4Cidr{mask_len, ..}) => mask_len,
+        }
+    }
+
+    /// Return the IP address of the CIDR.
+    pub fn addr(&self) -> Address {
```
whitequark: Nit: just `address`.

src/wire/ipv4.rs
```diff
@@ -75,6 +75,36 @@ impl fmt::Display for Address {
...
+    pub mask_len: u8,
```
whitequark: Since this field is `pub`, the struct needs an `is_valid()` method, which should be checked in `set_protocol_addrs`.

src/iface/ethernet.rs
batonius: Apparently not, I've made it explicit.

src/wire/ip.rs
batonius: Subnets have their own semantics, for example, 192.168.1.0/24 is exactly the same subnet as 192.168.1.255/24, this makes accepting a slice of subnets as interface addresses somewhat confusing. I've changed the documentation to better reflect the fact that CIRD != subnet, but I'd prefer to keep `Cidr` as the name.

src/wire/ip.rs
batonius: I've rephrased it so the constructors check prefix_len and can fail now.

src/wire/ip.rs
batonius: I was thinking about adding `contains_subnet` as well.

src/wire/ipv4.rs
```diff
@@ -75,6 +75,55 @@ impl fmt::Display for Address {
     }
 }
 
+/// An IPv4 CIDR containing an IPv4 address with a prefix lenght.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
+pub struct Cidr {
+    address: Address,
```
whitequark: This is a really odd case, because so far every other data structure in `wire` has all fields `pub`. It's even in the module documentation. Thoughts?

batonius: The idea was to prevent the creation of inconsistent values, instead of using `is_valid` method, since CIDRs have an invariant that needs to be enforced.

batonius:
Added `FromStr` implementation for `IpCidr`.

batonius:
Rebased the branch.

Btw, the current state of smoltcp on Redox:
icmp_udp_tcp
Besides this PR, I had no serious issues with reimplementing the current stack so far.

whitequark:
Cool!

batonius:
Done.

whitequark:
Merged manually with some changes:
  * Reformatted, renamed and reworded a few things for consistency.
  * `IpCidr`/`Ipv4Cidr` now always panic on error, no `Result`s,
  * Pass everything by reference in `IpCidr`/`Ipv4Cidr`. The former in particular will become much larger with introduction of IPv6 addresses.
  * Cleanup around the handling of `Option` in the default gateway code.
  * Cleanup of routing logic.
  * Updated README.

PR #45, opened 2017-09-20
Uncomment associated constants

batonius:
Associated constants were stabilized in 1.20: https://blog.rust-lang.org/2017/08/31/Rust-1.20.html .

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ include complicated compile-time computations, such as macro or type tricks, eve
 at cost of performance degradation.
 
 _smoltcp_ does not need heap allocation *at all*, is [extensively documented][docs],
-and compiles on stable Rust 1.19 and later.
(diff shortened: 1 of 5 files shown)


whitequark:
We say in README:

> smoltcp does not need heap allocation at all, is extensively documented, and compiles on stable Rust 1.19 and later.

This needs to be updated.

batonius:
Done.

whitequark:
Note: I'm going to hold off on merging this for a bit because m-labs has a fork of rustc that's stuck on 1.19, and there are higher priority tasks than updating rustc. But it shouldn't take too long, anyway.

batonius:
Sure, it's just something I've noticed while looking for something else.

whitequark:
(We've updated our internal fork.)

PR #60, opened 2017-10-16
IGMPv2

podhrmic:
Working example of IGMPv2 protocol.

Short summary:
 - subscribing to and leaving from multicast groups is now handled by the application layer (similar to Linux `mreq` struct and `sockopt()`)
- smoltcp only automatically responds to group membership requests (and at this point in a limited way - see comments in the code)
- I am providing a multicast example to illustrate how this could be used
- decided to enable IGMP by default since it is an important protocol (although it should be possible to disable it)
- still uncertain how to handle timers and how to propagate the states (i.e. when group subscription changes). Ideally this would be handled automatically based on what address is a socket bound to similarly to [join_multicast()](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.join_multicast_v4)

Caveats:
- missing framework for managing timers (like when a membership report has to be sent at some time in the future rather than immediately)
- I know about the hash set back end, I would like to replace it with some generalized Cache based on a ManagedSlice
- the TTL value for group query has to be set to 1, and to 255 for membership report (according to Wireshark and RFC 3717)

Other protocol implementations:
- see for example [picotcp](https://github.com/tass-belgium/picotcp/blob/master/modules/pico_igmp.c) (for IGMPv3)

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -11,7 +11,7 @@ use wire::{IpAddress, IpProtocol, IpRepr, IpCidr};
 use wire::{ArpPacket, ArpRepr, ArpOperation};
 use wire::{Ipv4Packet, Ipv4Repr};
 use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
-#[cfg(feature = "protocol-igmp")] use wire::{IgmpPacket, IgmpRepr, IgmpReportVersion};
+use wire::{IgmpPacket, IgmpRepr, IgmpReportVersion};
 #[cfg(feature = "socket-udp")] use wire::{UdpPacket, UdpRepr};
 #[cfg(feature = "socket-tcp")] use wire::{TcpPacket, TcpRepr, TcpControl};
 use socket::{Socket, SocketSet, AnySocket};
@@ -31,15 +31,14 @@ pub struct Interface<'a, 'b, 'c, DeviceT: Device + 'a> {
     ethernet_addr:  EthernetAddress,
     ip_addrs:       ManagedSlice<'c, IpCidr>,
     ipv4_gateway:   Option<Ipv4Address>,
-    eth_mcast_addr: HashSet<EthernetAddress>, // TODO: use ManagedMap
-    ip_mcast_addr: HashSet<IpAddress>, // TODO: use ManagedMap
+    eth_mcast_addr: HashSet<EthernetAddress>, // TODO: use Managed
+    ip_mcast_addr: HashSet<IpAddress>, // TODO: use Managed
 }
 
 enum Packet<'a> {
     None,
     Arp(ArpRepr),
     Icmpv4(Ipv4Repr, Icmpv4Repr<'a>),
-    #[cfg(feature = "protocol-igmp")]
     Igmp(Ipv4Repr, IgmpRepr),
     #[cfg(feature = "socket-raw")]
     Raw((IpRepr, &'a [u8])),
@@ -49,23 +48,25 @@ enum Packet<'a> {
     Tcp((IpRepr, TcpRepr<'a>))
 }
 
+
+
 /// Map IPv4 multicast address into an ethernet addres
 pub fn ip_multicast_to_mac(ip_addr: IpAddress) -> Option<EthernetAddress> {
     if !ip_addr.is_multicast() {
         return None;
     } else {
-    	match ip_addr {
-    		IpAddress::Unspecified => None,
-    		IpAddress::Ipv4(addr)  => {
-    			let mut hw_addr = EthernetAddress::MULTICAST_PREFIX;
-    			// first three octets are fixed
-			    hw_addr[3] = addr.as_bytes()[1] & 0x7f; // 4th octet, first zero fixed
-			    hw_addr[4] = addr.as_bytes()[2]; // 5th octet
-			    hw_addr[5] = addr.as_bytes()[3]; // last octet
-			    Some(EthernetAddress::from_bytes(&hw_addr))
-    		}
-    		IpAddress::__Nonexhaustive => unreachable!(),
-    	}
+         match ip_addr {
+            IpAddress::Unspecified => None,
+            IpAddress::Ipv4(addr)  => {
+                let mut hw_addr = EthernetAddress::MULTICAST_PREFIX;
+                // first three octets are fixed
+                hw_addr[3] = addr.as_bytes()[1] & 0x7f; // 4th octet, first zero fixed
+                hw_addr[4] = addr.as_bytes()[2]; // 5th octet
+                hw_addr[5] = addr.as_bytes()[3]; // last octet
+                Some(EthernetAddress::from_bytes(&hw_addr))
+            }
+            IpAddress::__Nonexhaustive => unreachable!(),
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


podhrmic:
Of course the build fails now because it depends on `std`.

src/iface/ethernet.rs
```diff
@@ -44,6 +48,28 @@ enum Packet<'a> {
...
+pub fn ip_multicast_to_mac(ip_addr: IpAddress) -> Option<EthernetAddress> {
+    if !ip_addr.is_multicast() {
+        return None;
+    } else {
+         match ip_addr {
+            IpAddress::Unspecified => None,
+            IpAddress::Ipv4(addr)  => {
+                let mut hw_addr = EthernetAddress::MULTICAST_PREFIX;
+                // first three octets are fixed
+                hw_addr[3] = addr.as_bytes()[1] & 0x7f; // 4th octet, first zero fixed
```
whitequark: `addr.1` also works.

src/iface/ethernet.rs
```diff
@@ -233,16 +323,30 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
 
         Ok(())
     }
+    
+    fn keep_ethernet_frame<'frame, T: AsRef<[u8]>>(&self, eth_frame: &EthernetFrame<&T>) -> bool {
+      // the multicast frame belongs to one of the subscribed groups
+      if self.eth_mcast_addr.contains(&eth_frame.dst_addr()) {
```
whitequark: Maybe check if `eth_frame.dst_addr()` is multicast before trying to do a hash lookup?

src/iface/ethernet.rs
```diff
@@ -336,15 +441,18 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
             }
         }
 
-        if !self.has_ip_addr(ipv4_repr.dst_addr) {
-            // Ignore IP packets not directed at us.
+        if !self.has_ip_addr(ipv4_repr.dst_addr) && !self.has_ip_mcast_addr(ipv4_repr.dst_addr) {
```
whitequark: Should probably be extracted into a function like `keep_ethernet_frame`; we also currently ignore all IPv4 broadcast packets, which is I think a bug.

src/lib.rs
```diff
@@ -124,11 +124,13 @@ pub enum Error {
     Fragmented,
     /// An incoming packet was recognized but was self-contradictory.
     /// E.g. a TCP packet with both SYN and FIN flags set.
+    /// Or a IGMP packet querying address that is not multicast
```
whitequark: The above is just an example, you don't need to add every conceivable error condition.

src/lib.rs
```diff
@@ -124,11 +124,13 @@ pub enum Error {
...
     Malformed,
     /// An incoming packet was recognized but contradicted internal state.
     /// E.g. a TCP packet addressed to a socket that doesn't exist.
     Dropped,
-
+	/// An unspecified IO error on the raw socket
+	IOError,
```
whitequark: This is completely unrelated to IGMP, and should not exist in its current form anyhow.

src/phy/raw_socket.rs
```diff
@@ -2,16 +2,17 @@ use std::cell::RefCell;
 use std::vec::Vec;
```
whitequark: Please don't reformat files, this makes it hard to review diffs.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,403 @@
...
+use {Error, Result};
+use phy::ChecksumCapabilities;
+use super::ip::checksum;
+
+use wire::Ipv4Address;
+
+enum_with_unknown! {
+    /// Internet Group Management Protocol v2 message type.
+    pub doc enum Message(u8) {
+    /// Membership Query
```
whitequark: Missing indent. Also, I sort enums by their numeric value.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,403 @@
...
+    pub const MAX_RESP_TIME: usize = 1;
+    pub const CHECKSUM: Field = 2..4;
+    pub const GROUP_ADDRESS: Field = 4..8;
+}
+
+impl fmt::Display for Message {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            &Message::MembershipQuery => write!(f, "membership query"),
+            &Message::MembershipReportV2 => write!(f, "version 2 membership report "),
```
whitequark: Trailing space in the string

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    /// Ensure that no accessor method will panic if called.
+    /// Returns `Err(Error::Truncated)` if the buffer is too short.
+    pub fn check_len(&self) -> Result<()> {
+        let len = self.buffer.as_ref().len();
+        if len < field::CHECKSUM.end {
+            Err(Error::Truncated)
+        } else {
+            if len < field::GROUP_ADDRESS.end as usize {
```
whitequark: You can just leave this one check, it implies `len < field::CHECKSUM.end`.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// Set the group address field
+    #[inline]
+    pub fn set_group_address(&mut self, addr: Ipv4Address) {
+        let data = self.buffer.as_mut();
+        let addr_bytes = addr.as_bytes();
+        // TODO: check for host endiannes?
+        let addr_u32: u32 = (addr_bytes[3] as u32) | (addr_bytes[2] as u32) << 8 |
+                            (addr_bytes[1] as u32) << 16 |
+                            (addr_bytes[0] as u32) << 24;
+        NetworkEndian::write_u32(&mut data[field::GROUP_ADDRESS], addr_u32)
```
whitequark: `data[field::GROUP_ADDRESS].copy_from_slice(addr.as_bytes())`

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,403 @@
...
+/// A high-level representation of an Internet Group Management Protocol v2 header.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Repr {
+    MembershipQuery {
+        max_resp_time: u8,
+        group_addr: Ipv4Address,
+    },
+    MembershipReport {
+        group_addr: Ipv4Address,
+        version: ReportVersion,
```
whitequark: Is there any reason to care about report version? Will you ever emit V1 report packets? If no and no, you should drop this field.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,403 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+
+    static LEAVE_PACKET_BYTES: [u8; 8] = [0x17, 0x00, 0x02, 0x69, 0xe0, 0x00, 0x06, 0x96];
+    static REPORT_PACKET_BYTES: [u8; 8] = [0x16, 0x00, 0x08, 0xda, 0xe1, 0x00, 0x00, 0x25];
+
+    static IPV4_LEAVE_PACKET: [u8; 32] = [0x46, 0xc0, 0x00, 0x20, 0x00, 0x00, 0x40, 0x00, 0x01,
```
whitequark: This is unused.

podhrmic:
@whitequark sorry I didn't get back to this for a while - I had to focus my attention elsewhere. I will revisit this PR in January, I am excited to get IGMP merged.

whitequark:
@podhrmic Ping?

podhrmic:
Yep, sorry for the delay. I am working now on adding a [seL4](https://sel4.systems/) backend for smoltcp so that should be an interesting addition. Then I should be able to get back to this. But if you want to clean up the stale PR's, I am ok with closing it (the list of merge conflicts is getting longer) and then making a new PR later. Thoughts?

whitequark:
Sure.

PR #61, opened 2017-10-19
Tests: Add basic interface tests

dlrobertson:
Was looking at how the `EthernetInterface` worked and wrote some basic unit tests. Feel free to close this if you feel these tests aren't that useful.

 - Add tests for the following
   - ICMP error responses are not sent in response to broadcast requests (this
     isn't super relevant now, but this ensures we don't cause broadcast storms
     once we consume broadcast packets).
   - ARP requests are responded to and inserted into the cache
   - ARP requests for someone else are not responded to, but the sender
     is still inserted in the cache

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -32,6 +32,7 @@ pub struct Interface<'a, 'b, 'c, DeviceT: Device + 'a> {
     ipv4_gateway:   Option<Ipv4Address>,
 }
 
+#[derive(Debug, PartialEq)]
 enum Packet<'a> {
     None,
     Arp(ArpRepr),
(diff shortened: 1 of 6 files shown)


src/iface/ethernet.rs
```diff
@@ -637,3 +637,168 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+            0x08, 0x00,
+            // Ipv4 Header
+            0x45, 0x00, 0x00, 0x14,
+            0x00, 0x00, 0x00, 0x00,
+            0x40, 0x0c, 0xfb, 0xdd,
+            0x7f, 0x00, 0x00, 0x01,
+            0xff, 0xff, 0xff, 0xff,
+        ];
+
+        // Create a basic device
```
whitequark: This common initialization code can be taken out of the tests and into a setup function.

src/iface/ethernet.rs
```diff
@@ -637,3 +637,168 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+        let mut socket_set_entries: [_; 1] = Default::default();
+        let mut socket_set = SocketSet::new(&mut socket_set_entries[..]);
+
+        // Create a frame from the bytes
+        let frame = EthernetFrame::new(&bytes);
+
+        // Ensure that the unknown protocol frame does not trigger an
+        // ICMP error response when the destination address is a
+        // broadcast address
+        match iface.process_ipv4(&mut socket_set, 0, &frame).unwrap() {
```
whitequark: A better way to express this would be:
```
assert_eq!(iface.process_ipv4(&mut socket_set, 0, &frame),
           Ok(Packet::None));
```

src/iface/ethernet.rs
```diff
@@ -637,3 +637,168 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+
+        let mut socket_set_entries: [_; 1] = Default::default();
+        let mut socket_set = SocketSet::new(&mut socket_set_entries[..]);
+
+        let local_addr = Ipv4Address([0x7f, 0x00, 0x00, 0x01]);
+        let remote_addr = Ipv4Address([0x7f, 0x00, 0x00, 0x02]);
+
+        // Ensure an ARP Request for us triggers an ARP Reply
+        match iface.process_ethernet(&mut socket_set, 0, &bytes).unwrap() {
+            Packet::None => { panic!("Respond to ARP Requests") },
```
whitequark: Same as above.

src/iface/ethernet.rs
```diff
@@ -637,3 +637,168 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+        let mut iface = EthernetInterface::new(
+            &mut device, &mut arp_cache as &mut ArpCache,
+            EthernetAddress::default(), &mut ip_addrs[..], None);
+
+        let mut socket_set_entries: [_; 1] = Default::default();
+        let mut socket_set = SocketSet::new(&mut socket_set_entries[..]);
+
+        // Ensure an ARP Request for someone else does not trigger an ARP Reply
+        match iface.process_ethernet(&mut socket_set, 0, &bytes).unwrap() {
+            Packet::None => {},
```
whitequark: Same as above.

src/iface/ethernet.rs
```diff
@@ -637,3 +637,168 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+    #[test]
+    fn no_icmp_to_broadcast() {
+        // Unknown Ipv4 Protocol
+        //
+        // Because the destination is the broadcast address
+        // this should not trigger and Destination Unreachable
+        // response. See RFC 1122 § 3.2.2.
+        let bytes = vec![
+            // Ethernet Frame
+            0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
```
whitequark: Since we are not testing the packet parsing/construction routines here, do you think you could use `EthernetFrame`/`IpRepr`/`ArpRepr`/etc to construct the bytewise representations? That would make tests much easier to understand and write.

dlrobertson: This would definitely be feasible for the ARP tests. I'm not sure about how feasible it would be for this test since the protocol type is a protocol that isn't in `IpProtocol`.

src/iface/ethernet.rs
dlrobertson: Good point

src/iface/ethernet.rs
dlrobertson: :+1:

src/iface/ethernet.rs
dlrobertson: :+1:

src/iface/ethernet.rs
whitequark: `IpProtocol::Unknown(0xwhatever)`

src/iface/ethernet.rs
```diff
@@ -32,6 +32,7 @@ pub struct Interface<'a, 'b, 'c, DeviceT: Device + 'a> {
     ipv4_gateway:   Option<Ipv4Address>,
 }
 
+#[derive(Debug, PartialEq)]
```
dlrobertson: Added these so that `assert_eq!` would work

src/iface/ethernet.rs
```diff
@@ -644,3 +645,157 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+                ttl:         0x40
+            });
+
+            let frame = {
+                let mut frame = EthernetFrame::new(&mut eth_bytes);
+                frame.set_dst_addr(EthernetAddress::BROADCAST);
+                frame.set_src_addr(EthernetAddress([0x52, 0x54, 0x00, 0x00, 0x00, 0x00]));
+                frame.set_ethertype(EthernetProtocol::Ipv4);
+                repr.emit(frame.payload_mut(), &ChecksumCapabilities::default());
+                EthernetFrame::new(&*frame.into_inner())
```
dlrobertson: Kind of ugly, but inner type mutability differs without this

whitequark: That's a perfectly fine construct in smoltcp.

src/iface/ethernet.rs
```diff
@@ -644,3 +645,157 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+    use wire::{Ipv4Address, Ipv4Repr};
+    use iface::{ArpCache, SliceArpCache, EthernetInterface};
+    use socket::SocketSet;
+
+    fn loopback_test<F>(f: F)
+            where F: FnOnce(&mut EthernetInterface<Loopback>) -> () {
+        // Create a basic device
+        let mut device = Loopback::new();
+
+        let mut arp_cache_entries: [_; 8] = Default::default();
```
whitequark: You can box everything here and then you can just return EthernetInterface by value.

src/iface/ethernet.rs
```diff
@@ -644,3 +645,157 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+        let mut arp_cache = SliceArpCache::new(&mut arp_cache_entries[..]);
+
+        let mut ip_addrs = [IpCidr::new(IpAddress::v4(127, 0, 0, 1), 8)];
```
whitequark: Same here.

src/iface/ethernet.rs
```diff
@@ -644,3 +645,157 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
+                let mut frame = EthernetFrame::new(&mut eth_bytes);
+                frame.set_dst_addr(EthernetAddress::BROADCAST);
+                frame.set_src_addr(EthernetAddress([0x52, 0x54, 0x00, 0x00, 0x00, 0x00]));
+                frame.set_ethertype(EthernetProtocol::Ipv4);
+                repr.emit(frame.payload_mut(), &ChecksumCapabilities::default());
+                EthernetFrame::new(&*frame.into_inner())
+            };
+
+            let mut socket_set_entries: [_; 1] = Default::default();
+            let mut socket_set = SocketSet::new(&mut socket_set_entries[..]);
```
whitequark: Same here.

src/iface/ethernet.rs
dlrobertson: Good point. It's a bit more readable that way

Issue #62, opened 2017-10-25
TCP hangs when sending


pothos:
Hello,
actually I wanted to debug it, but need some help because it is disappears e.g. if writing to pcap or extensive logging is used.

The smoltcp server thread on tap0 will send in 64 byte steps, a Linux socket client thread will read only to ~292800 bytes instead of reading all 10000000 bytes. Smoltcp is handing out keep-alives from this point on. The minimal example is here:

https://github.com/pothos/taptest

`cargo run --release` should reproduce the behavior.

`RUST_LOG=trace cargo run --release` should let it _sometimes_ complete (same as adding print statements in the sending server).
Uncommenting the pcap writer will let it complete reliably.

I have experienced it with other values as well, but hope that this combination also reproduces it on other machines.

Best regards,
Kai

whitequark:
Oh yeah you hit a bug; retransmit timer gets reset by ACKs whereas it should not.

pothos:
Thanks for looking at it, now I have to read the RFCs along with the code. Because I don't understand yet why only `ack_len` is used to dequeue and a change in the window size is not relevant(?). Rightnow I don't see where the retransmit would be involved… I'll continue tomorrow.

pothos:
Ok, finally got it working. Besides keeping the retransmission timer there was an overflow that needed to be fixed. Preparing a PR now.

whitequark:
@pothos I'm finally making progress on this—see d1e2292 for one method I'm using to expose underlying bugs. I am afraid that your fix is hiding one of them.

whitequark:
@pothos I've integrated a test quite like your `taptest` in the commit 44db954. I've managed to expose one other bug as well.

whitequark:
Fixed in b247f64.

pothos:
Thanks for including the fix - I am happy that the wrapping can be covered without distinction of cases ;)

PR #65, opened 2017-10-28
Fix retransmission and send offset overflow

pothos:
Resolves https://github.com/m-labs/smoltcp/issues/62

Duplicate ACKs should not replace the retransmission timer
    but otherwise, as normal ACKs, set the keep-alive timer.

During the retransmission, higher sequence numbers than the
    last one sent could be acknowledged. This led to an overflow
    in the offset calculation, which resulted in no further data
    being sent. Now the maximum is used to correctly set it to 0.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1115,9 +1115,29 @@ impl<'a> TcpSocket<'a> {
         }
 
         if let Some(ack_number) = repr.ack_number {
+            let old_local_seq_no = self.local_seq_no;
             // We've processed everything in the incoming segment, so advance the local
             // sequence number past it.
             self.local_seq_no = ack_number;
+            // During retransmission, a higher sequence number can get acknowledged than
+            // and has to be used for further sending instead of the last calculated after
+            // retransmitting an older packet. Thus have to forward remote_last_seq to the
+            // updated local sequence number here.
+            // In normal operation, nothing should be done here.
+            // Due to wrapping of the sequence numbers it is not enough to check
+            // the two cases of local_seq_no < remote_last_seq to see if it is a
+            // retransmission or not, but instead the two wrapped cases also have to be
+            // considered. To distinguish them old_local_seq_no is needed:
+            // These two cases describe retransmission where remote_last_seq lags behind
+            // local_seq_no and old_local_seq_no lies not inbetween (because that would
+            // be the case when for wrapped non-retransmission) or lies inbetween if
+            // local_seq_no just got wrapped (it would not be inbetween for non-wrapped
+            // no-retransmission sending).
+            if (self.remote_last_seq < self.local_seq_no &&
+                !(self.remote_last_seq < old_local_seq_no && old_local_seq_no < self.local_seq_no)) ||
+               (self.local_seq_no < old_local_seq_no && old_local_seq_no < self.remote_last_seq) {
+                self.remote_last_seq = self.local_seq_no;
+            }
         }
 
         let payload_len = repr.payload.len();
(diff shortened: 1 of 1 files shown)


whitequark:
Thanks. Can you add tests?

pothos:
Good idea, there was one more problem I could spot through the tests: When it continued to send data, it was using a wrong (too low) sequence number.
Also in one debug print there was an overflow.
So I decided to just update the value correctly at the place where the local sequence number is forwarded. Is it ok now?

src/wire/tcp.rs
```diff
@@ -10,7 +10,7 @@ use super::ip::checksum;
 ///
 /// A sequence number is a monotonically advancing integer modulo 2<sup>32</sup>.
 /// Sequence numbers do not have a discontiguity when compared pairwise across a signed overflow.
-#[derive(Debug, PartialEq, Eq, Clone, Copy, Default)]
+#[derive(Debug, PartialEq, Eq, Ord, Clone, Copy, Default)]
```
whitequark: No no no, this is very intentional! Sequence numbers are **not** totally ordered! They wrap!

whitequark:
I'll need to take a closer look at these changes. I'm not yet completely convinced they're correct. But, they do seem to go in the right direction.

src/wire/tcp.rs
pothos: Ah, yes. Then by just looking at the two numbers there is no way to distinguish between the case of normal operation with local_seq_no just wrapped before the max statement and the retransmission case. There is also a wrapped case for retransmission where the minimum would have to be taken. So first a way to find out if the value of  remote_last_seq was coming from a retransmission is needed, then it could be updated directly and there is no min/max involved.

pothos:
This is roughly how I would solve it now. Instead of the casts it could use PartialOrd::partial_cmp.
Just considering the two values local_seq_no (newly set) and  remote_last_seq and the two cases normal vs retransmission there are four combinations because of wrapping (i.e. ):

(N1) Normal with local_seq_no < remote_last_seq and (N2) normal with remote_last_seq < local_seq_no. (Wrapping happened already for remote_last_seq in N2.)

(R1) Retransmission with remote_last_seq < local_seq_no and (R2) retransmission with local_seq_no < remote_last_seq. (Wrapping happend just now for local_seq_no in R2)

Looking at old_local_seq_no helps to distinguish them, because in the wrapped cases it is guaranteed to be in the interval of the other two numbers while in the non-wrapped cases the opposite holds.

It looks ugly rightnow and needs to presented in a better way. I think it is correct, would you agree that handling these cases is an acceptable approach?

pothos:
Maybe this is now a bit more readable? I have included the wrapping cases in the test. In wire.rs I had to make the two wrapping subtractions explicit because otherwise sending gave panics in debug mode.

src/socket/tcp.rs
```diff
@@ -1110,9 +1122,36 @@ impl<'a> TcpSocket<'a> {
...
+                // Second case for normal operation where remote_last_seq is already wrapped.
+            } else if (self.remote_last_seq < self.local_seq_no &&
+                       !(self.remote_last_seq < old_local_seq_no && old_local_seq_no < self.local_seq_no)) ||
+                       // Like the second case but old_loc_seq_no is at a different position.
+                       // Next like the first case but again because local_seq_no is wrapped,
+                       // but old_loc_seq_no is at a different position.
+                      (self.local_seq_no < old_local_seq_no && old_local_seq_no < self.remote_last_seq) {
+                // These two cases describe retransmission where remote_last_seq lags behind local_seq_no.
+                // The value for the next sequence number to be used is updated.
+                self.remote_last_seq = self.local_seq_no;
```
pothos: I can remove all empty cases and the unreachable which I added to see if something was missed. Then if would just be the following:

```
if (self.remote_last_seq < self.local_seq_no &&
   !(self.remote_last_seq < old_local_seq_no && old_local_seq_no < self.local_seq_no)) ||
  (self.local_seq_no < old_local_seq_no && old_local_seq_no < self.remote_last_seq) {
    self.remote_last_seq = self.local_seq_no;
}
```

src/socket/tcp.rs
```diff
@@ -1025,9 +1025,18 @@ impl<'a> TcpSocket<'a> {
                 self.timer.set_for_idle(timestamp, self.keep_alive);
             }
 
-            // ACK packets in ESTABLISHED state reset the retransmit timer.
+            // ACK packets in ESTABLISHED state reset the retransmit timer,
+            // except for duplicate ACK packets.
             (State::Established, TcpControl::None) => {
-                self.timer.set_for_idle(timestamp, self.keep_alive);
+                match self.timer {
```
whitequark: This seems obviously correct and OK to merge. Do you think you can extract this part and a dedicated test, and submit a PR for it?

Also, would you mind rewriting the condition a bit? I would add a `Timer::is_retransmit() -> bool` function, and write the condition as `if !timer.is_retransmit() || ack_len != 0 { self.timer.set_for_idle(...) }`

src/socket/tcp.rs
whitequark: I am sure that you have discovered a real issue, but I do not seem to understand your explanation. Do you think you could expand on it or rephrase it in a different way?

src/wire/tcp.rs
```diff
@@ -51,13 +51,13 @@ impl ops::Sub for SeqNumber {
     type Output = usize;
 
     fn sub(self, rhs: SeqNumber) -> usize {
-        (self.0 - rhs.0) as usize
+        self.0.wrapping_sub(rhs.0) as usize
```
whitequark: Can you add a test for this?

pothos:
Hello,
thanks a lot for the feedback. I've split this commit into two again, first for the regular overflow (test case in socket/tcp.rs because it is just a test for normal send operations which depends on the explicit wrapping), then for the special retransmission wrong-seq-number/overflow bug when the value is not updated.
Regards,
Kai

src/socket/tcp.rs
```diff
@@ -3062,6 +3082,71 @@ mod test {
         }));
     }
 
+    #[test]
+    fn test_established_retransmit_ack_higher() {
+        // Generate cases where local_seq_no and remote_last_seq wrap when
+        // ACKs are received, both for regular and retransmitted data.
+        for local_seq_start in vec![LOCAL_SEQ, TcpSeqNumber(i32::MAX - 7),
+            TcpSeqNumber(i32::MAX - 6), TcpSeqNumber(i32::MAX - 5), TcpSeqNumber(i32::MAX - 4),
+            TcpSeqNumber(i32::MAX - 3), TcpSeqNumber(i32::MAX - 2)] {
```
whitequark: I don't understand how wrapping is relevant here. TCP sequence numbers are compared modulo 2³². Specifically, `TcpSeqNumber(i32::MAX - 6) - TcpSeqNumber(i32::MAX - 3)` equals `TcpSeqNumber(-6) - TcpSeqNumber(-3)`.

Can you show me a test case that results in an error only when the sequence number wraps? If one exists, it means there is a bug somewhere else apart from the retransmission code.

pothos: The overflow assertions are fixed in the other commit with an extra test: https://github.com/m-labs/smoltcp/pull/65/commits/3480238359d1d26c7c570af471073049ade002b9 (but are not shown by github anymore since the master merge).
Here the test also includes wrapping and non-wrapping cases to cover the constellations of `local_seq_no`, `old_local_seq_no` and `remote_last_seq`.
When the ACK arrives after the first packet retransmitted for MAX-7 we have: `old_local_seq_no < remote_last_seq < local_seq_no`
For MAX-6 `local_seq_no` already wraps to: `local_seq_no < old_local_seq_no < remote_last_seq`
For MAX-3 `local_seq_no` and `remote_last_seq` now wrapped to: `remote_last_seq < local_seq_no < old_local_seq_no`
In all these cases the value of `remote_last_seq` will be updated.
The last send macro sends an ACK for a normal transmission and the idea was that also the constellations are permuted for this condition where no update should happen. But I just realized that `tx_buffer.len() == 0` is not enough to ensure that this is the case, so I can add some more data to be transfered.

pothos: Ah, now saw the picked commit, thanks.

pothos: Expanded the test coverage for normal transmission where remote_last_seq should not be touched.

pothos: (The example project mentioned in the issue can also be used to reproduce this, but it does not check the contents sent as the test cases do: https://github.com/pothos/taptest)

whitequark:
You have correctly identified the root cause of the problem however your patch is overly complicated. There is no need to test for wraparound since that is already handled by the `TcpSeqNumber` newtype. You can verify that your testcase works on master. I've merged a simpler version of both your fix and your test as b247f64.

PR #66, opened 2017-10-30
Allow receiving broadcast packets in UDP

jackpot51:
This allows the reception of UDP broadcast packets. They will fall through to the process_udp function. The tcp and icmp functions also receive potential broadcast packets, but they should already be discarded.


dlrobertson:
The destination address is not checked to see if it is the broadcast address before sending ICMP error responses currently (RFC 1122 § 3.2.2). Currently this conditional is the only thing keeping us from causing a broadcast storm due to a weird broadcast packet. The specific panic is [here](https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs#L369), and is due to the packet created for the [`no_icmp_to_broadcast`](https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs#L675-L707) test having no payload.

whitequark:
@dlrobertson Wow, your tests already paid off. Thanks a lot!

jackpot51:
@whitequark, I believe @dlrobertson is working on a fix, to prevent the ICMP error message from being sent

dlrobertson:
After a rebase, tests should pass here.

whitequark:
@jackpot51 Can you rebase?

jackpot51:
Yep, will do

jackpot51:
@whitequark it is done

whitequark:
Any chance you can write a test for this?

jackpot51:
Sure!

jackpot51:
@whitequark I do not see any current tests for UDP in the ethernet interface, am I mistaken?

dlrobertson:
@jackpot51 https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs#L828-L911 is an example of a test in the ethernet interface for UDP. Essentially you could do the same thing as the second half of that test, but add a socket to the socket set that is listening to the port of interest.

jackpot51:
Ok, cool

whitequark:
@jackpot51 ping?

jackpot51:
Sorry @whitequark, I have been busy porting cargo.

dlrobertson:
I added tests with redox-os/smoltcp#1

jackpot51:
@whitequark the tests from @dlrobertson are now merged

PR #67, opened 2017-10-30
Do not send ICMPv4 responses to broadcasts

dlrobertson:
- Do not send ICMPv4 responses for packets with a broadcast destination
   address.
   - Do not send DstUnreachable with ProtoUnreachable on receipt of a
     packet with an unknown protocol with a non-unicast destination
     address.
   - Do not send DstUnreachable with PortUnreachable on receipt of a
     UDP packet when no sockets are listening on the destination port
     and the destination address is a non-unicast address.
 - Send the correct amount of the original datagram when sending Destination
   Unreachable error responses.
   - Do not assume that a ip datagram has a payload when sending a proto
     unreachable ICMPv4 error response.
 - Add tests to iface tests.
   - Ensure ICMP error responses are correctly formed when the
     datagram has no payload.
   - Ensure ICMP error responses are correctly handled for UDP packets
     when no socket is listening on the destination port.
   - Ensure the correct amount of the original payload is returned in
     Destination Unreachable responses.

Related to: #66

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -1,6 +1,7 @@
 // Heads up! Before working on this file you should read the parts
 // of RFC 1122 that discuss Ethernet, ARP and IP.
 
+use core::cmp;
 use managed::{Managed, ManagedSlice};
 
 use {Error, Result};
@@ -158,6 +159,26 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
             Ok(Some(0))
         } else {
             Ok(sockets.iter().filter_map(|socket| socket.poll_at()).min())
+        }
+    }
+
+    fn icmpv4_reply<'frame, 'icmp: 'frame>(&self,
+                                           ipv4_repr: Ipv4Repr,
+                                           icmp_repr: Icmpv4Repr<'icmp>)
+            -> Packet<'frame> {
+        if ipv4_repr.dst_addr.is_unicast() {
+            let ipv4_reply_repr = Ipv4Repr {
+                src_addr:    ipv4_repr.dst_addr,
+                dst_addr:    ipv4_repr.src_addr,
+                protocol:    IpProtocol::Icmp,
+                payload_len: icmp_repr.buffer_len(),
+                ttl:         64
+            };
+            Packet::Icmpv4(ipv4_reply_repr, icmp_repr)
+        } else {
+            // Do not send Protocol Unreachable ICMPv4 error responses to datagrams
+            // with a broadcast destination address.
+            Packet::None
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/iface/ethernet.rs
```diff
@@ -19,6 +20,8 @@ use socket::{Socket, SocketSet, AnySocket};
 #[cfg(feature = "proto-tcp")] use socket::TcpSocket;
 use super::ArpCache;
 
+const SIZEOF_ICMP_ERR_PAYLOAD: usize = 44;
```
dlrobertson: Does something like this belong in `wire::icmpv4`?

src/iface/ethernet.rs
```diff
@@ -362,11 +365,15 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
             _ if handled_by_raw_socket =>
                 Ok(Packet::None),
 
-            _ => {
+            _ if eth_frame.dst_addr().is_unicast() && ipv4_repr.dst_addr.is_unicast() => {
+                // If the datagram does not have a payload greater than or equal
+                // to 64 bytes minus the size of the ipv4 header in length, only
+                // send back the size of given payload.
+                let payload_len = min(ip_payload.len(), SIZEOF_ICMP_ERR_PAYLOAD);
```
dlrobertson: According to RFC 792 up to 64 bytes of the original payload should be included. Since the header is separate, that is `64 - sizeof(ip header)`.

src/iface/ethernet.rs
```diff
@@ -1,6 +1,7 @@
 // Heads up! Before working on this file you should read the parts
 // of RFC 1122 that discuss Ethernet, ARP and IP.
 
+use core::cmp::min;
```
whitequark: `use core::cmp` and then `cmp::min(...)`.

src/iface/ethernet.rs
```diff
@@ -362,11 +365,14 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
             _ if handled_by_raw_socket =>
                 Ok(Packet::None),
 
-            _ => {
+            _ if eth_frame.dst_addr().is_unicast() && ipv4_repr.dst_addr.is_unicast() => {
+                // If the datagram does not have a payload greater than or equal
+                // to 64 bits, only send back the size of given payload.
+                let payload_len = min(ip_payload.len(), SIZEOF_ICMP_ERR_PAYLOAD);
```
whitequark: Just push the entire payload there, up to MTU. It's more useful when debugging networking issues anyway, and it's free for us because we don't allocate.

src/iface/ethernet.rs
```diff
@@ -389,19 +399,25 @@ impl<'a, 'b, 'c, DeviceT: Device + 'a> Interface<'a, 'b, 'c, DeviceT> {
...
-                };
-                let ipv4_reply_repr = Ipv4Repr {
-                    src_addr:    ipv4_repr.dst_addr,
-                    dst_addr:    ipv4_repr.src_addr,
-                    protocol:    IpProtocol::Icmp,
-                    payload_len: icmp_reply_repr.buffer_len(),
-                    ttl:         64
-                };
-                Ok(Packet::Icmpv4(ipv4_reply_repr, icmp_reply_repr))
+                if ipv4_repr.dst_addr.is_unicast() {
```
whitequark: This code *asks* to be refactored into something like `icmpv4_reply` that would make the `Ipv4Repr` with the given `Icmpv4Repr`, and also have the condition for unicast addresses.

dlrobertson: 100% agree

src/iface/ethernet.rs
dlrobertson: Updated.

src/iface/ethernet.rs
dlrobertson: Updated.

whitequark:
Thanks!

Issue #64, opened 2017-10-28
Implement ICMP sockets


dlrobertson:
Implement ICMP sockets that are somewhat equivalent to linux `IPPROTO_ICMP` sockets. Much of the implementation may be similar to the other socket types, but there will likely be a few crucial differences. 

 - `Icmpv4Socket::accepts` is determined *only* by the source IP and destination IP. This means that if two separate icmp sockets sent from the same source to the same destination, things would get mixed up. Without access to the IP id, I don't see a way around this. If this direction is taken, a `should_fail` test  with a note and an issue to make sure we don't lose track of this would be nice.
 - The `send` and `send_slice` methods should be somewhat similar to the current udp implementation where the destination address is an arg.
 - It might be more readable to use a `recv_from` method that takes the address to receive packets from as an arg instead of the typical `recv` function. If we really want to stick to `recv` instead of `recv_from` we could keep the destination that will be used by `Icmpv4Socket::accepts` from the previous send.

Related to: #50

PR #69, opened 2017-11-04
[WIP] Add ICMPv4 sockets

dlrobertson:
- Add support for ICMP sockets
 - Add tests for ICMP sockets
 - Rename proto-*type* features to socket-*type*
 - Update documentation

Resolves: #64

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -24,6 +24,8 @@ use socket::RawSocket;
 use socket::UdpSocket;
 #[cfg(feature = "socket-tcp")]
 use socket::TcpSocket;
+#[cfg(feature = "socket-icmp")]
+use socket::IcmpSocket;
 use super::ArpCache;
 
 /// An Ethernet network interface.
@@ -55,7 +57,7 @@ struct InterfaceInner<'b, 'c> {
 enum Packet<'a> {
     None,
     Arp(ArpRepr),
-    Icmpv4(Ipv4Repr, Icmpv4Repr<'a>),
+    Icmpv4((Ipv4Repr, Icmpv4Repr<'a>)),
     #[cfg(feature = "socket-raw")]
     Raw((IpRepr, &'a [u8])),
     #[cfg(feature = "socket-udp")]
@@ -237,6 +239,16 @@ impl<'a, 'b, 'c, DeviceT> Interface<'a, 'b, 'c, DeviceT>
                             device_result = inner.dispatch(tx_token, timestamp, Packet::Tcp(response));
                             device_result
                         }),
+                    #[cfg(feature = "socket-icmp")]
+                    Socket::Icmp(ref mut socket) =>
+                        socket.dispatch(&caps, |response| {
+                            let tx_token = device.transmit().ok_or(Error::Exhausted)?;
+                            match response {
+                                (IpRepr::Ipv4(repr), icmp_repr) =>
+                                    inner.dispatch(tx_token, timestamp, Packet::Icmpv4((repr, icmp_repr))),
+                                _ => Err(Error::Unaddressable),
+                            }
+                        }),
                     Socket::__Nonexhaustive(_) => unreachable!()
                 };
             match (device_result, socket_result) {
@@ -386,7 +398,7 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
 
         match ipv4_repr.protocol {
             IpProtocol::Icmp =>
-                self.process_icmpv4(ipv4_repr, ip_payload),
+                self.process_icmpv4(sockets, ip_repr, ip_payload),
 
             #[cfg(feature = "socket-udp")]
             IpProtocol::Udp =>
@@ -414,12 +426,24 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         }
     }
 
-    fn process_icmpv4<'frame>(&self, ipv4_repr: Ipv4Repr, ip_payload: &'frame [u8]) ->
-                             Result<Packet<'frame>>
+    fn process_icmpv4<'frame>(&self, _sockets: &mut SocketSet, ip_repr: IpRepr,
+                              ip_payload: &'frame [u8]) -> Result<Packet<'frame>>
     {
         let icmp_packet = Icmpv4Packet::new_checked(ip_payload)?;
         let checksum_caps = self.device_capabilities.checksum.clone();
         let icmp_repr = Icmpv4Repr::parse(&icmp_packet, &checksum_caps)?;
+
+        #[cfg(feature = "socket-icmp")]
+        for mut icmp_socket in _sockets.iter_mut().filter_map(IcmpSocket::downcast) {
+            if !icmp_socket.accepts(&ip_repr, &icmp_repr, &checksum_caps) { continue }
+
+            match icmp_socket.process(&ip_repr, ip_payload) {
+                // The packet is valid and handled by socket.
+                Ok(()) => return Ok(Packet::None),
+                // The packet is malformed, or the socket buffer is full.
+                Err(e) => return Err(e)
+            }
+        }
 
         match icmp_repr {
             // Respond to echo requests.
@@ -429,7 +453,10 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
                     seq_no: seq_no,
                     data:   data
                 };
-                Ok(self.icmpv4_reply(ipv4_repr, icmp_reply_repr))
+                match ip_repr {
+                    IpRepr::Ipv4(ipv4_repr) => Ok(self.icmpv4_reply(ipv4_repr, icmp_reply_repr)),
+                    _ => Err(Error::Unrecognized),
+                }
             }
 
             // Ignore any echo replies.
@@ -452,7 +479,7 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
                 payload_len: icmp_repr.buffer_len(),
                 ttl:         64
             };
-            Packet::Icmpv4(ipv4_reply_repr, icmp_repr)
+            Packet::Icmpv4((ipv4_reply_repr, icmp_repr))
         } else {
             // Do not send any ICMP replies to a broadcast destination address.
             Packet::None
@@ -551,7 +578,7 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
                     arp_repr.emit(&mut packet);
                 })
             },
-            Packet::Icmpv4(ipv4_repr, icmpv4_repr) => {
+            Packet::Icmpv4((ipv4_repr, icmpv4_repr)) => {
                 self.dispatch_ip(tx_token, timestamp, IpRepr::Ipv4(ipv4_repr), |_ip_repr, payload| {
                     icmpv4_repr.emit(&mut Icmpv4Packet::new(payload), &checksum_caps);
                 })
@@ -704,17 +731,19 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
 #[cfg(test)]
 mod test {
     use std::boxed::Box;
-    use super::Packet;
+    use {Result, Error};
+
+    use iface::{ArpCache, SliceArpCache, EthernetInterface};
     use phy::{self, Loopback, ChecksumCapabilities};
+    use socket::SocketSet;
     use wire::{ArpOperation, ArpPacket, ArpRepr};
     use wire::{EthernetAddress, EthernetFrame, EthernetProtocol};
     use wire::{IpAddress, IpCidr, IpProtocol, IpRepr};
     use wire::{Ipv4Address, Ipv4Repr};
     use wire::{Icmpv4Repr, Icmpv4DstUnreachable};
     use wire::{UdpPacket, UdpRepr};
-    use iface::{ArpCache, SliceArpCache, EthernetInterface};
-    use socket::SocketSet;
-    use {Result, Error};
+
+    use super::Packet;
 
     fn create_loopback<'a, 'b>() ->
             (EthernetInterface<'a, 'static, 'b, Loopback>, SocketSet<'static, 'a, 'b>) {
@@ -814,7 +843,7 @@ mod test {
             data: &NO_BYTES
         };
 
-        let expected_repr = Packet::Icmpv4(
+        let expected_repr = Packet::Icmpv4((
             Ipv4Repr {
                 src_addr: Ipv4Address([0x7f, 0x00, 0x00, 0x01]),
                 dst_addr: Ipv4Address([0x7f, 0x00, 0x00, 0x02]),
@@ -823,7 +852,7 @@ mod test {
                 ttl: 64
             },
             icmp_repr
-        );
+        ));
 
         // Ensure that the unknown protocol triggers an error response.
         // And we correctly handle no payload.
@@ -879,7 +908,7 @@ mod test {
             },
             data: &data
         };
-        let expected_repr = Packet::Icmpv4(
+        let expected_repr = Packet::Icmpv4((
             Ipv4Repr {
                 src_addr: Ipv4Address([0x7f, 0x00, 0x00, 0x01]),
                 dst_addr: Ipv4Address([0x7f, 0x00, 0x00, 0x02]),
@@ -888,7 +917,7 @@ mod test {
                 ttl: 64
             },
             icmp_repr
-        );
+        ));
 
         // Ensure that the unknown protocol triggers an error response.
         // And we correctly handle no payload.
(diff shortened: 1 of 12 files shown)


src/iface/ethernet.rs
```diff
@@ -49,7 +50,7 @@ struct InterfaceInner<'b, 'c> {
 enum Packet<'a> {
     None,
     Arp(ArpRepr),
-    Icmpv4(Ipv4Repr, Icmpv4Repr<'a>),
+    Icmpv4((Ipv4Repr, Icmpv4Repr<'a>)),
```
dlrobertson: Updated this to be more consistent with the other packet enums

src/wire/icmpv4.rs
```diff
@@ -265,13 +273,71 @@ impl<T: AsRef<[u8]>> Packet<T> {
...
     /// The result depends on the value of the message type field.
     pub fn header_len(&self) -> usize {
         match self.msg_type() {
-            Message::EchoRequest    => field::ECHO_SEQNO.end,
-            Message::EchoReply      => field::ECHO_SEQNO.end,
-            Message::DstUnreachable => field::UNUSED.end,
+            Message::EchoRequest      => field::ECHO_SEQNO.end,
+            Message::EchoReply        => field::ECHO_SEQNO.end,
+            Message::DstUnreachable   => field::UNUSED.end,
+            Message::TimestampRequest => field::TM_TX_TM.end,
```
dlrobertson: Added `TimestampRequest` and `Timestamp` reply. They're not essential to this PR. The main reason for adding them was for the test case in `iface::ethernet::test::test_icmpv4_socket`

src/wire/icmpv4.rs
```diff
@@ -339,6 +405,59 @@ impl<T: AsRef<[u8]> + AsMut<[u8]>> Packet<T> {
         NetworkEndian::write_u16(&mut data[field::ECHO_SEQNO], value)
     }
 
+    /// Set the identifier field (for timestamp request and reply packets).
+    ///
+    /// # Panics
+    /// This function may panic if this packet is not an timestamp request or reply packet.
+    #[inline]
+    pub fn set_timestamp_ident(&mut self, value: u16) {
```
dlrobertson: the `timestamp_ident` and `timestamp_seq_no` are exact copies of `echo_ident` and `echo_seq_no`, but I added them for consistency/readability.

src/iface/ethernet.rs
```diff
@@ -990,4 +1011,56 @@ mod test {
...
+
+        let icmpv4_socket = Icmpv4Socket::new(rx_buffer, tx_buffer);
+
+        let socket_handle = socket_set.add(icmpv4_socket);
+
+        {
+            let mut icmp_socket = socket_set.get::<Icmpv4Socket>(socket_handle);
+
+            assert_eq!(icmp_socket.recv(Ipv4Address([0x7f, 0x00, 0x00, 0x02])),
+                       Err(Error::Exhausted));
```
dlrobertson: This is what I really don't like about this implementation. I originally wanted to stay away from using a `bind` call, but now as a result if you need to use a `recv` as the `bind` in a way.

src/iface/ethernet.rs
```diff
@@ -18,6 +18,7 @@ use socket::{Socket, SocketSet, AnySocket};
 #[cfg(feature = "proto-raw")] use socket::RawSocket;
 #[cfg(feature = "proto-udp")] use socket::UdpSocket;
 #[cfg(feature = "proto-tcp")] use socket::TcpSocket;
+#[cfg(feature = "proto-icmpv4")] use socket::Icmpv4Socket;
```
whitequark: Ugh. Just after I migrated `socket-*` features to `proto-*` I think we'll need to go back.

src/iface/ethernet.rs
```diff
@@ -231,6 +232,13 @@ impl<'a, 'b, 'c, DeviceT> Interface<'a, 'b, 'c, DeviceT>
                             device_result = inner.dispatch(tx_token, timestamp, Packet::Tcp(response));
                             device_result
                         }),
+                    #[cfg(feature = "proto-icmpv4")]
+                    Socket::Icmpv4(ref mut socket) =>
+                        socket.dispatch(&caps, |response| {
+                            let tx_token = device.transmit().ok_or(Error::Exhausted)?;
+                            device_result = inner.dispatch(tx_token, timestamp, Packet::Icmpv4(response));
+                            device_result
+                        }),
```
whitequark: This match statement *begs* being put in a macro.

src/iface/ethernet.rs
whitequark: I think this will need to be reworked to have a proper `bind` call, at least if we want to support downstreams like redox which present a POSIX-like interface. Take a look at this: https://www.planet-lab.org/raw_sockets/api_icmp.html.

src/iface/ethernet.rs
```diff
@@ -990,4 +1011,56 @@ mod test {
...
+        let socket_handle = socket_set.add(icmpv4_socket);
+
+        {
+            let mut icmp_socket = socket_set.get::<Icmpv4Socket>(socket_handle);
+
+            assert_eq!(icmp_socket.recv(Ipv4Address([0x7f, 0x00, 0x00, 0x02])),
+                       Err(Error::Exhausted));
+        }
+
+        let icmp_repr = Icmpv4Repr::TimestampRequest {
```
whitequark: That request is deprecated, no? I try to avoid adding deprecated things to `wire` unless they're used in practice.

src/socket/icmpv4.rs
```diff
@@ -0,0 +1,378 @@
+use core::cmp;
+use managed::Managed;
+
+use {Error, Result};
+use phy::DeviceCapabilities;
+use socket::{Socket, SocketHandle};
+use storage::{Resettable, RingBuffer};
+use wire::{IpProtocol, Ipv4Address, Ipv4Repr, Icmpv4Repr, Icmpv4Packet};
+
+/// A buffered ICMPv4 packet.
```
whitequark: I think that ICMP sockets could easily serve both IPv4 and IPv6. After all, the first nibble of the header already tells you which packet it is.

src/socket/icmpv4.rs
```diff
@@ -0,0 +1,378 @@
...
+/// owns transmit and receive packet buffers.
+#[derive(Debug)]
+pub struct Icmpv4Socket<'a, 'b: 'a> {
+    handle:    SocketHandle,
+    rx_buffer: SocketBuffer<'a, 'b>,
+    tx_buffer: SocketBuffer<'a, 'b>,
+    /// The endpoint this socket is communicating with
+    dst_addr:  Ipv4Address,
+    /// The time-to-live (IPv4) or hop limit (IPv6) value used in outgoing packets.
+    ttl:       Option<u8>
```
whitequark: Since this socket works with raw packets the TTL is already in there.

src/iface/ethernet.rs
dlrobertson: +1

src/iface/ethernet.rs
dlrobertson: Np. Should I add that in this PR?

src/iface/ethernet.rs
dlrobertson: Yeah, it's not really used. It's really only used when echo requests are blocked by a firewall. I'll just manually write out the bytes for the test in the interface tests.

src/socket/icmpv4.rs
dlrobertson: True

src/socket/icmpv4.rs
dlrobertson: We still use `IpRepr::emit`, so we still need to populate the ttl of the repr [here](https://github.com/dlrobertson/smoltcp/blob/icmpv4_socket/src/socket/icmpv4.rs#L209-L224) right? It is a "raw packet" but I made the assumption that the raw data passed to `send` would not include the IP header. That may be an error in my implementation.

src/iface/ethernet.rs
dlrobertson: That is a super helpful doc

src/socket/icmpv4.rs
dlrobertson: Id be interested to see how the type based `binds` are implemented for IPv6. AFAIK we'd need to support the Fragment header to get an identification number, but the id is 32 bits in IPv6. I'll have to dig through linux and/or some docs.

src/iface/ethernet.rs
whitequark: Sure.

src/iface/ethernet.rs
whitequark: Why not use an echo request?

src/socket/icmpv4.rs
whitequark: Don't bother with IPv6 for now, its headers are fairly broken and we'll need to aggressively compress them into Reprs.

src/socket/icmpv4.rs
whitequark: Ah, true.

src/iface/ethernet.rs
dlrobertson: Updated

src/iface/ethernet.rs
dlrobertson: Updated

src/iface/ethernet.rs
dlrobertson: Only comment I haven't fixed yet. Thinking about how to best accomplish this since some branches need `ChecksumCaps` etc.

dlrobertson:
See [src/socket/icmp.rs:137-174](https://github.com/dlrobertson/smoltcp/blob/icmpv4_socket/src/socket/icmp.rs#L137-L174) for the updated API using `bind`.

whitequark:
@dlrobertson The updated API doesn't look quite right. Shouldn't you be binding to the identifier field in the IP header rather than IP endpoint? Imagine you are implementing a traceroute program. With the current API this is completely impossible.

dlrobertson:
> To send and receive ICMP messages that are not associated with a specific TCP/UDP port number (e.g., Echo, Echo Reply, Timestamp, Timestamp Reply, Information Request, Information Reply), the socket has to be bound to a specific ICMP identifier. The ICMP identifier is a 16-bit field present in bytes 5/6 in the header of these messages. Only messages containing the right identifier can be sent or received through a safe raw ICMP socket of this type.

So it isn't the identifier in the IP header (I was wrong in my previous comment) it is the identifier in the Icmp message. I used the `IpEndpoint` port value for consistency with other implementations and for the case where you're binding to ICMP error responses to a UDP port. Then the socket creation takes an additional parameter that determines how the port value of the bound `IpEndpoint` works. As a result, the current API includes the following two cases.

Bind to ICMP error responses for UDP packets sent from port 53.

```rust
use smoltcp::socket::{Socket, IcmpSocket, IcmpSocketType};
// Created with type Udp. IpEndpoint::port is truely a port.
let mut icmp_socket = match IcmpSocket::new(rx_buffer, tx_buffer, IcmpSocketType::Udp) {
     Socket::Icmp(socket) => socket,
     _ => unreachable!()
};
icmp_socket.bind(53).unwrap();
```

Bind to ICMP messages with the identifier 0x1234

```rust
use smoltcp::socket::{Socket, IcmpSocket, IcmpSocketType};
// Created with type Icmp. IpEndpoint::prt is actually the 16 bit identifier
let mut icmp_socket = match IcmpSocket::new(rx_buffer, tx_buffer, IcmpSocketType::Icmp) {
    Socket::Icmp(socket) => socket,
    _ => unreachable!()
};
icmp_socket.bind(0x1234).unwrap();
```

I chose to implement this using a parameter added to socket creation, but this could also be accomplished using an enum passed to bind. I'm starting to realize the use `IpEndpoint` makes sense for C, but I think it would be more readable if I use some sort of enum.

whitequark:
Ah, I see. Yes, I think a dedicated enum `IcmpEndpoint` would be best.

dlrobertson:
How important is it to keep the standard `bind` API? E.g. How bad would it be if we ended up with something like.

```rust
let mut socket = ...
socket.bind(IcmpEndpoint::Identifier(0x1234))
```

When `bind` is only given a `u16` it gets to be a bit ambiguous. In theory `bind` could keep the same signature. Then we could implement the setting of the sockets endpoint with something like the following.

```rust
match self.socket_type {
    SocketType::Udp => self.endpoint = IcmpEndpoint::Udp(endpoint.into()),
    SocketType::Icmp => {
        let tmp_endpoint: IpEndpoint = endpoint.into();
        self.endpoint = IcmpEndpoint::Icmp(tmp_endpoint.port),
    }
}
```

whitequark:
There's no "standard bind API", we do not try to implement POSIX. There aren't ICMP sockets in POSIX anyway. So, do the most clear thing.

dlrobertson:
Rebased on master and updated the `ping` example. I got it to work on my system, but otherwise I did very little testing of it.

src/iface/ethernet.rs
```diff
@@ -996,4 +1025,59 @@ mod test {
...
+
+        let ip_repr = IpRepr::Ipv4(Ipv4Repr {
+            src_addr:    Ipv4Address::new(0x7f, 0x00, 0x00, 0x02),
+            dst_addr:    Ipv4Address::new(0x7f, 0x00, 0x00, 0x01),
+            protocol:    IpProtocol::Icmp,
+            payload_len: 24,
+            ttl:         64
+        });
+
+        // Timestamp Requests are currently ignored by the interface. Open a
```
whitequark: Outdated comment?

src/socket/icmp.rs
```diff
@@ -0,0 +1,569 @@
...
+use wire::{Icmpv4Packet, Icmpv4Repr};
+use wire::{UdpPacket, UdpRepr};
+
+/// Type of endpoint to bind the ICMP socket to. See [IcmpSocket::bind] for
+/// more details.
+///
+/// [IcmpSocket::bind]: struct.IcmpSocket.html#method.bind
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+pub enum Endpoint {
+    Identifier(u16),
```
whitequark: The header field is called "Identification" by the spec, I suggest just `Ident`

src/socket/icmp.rs
```diff
@@ -0,0 +1,569 @@
...
+    IpEndpoint(IpEndpoint),
```
whitequark: `Endpoint::IpEndpoint`? I suggest just `Ip`

src/socket/icmp.rs
```diff
@@ -0,0 +1,569 @@
...
+    Unspecified
```
whitequark: In general we place `Unspecified` variants first

src/socket/icmp.rs
```diff
@@ -0,0 +1,569 @@
...
+        }
+
+        self.ttl = ttl
+    }
+
+    /// Bind the socket to the given endpoint.
+    ///
+    /// This function returns `Err(Error::Illegal)` if the socket was open
+    /// (see [is_open](#method.is_open)), and `Err(Error::Unaddressable)`
+    /// if `endpoint` is `IpEndpoint` and the port is zero or if `endpoint`
```
whitequark: I'd say something like:

    if `endpoint` is unspecified (see [is_unspecified])

src/socket/icmp.rs
```diff
@@ -0,0 +1,569 @@
...
+    /// is `Unspecified`
+    ///
+    /// # Examples
+    ///
+    /// Bind the socket to ICMP Error messages associated with
+    /// a specific UDP port
```
whitequark: Missing colon

whitequark: In general, I think these examples don't have a burning need to be executable, whereas the accompanying text should elaborate more on the "why". Why is it useful to bind to an UDP endpoint with an ICMP socket?

src/socket/mod.rs
```diff
@@ -11,14 +11,15 @@
 //! size for a buffer, allocate it, and let the networking stack use it.
 
 use core::marker::PhantomData;
-use wire::IpRepr;
 
 #[cfg(feature = "socket-raw")]
 mod raw;
```
whitequark: Can you order `icmp` after `raw`? Here and in `set.rs` and in `ref_.rs`

whitequark:
After the comments above are fixed this is ready to be merged.

src/socket/icmp.rs
dlrobertson: What about `Udp`? Since you're binding to ICMP `DstUnreachables` for UDP packets from the given port.

dlrobertson: I'm okay with `Ip` though.. just trying to think about how to make this as readable as possible

dlrobertson:
Thanks for the reviews. The final product was much much better than the first few revisions

PR #72, opened 2017-11-07
[WIP] IPv6 Address and CIDR to wire

dlrobertson:
Add some of the basics of `ipv6`.

- Add the `ipv6` feature
   - Ensure there is a travis build with the `ipv6` feature enabled.
 - Add the basic infrastructure to wire for IPv6 support.
   - `Ipv6Address`
   - `Ipv6Cidr`


Cargo.toml
```diff
@@ -12,7 +12,7 @@ categories = ["embedded"]
 license = "0BSD"
 
 [dependencies]
-byteorder = { version = "1.0", default-features = false }
+byteorder = { version = "1.0", default-features = false, features = ["i128"] }
```
dlrobertson: I couldn't find any docs on a way to enable a dependency feature based on a feature enabled in the parent project when the dependency isn't optional. If there is a way to conditionally enable this feature based on the `ipv6` feature. We should definitely do that.

src/wire/ipv6/extops.rs
```diff
@@ -0,0 +1,421 @@
...
+    pub fn set_data_len(&mut self, data_len: u8) {
+        assert!(self.opt_type() != ExtOpType::Pad1);
+        let data = self.buffer.as_mut();
+        data[field::DATA_LEN] = data_len;
+    }
+}
+
+impl<'a, T: AsRef<[u8]> + AsMut<[u8]> + ?Sized> Packet<&'a mut T> {
+    /// Return a mutable pointer to the payload.
+    pub fn payload_mut(&mut self) -> &mut [u8] {
```
dlrobertson: Using this on a `Pad1` option is more or less a user error since it cannot have a payload. I went back and forth on using an `assert(self.opt_type() != ExtOpType::Pad1)`.

whitequark:
Okay, this is *way* too much code to even discuss, much less merge, in a single PR. Can you please split it into many PRs, and start small? I suggest the following:
  1. Make IPv4 optional by adding a feature gating it.
  2. Add Ipv6Address/Ipv6Cidr.
  3. Add Ipv6Packet/Ipv6Repr.
Please do not yet make any provisions for handling extension headers, multicast, etc, I would like to first see a simple, clean version of the IPv6 basics go in. Then we can think about those superfluous things.

dlrobertson:
> Can you please split it into many PRs, and start small?

Np

> Please do not yet make any provisions for handling extension headers, multicast, etc

We can go without extension headers. We will not be able to get to the next header for any packet that contains an option, but even if we find out we need to implement options, we might be able to get away with implementing the absolute basics and not actually parsing the options contained within the given header. We can delay implementing multicast, but we'll need multicast to implement address resolution.

whitequark:
> We can go without extension headers.

To elaborate: I'm not saying that smoltcp *doesn't need* extension headers or multicast. I'm saying that I want to have a good view of the core functionality before I start building on it, so that I don't go overboard with complexity.

dlrobertson:
Gotcha. Makes way more sense now :smile: I'll try to get a super trimmed down version tomorrow.

dlrobertson:
Implemented 2 first so that when we add the ipv4 feature we don't hit a ton of unused and unreachable errors

whitequark:
I would strongly prefer that you do not use the i128 type. Not only it requires a feature, but also it is a code generation hazard on the kinds of targets smoltcp aims to be suitable for. E.g. the OR1K backend we use at M-Labs crashes trying to codegen it.

whitequark:
Well, and it breaks the build.

Cargo.toml
whitequark: For your reference, you can do `"ipv6" = ["byteorder/i128"]`.

dlrobertson:
> I would strongly prefer that you do not use the i128 type. Not only it requires a feature, but also it is a code generation hazard on the kinds of targets smoltcp aims to be suitable for. E.g. the OR1K backend we use at M-Labs crashes trying to codegen it.

Np, I'll figure out something

src/wire/ipv6/core.rs
```diff
@@ -0,0 +1,269 @@
...
+    fn test_is_loopback() {
+        assert!(!Address::LOOPBACK.is_unspecified());
+        assert!(!Address::LOOPBACK.is_multicast());
+        assert!(!Address::LOOPBACK.is_link_local());
+        assert!(Address::LOOPBACK.is_loopback());
+    }
+
+    #[test]
+    fn address_format() {
+        assert_eq!("ff02:0000:0000:0000:0000:0000:0000:0001",
```
whitequark: This PR would also be well served if you added an IPv6 parser to it.

src/wire/ip.rs
```diff
@@ -181,6 +239,12 @@ impl Cidr {
         match (self, addr) {
             (&Cidr::Ipv4(ref cidr), &Address::Ipv4(ref addr)) =>
                 cidr.contains_addr(addr),
+            #[cfg(feature = "ipv6")]
+            (&Cidr::Ipv6(ref cidr), &Address::Ipv6(ref addr)) =>
+                cidr.contains_addr(addr),
+            #[cfg(feature = "ipv6")]
+            (&Cidr::Ipv4(_), &Address::Ipv6(_)) | (&Cidr::Ipv6(_), &Address::Ipv4(_)) =>
```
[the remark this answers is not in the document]
dlrobertson: Good question. Would the conversions typically happen in `Repr::lower`

src/parsers.rs
```diff
@@ -115,6 +119,96 @@ impl<'a> Parser<'a> {
         Err(())
     }
 
+    #[cfg(feature = "ipv6")]
+    fn accept_ipv6(&mut self, is_cidr: bool) -> Result<Ipv6Address> {
```
dlrobertson: I tried to simplify this as much as possible, but since it is of a variable length, parsing is a bit complex. Suggestions on methods to make this more readable are very much welcomed.

src/parsers.rs
```diff
@@ -130,6 +224,13 @@ impl<'a> Parser<'a> {
         if let Some(ipv4) = self.try(|p| p.accept_ipv4()) {
             return Ok(IpAddress::Ipv4(ipv4))
         }
+
+        #[cfg(feature = "ipv6")]
+        match self.try(|p| p.accept_ipv6(false)) {
```
dlrobertson: Used a `match` because you can't add a `cfg` to an `if`.

src/parsers.rs
```diff
@@ -175,19 +286,59 @@ impl FromStr for Ipv4Cidr {
...
 
     /// Parse a string representation of an IP CIDR.
     fn from_str(s: &str) -> Result<IpCidr> {
-        Ipv4Cidr::from_str(s).map(IpCidr::Ipv4)
+        if let Ok(ipv4) = Ipv4Cidr::from_str(s) {
+            return Ok(IpCidr::Ipv4(ipv4))
+        }
+
+        #[cfg(feature = "ipv6")]
+        match Ipv6Cidr::from_str(s) {
```
dlrobertson: Used a `match` so that a `cfg` could be used here.

src/parsers.rs
whitequark: The easiest way to make this parser much simpler would be to make it recursive. Here's an outline:
```
fn accept_ipv6_part(&mut self, 
      (head, tail): (&mut [u16; 8], &mut [u16; 8]), 
      (head_idx, tail_idx): (&mut usize, &mut usize),
      mut use_tail: bool) -> Result<()> {
  match self.accept_str(b"::") {
    Ok(()) => use_tail = true,
    Err(()) => ()
  }

  if head_idx != 0 || use_tail && tail_idx != 0 {
    self.accept_char(b':')?;
  }

  match self.accept_number(4, 0x10000, true)? {
    Ok(part) if use_tail => {
      tail[tail_idx] = part;
      tail_idx += 1;
    }
    Ok(part) => {
      head[head_idx] = part;
      head_idx += 1;        
    }
    Err(()) => ()
  }

  if head_idx + tail_idx > 8 {
    Err(())
  } else if self.pos == self.data.len() {
    head[8 - tail.len()..].copy_from_slice(tail);
    Ok(())
  } else {
    self.accept_ipv6_part((head, tail), (head_idx, tail_idx), use_tail)
  }
}
```

src/wire/ipv6.rs
```diff
@@ -0,0 +1,382 @@
+use core::fmt;
+
+use byteorder::{ByteOrder, NetworkEndian};
+
+pub use super::IpProtocol as Protocol;
+
+// Helper used to convert 8 given u16 arguments to a [u8; 16]
+// slice
+macro_rules! write_to_addr {
```
whitequark: Do you really need a macro for something that will be used exactly once? Just expand it with copy/paste, it's more clear how it works then.

dlrobertson: Nah, just being lazy. Updated.

src/parsers.rs
dlrobertson: Can we rely on `(self.pos == self.data.len()) == true` event for a CIDR? I thought for the current implementation I just had to hand processing over as soon as I hit a non-number in the second parsing round or when I filled the addr buffer. Either way, adding that to the above would be pretty easy and this is def more readable. I'll work on this tomorrow after I get some sleep. I need more brain power to think through this :smile:

whitequark: Oh I guess you can look ahead for whether there's a `:` and bail out if there is not, instead of looking for EOF.

dlrobertson: Updated to use the recursive method and used `lookahead_char` instead of `is_cidr`. IMO it makes it less confusing and there are less mutable variables you need to keep track of.

batonius:
Should `IpEndpoint` have `FromStr` implementation as well? While it's trivial to split an IPv4 endpoint at ':', IPv6 endpoints are quite tricky: https://tools.ietf.org/html/rfc5952#section-6 . I think the recommended style would be enough, it's the only one `std::net::SocketAddrV6` supports: https://doc.rust-lang.org/nightly/src/std/net/parser.rs.html#280 .

whitequark:
@batonius Yes.

src/parsers.rs
```diff
@@ -115,6 +136,110 @@ impl<'a> Parser<'a> {
...
+                // Valid u16 to be added to the address
+                tail[*tail_idx] = part as u16;
+                *tail_idx += 1;
+                Ok(())
+            },
+            Some(_) => {
+                // Tail or head section is too long
+                Err(())
+            }
+            None if self.lookahead_char(b'/') => {
```
whitequark: No, this is not right. You shouldn't look for any specific character. Rather, you should bail out if you detect anything you don't expect.

src/parsers.rs
```diff
@@ -115,6 +136,110 @@ impl<'a> Parser<'a> {
...
+        // combine the second optional section to the end of the
+        // final address.
+        let (mut addr, mut tail) = ([0u16; 8], [0u16; 6]);
+        let (mut head_idx, mut tail_idx) = (0, 0);
+
+        self.accept_ipv6_part((&mut addr, &mut tail), (&mut head_idx, &mut tail_idx), false)?;
+
+        // If the tail_idx is not 0, we need to copy the tail portion
+        // (the portion following the "::") to the end of the address.
+        if tail_idx != 0 {
```
whitequark: This check isn't doing anything; it would prevent a zero-length copy.

src/parsers.rs
```diff
@@ -115,6 +136,110 @@ impl<'a> Parser<'a> {
...
+        // May be written as
+        //
+        // fe80::1
+        //
+        // As a result, we need to find the first section of colon
+        // delimited u16's before a possible "::", then the
+        // possible second section after the "::", and finally
+        // combine the second optional section to the end of the
+        // final address.
+        let (mut addr, mut tail) = ([0u16; 8], [0u16; 6]);
```
whitequark: Why does tail have the type `[u16; 6]`?

src/parsers.rs
dlrobertson: This particular branch is really only applicable to CIDRs, so maybe I should reintroduce `is_cidr` then

src/parsers.rs
dlrobertson: It can only be 6 words in length. The double colon consumes 32 bits at a minimum.

src/parsers.rs
dlrobertson: Correct

dlrobertson:
Cleaned up some logic

Cargo.toml
```diff
@@ -33,6 +33,7 @@ alloc = ["managed/alloc"]
 verbose = []
 "phy-raw_socket" = ["std", "libc"]
 "phy-tap_interface" = ["std", "libc"]
+"ipv6" = []
```
whitequark: `"proto-ipv6"`

src/parsers.rs
```diff
@@ -17,6 +21,15 @@ impl<'a> Parser<'a> {
         }
     }
 
+    #[cfg(feature = "ipv6")]
```
whitequark: Let's mark this entire file `#[cfg_attr(not(and(feature = "proto-ipv4", feature = "proto-ipv6")), allow(dead_code)]`.

src/parsers.rs
```diff
@@ -115,6 +136,103 @@ impl<'a> Parser<'a> {
         Err(())
     }
 
+    #[cfg(feature = "ipv6")]
+    fn accept_ipv6_part(&mut self, (head, tail): (&mut [u16; 8], &mut [u16; 6]),
```
whitequark: :+1: An excellent job refactoring this method, it is very easy to see now that it is correct.

src/wire/ip.rs
```diff
@@ -75,38 +98,53 @@ impl Address {
...
 
     /// Return an unspecified address that has the same IP version as `self`.
     pub fn to_unspecified(&self) -> Address {
         match self {
-            &Address::Unspecified => Address::Unspecified,
-            &Address::Ipv4(_) => Address::Ipv4(Ipv4Address::UNSPECIFIED),
+            &Address::Unspecified     => Address::Unspecified,
+            &Address::Ipv4(_)         => Address::Ipv4(Ipv4Address::UNSPECIFIED),
+            #[cfg(feature = "ipv6")]
+            &Address::Ipv6(_)         => Address::Ipv4(Ipv4Address::UNSPECIFIED),
```
whitequark: Copy-paste mistake.

src/wire/ipv6.rs
```diff
@@ -0,0 +1,372 @@
...
+/// A sixteen-octet IPv6 address.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
+pub struct Address(pub [u8; 16]);
+
+impl Address {
+    /// An unspecified address.
+    pub const UNSPECIFIED: Address = Address([0x00; 16]);
+
+    /// Link local all routers multicast address
+    pub const LINK_LOCAL_ALL_NODES: Address = Address([0xff, 0x02, 0x00, 0x0,
```
whitequark: These look ugly, perhaps add a line break after `=`, indent, and then lay out in two lines?

src/wire/ipv6.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub fn from_bytes(data: &[u8]) -> Address {
+        let mut bytes = [0; 16];
+        bytes.copy_from_slice(data);
+        Address(bytes)
+    }
+
+    /// Construct an IPv6 address from a sequence of words, in big-endian.
+    ///
+    /// # Panics
+    /// The function panics if `data` is not 8 words long
```
whitequark: Should panic but doesn't.

src/wire/ipv6.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    /// Query whether the IPv6 address is in the "link-local" range
+    pub fn is_link_local(&self) -> bool {
+        self.0[0..8] == [0xfe, 0x80, 0x00, 0x00,
+                         0x00, 0x00, 0x00, 0x00]
+    }
+
+    /// Query whether the IPv6 address is the "loopback" address
+    pub fn is_loopback(&self) -> bool {
+        self.0 == [0x00, 0x00, 0x00, 0x00,
```
whitequark: `== Self::LOOPBACK`?

src/wire/ipv6.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub fn is_unicast(&self) -> bool {
+        !(self.is_multicast() || self.is_unspecified())
+    }
+
+    /// Query whether the IPv6 address is a multicast address.
+    pub fn is_multicast(&self) -> bool {
+        self.0[0] == 0xff
+    }
+
+    /// Query whether the IPv6 address is the "unspecified" address
```
whitequark: Missing dots in doc-comments here and in many other places.

src/wire/ipv6.rs
```diff
@@ -0,0 +1,372 @@
...
+impl fmt::Display for Address {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let bytes = self.0;
+        write!(f,
+               "{:02x}{:02x}:{:02x}{:02x}:{:02x}{:02x}:{:02x}{:02x}:\
+               {:02x}{:02x}:{:02x}{:02x}:{:02x}{:02x}:{:02x}{:02x}",
+               bytes[0], bytes[1], bytes[2], bytes[3],
+               bytes[4], bytes[5], bytes[6], bytes[7],
+               bytes[8], bytes[9], bytes[10], bytes[11],
+               bytes[12], bytes[13], bytes[14], bytes[15])
```
whitequark: This could really use a very simple `::` shortener (start munging at the first `0u16` part).

src/wire/ipv6.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+/// A specification of an IPv6 CIDR block, containing an address and a variable-length
+/// subnet masking prefix length.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default)]
```
whitequark: `Cidr` is not `PartialOrd` in any meaningful way I can see.

src/wire/ipv6.rs
dlrobertson: I took a very simple approach and added the IPv4 mapped IPv6 address case. Ipv4 Compatible addresses have been deprecated, so there is no need to add a case for them. I thought about essentially copy and pasting https://github.com/rust-lang/rust/blob/1.19.0/src/libstd/net/ip.rs#L1113-L1182, but slice pattern matching is used

src/wire/ipv6.rs
dlrobertson: Good catch

src/wire/ip.rs
dlrobertson: Good catch

src/parsers.rs
dlrobertson: Added `#[cfg_attr(not(feature = "proto-ipv6"), allow(dead_code))]`. Will add the `proto-ipv4` case once I move on to the `proto-ipv4` feature component of #76

src/wire/ipv6.rs
```diff
@@ -0,0 +1,413 @@
...
+
+impl fmt::Display for Address {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.0[0..10] == [0u8; 10] && self.0[10..12] == [0xffu8; 2] {
+            // The address is a IPv4 mapped address
+            write!(f, "::ffff:{}.{}.{}.{}", self.0[12], self.0[13], self.0[14], self.0[15])
+        } else {
+            let mut words = [0u16; 8];
+            self.write_parts(&mut words);
+            match words.iter().position(|x| *x == 0) {
```
whitequark: Maybe...
```
enum State { Head, Body, Tail }
let mut state = Head;
for word in words {
  state = match (word, state) {
    (0, Head) | 
    (0, Body) => { write!(f,     "::"); Tail },
    (0, Tail) => { write!(f,   "0000"); Tail }
    (_, Head) => { write!(f,  "{:04x}", word); Body },
    (_, Body) | 
    (_, Tail) => { write!(f, ":{:04x}", word); state } 
  }
}
```
?

src/wire/ipv6.rs
```diff
@@ -0,0 +1,413 @@
...
+        }
+        bytes
+    }
+}
+
+impl fmt::Display for Address {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.0[0..10] == [0u8; 10] && self.0[10..12] == [0xffu8; 2] {
+            // The address is a IPv4 mapped address
+            write!(f, "::ffff:{}.{}.{}.{}", self.0[12], self.0[13], self.0[14], self.0[15])
```
whitequark: Seems weird to stringify those since we cannot parse them. It should be one or the other.

dlrobertson: After thinking about this again, adding IPv4 mapped addresses could be a good follow up PR. AFAIK `From<Ipv4Address>` should be implemented for `Ipv6Addr`, we should be able to parse IPv4 mapped addresses, and we should print them differently. I'll remove this for now

src/wire/ipv6.rs
dlrobertson: Nice!

dlrobertson:
Added some comments to the `Ipv6Address` `Display` implementation and a test case for the `Ipv6Address` parser.

PR #73, opened 2017-11-08
Store plain DeviceT instead of Managed<_> in EthernetInterface

phil-opp:
This avoids unnecessary boxing of the device and simplifies the code. However, it does no longer allow EthernetInterfaces with a borrowed device.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -42,7 +42,7 @@ fn main() {
     let ip_addrs = [IpCidr::new(IpAddress::v4(192, 168, 69, 2), 24)];
     let default_v4_gw = Ipv4Address::new(192, 168, 69, 100);
     let mut iface = EthernetInterface::new(
-        Box::new(device), Box::new(arp_cache) as Box<ArpCache>,
+        device, Box::new(arp_cache) as Box<ArpCache>,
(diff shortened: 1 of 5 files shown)


whitequark:
I *think* the idea was to make it possible to store a trait object, but Device is no longer object-safe anyway, right?

phil-opp:
In https://github.com/m-labs/rust-managed/issues/3#issuecomment-339386137, you assumed that it used to store a trait object. Now it is a generic parameter, so we can store `DeviceT` directly instead of `Managed<DeviceT>`.

phil-opp:
I'm not sure about the object safety of `Device`, but I will take a look.

phil-opp:
> I'm not sure about the object safety of Device, but I will take a look.

The problem seems to be that we need to specify the associated types in order to construct a Box<Device>:

```
error[E0191]: the value of the associated type `TxToken` (from the trait `phy::Device`) must be specified
  --> src/iface/ethernet.rs:35:31
   |
35 |     device: ::std::boxed::Box<Device<'a>>,
   |                               ^^^^^^^^^^ missing associated type `TxToken` value
```

So we would need generic parameters again…

Issue #63, opened 2017-10-25
DHCP support


phil-opp:
Is DHCP support planned? And is there a way to implement DHCP on top of smoltcp, i.e. without writing own UDP packet construction code?

whitequark:
> Is DHCP support planned?

Definitely.

> And is there a way to implement DHCP on top of smoltcp, i.e. without writing own UDP packet construction code?

I'm still thinking about the best way to implement protocols like DHCP. In a nutshell, I'd like them to be runnable on top of smoltcp, but also on top of regular `std::net::UdpSocket`s. I'm thinking a state machine without IO plus an adapter to IO might be right here, but I still need to experiment a bit.

whitequark:
OK here's the plan.
  1. DHCP gets a `wire::dhcp` module, just like all the other wire modules. For options you can take a look at how TCP options are implemented. This is very straightforward since TCP and DHCP are nearly identical in the general format of their packets.
  2. We add `smoltcp::io`, which has the trait `UdpIoAdapter` described below. Then we `impl UdpIoAdapter for UdpSocket` as well as `impl UdpIoAdapter for std::io::UdpSocket`. If @batonius wants to run DHCP in a separate daemon then he can also `impl UdpIoAdapter for RedoxUdpSocket` or however that works.

```
trait UdpIoAdapter {
  type Error;
  // still not sure about the signature of `send`
  fn send(&mut self, size: usize, endpoint: IpEndpoint) -> Result<&mut [u8], Self::Error>;
  fn recv(&mut self) -> Result<(&[u8], IpEndpoint), Self::Error>
}
```
  3. We add `smoltcp::proto`, which has the struct `Dhcp` in it. This struct contains a state machine that can be polled with a `&mut UdpIoAdapter` every time the adapter readiness changes or a timer is hit, and which exposes functions to get e.g. the current acquired address. Most importantly the struct does not hold any pointers to the `EthernetInterface` or any sockets, and so it does not prevent us from manipulating those.

I think you can start with implementing (1), and then I'll look at how (2) works best. The logic of (3) is mostly glue so it's best done last.

phil-opp:
Ok, I'm starting to work on (1).

dhslichter:
@whitequark @sbourdeauducq when do you anticipate DHCP support being implemented?

whitequark:
@dhslichter No ETA currently.

whitequark:
[...]

PR #75, opened 2017-11-08
DHCP wire module

phil-opp:
Implement a `wire/dhcp` module as proposed in https://github.com/m-labs/smoltcp/issues/63.

### DIFF
--- a/src/wire/dhcpv4.rs
+++ b/src/wire/dhcpv4.rs
@@ -540,16 +540,84 @@ impl<'a, T: AsRef<[u8]> + AsMut<[u8]> + ?Sized> Packet<&'a mut T> {
 }
 
 /// A high-level representation of a Dynamic Host Configuration Protocol packet.
+///
+/// DHCP messages have the following layout (see [RFC 2131](https://tools.ietf.org/html/rfc2131)
+/// for details):
+///
+/// ```no_rust
+/// 0                   1                   2                   3
+/// 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+/// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+/// | message_type  | htype (N/A)   |   hlen (N/A)  |   hops        |
+/// +---------------+---------------+---------------+---------------+
+/// |                       transaction_id                          |
+/// +-------------------------------+-------------------------------+
+/// |           secs                |           flags               |
+/// +-------------------------------+-------------------------------+
+/// |                           client_ip                           |
+/// +---------------------------------------------------------------+
+/// |                            your_ip                            |
+/// +---------------------------------------------------------------+
+/// |                           server_ip                           |
+/// +---------------------------------------------------------------+
+/// |                        relay_agent_ip                         |
+/// +---------------------------------------------------------------+
+/// |                                                               |
+/// |                    client_hardware_address                    |
+/// |                                                               |
+/// |                                                               |
+/// +---------------------------------------------------------------+
+/// |                                                               |
+/// |                          sname  (N/A)                         |
+/// +---------------------------------------------------------------+
+/// |                                                               |
+/// |                          file    (N/A)                        |
+/// +---------------------------------------------------------------+
+/// |                                                               |
+/// |                          options                              |
+/// +---------------------------------------------------------------+
+/// ```
+///
+/// It is assumed that the access layer is Ethernet, so `htype` (the field representing the
+/// hardware address type) is always set to `1`, and `hlen` (which represents the hardware address
+/// length) is set to `6`.
+///
+/// The `options` field has a variable length.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub struct Repr<'a> {
+    /// This field is also known as `op` in the RFC. It indicates the type of DHCP message this
+    /// packet represents.
     pub message_type: MessageType,
+    /// This field is also known as `xid` in the RFC. It is a random number chosen by the client,
+    /// used by the client and server to associate messages and responses between a client and a
+    /// server.
     pub transaction_id: u32,
+    /// This field is also known as `chaddr` in the RFC and for networks where the access layer is
+    /// ethernet, it is the client MAC address.
     pub client_hardware_address: EthernetAddress,
+    /// This field is also known as `ciaddr` in the RFC. It is only filled in if client is in
+    /// BOUND, RENEW or REBINDING state and can respond to ARP requests.
     pub client_ip: Ipv4Address,
+    /// This field is also known as `yiaddr` in the RFC.
     pub your_ip: Ipv4Address,
+    /// This field is also known as `siaddr` in the RFC. It may be set by the server in DHCPOFFER
+    /// and DHCPACK messages, and represent the address of the next server to use in bootstrap.
     pub server_ip: Ipv4Address,
+    /// This field is also known as `giaddr` in the RFC. In order to allow DHCP clients on subnets
+    /// not directly served by DHCP servers to communicate with DHCP servers, DHCP relay agents can
+    /// be installed on these subnets. The DHCP client broadcasts on the local link; the relay
+    /// agent receives the broadcast and transmits it to one or more DHCP servers using unicast.
+    /// The relay agent stores its own IP address in the `relay_agent_ip` field of the DHCP packet.
+    /// The DHCP server uses the `relay_agent_ip` to determine the subnet on which the relay agent
+    /// received the broadcast, and allocates an IP address on that subnet. When the DHCP server
+    /// replies to the client, it sends the reply to the `relay_agent_ip` address, again using
+    /// unicast. The relay agent then retransmits the response on the local network
     pub relay_agent_ip: Ipv4Address,
+    /// Broadcast flags. It can be set in DHCPDISCOVER, DHCPINFORM and DHCPREQUEST message if the
+    /// client requires the response to be broadcasted.
     pub broadcast: bool,
+    /// The "requested IP address" option. It can be used by clients in DHCPREQUEST or DHCPDISCOVER
+    /// messages, or by servers in DHCPDECLINE messages.
     pub requested_ip: Option<Ipv4Address>,
     /// The "client identifier" option.
     ///
@@ -561,6 +629,8 @@ pub struct Repr<'a> {
     /// it MUST use that same identifier in all subsequent messages, to ensure that all servers
     /// correctly identify the client.
     pub client_identifier: Option<EthernetAddress>,
+    /// The "server identifier" option. It is used both to identify a DHCP server
+    /// in a DHCP message and as a destination address from clients to servers.
     pub server_identifier: Option<Ipv4Address>,
     /// The parameter request list informs the server about which configuration parameters
     /// the client is interested in.


src/wire/dhcp.rs
```diff
@@ -0,0 +1,416 @@
...
+            MessageType::Discover | MessageType::Inform | MessageType::Request |
+                MessageType::Decline | MessageType::Release => 1,
+            MessageType::Offer | MessageType::Ack | MessageType::Nak => 2,
+        }
+    }
+}
+
+/// The possible hardware address types of a DHCP packet.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum HardwareAddressType {
```
whitequark: You should take a look at how other enums are defined. There's a macro that writes conversion functions for you:
```
enum_with_unknown! {
    /// ARP operation type.
    pub enum Operation(u16) {
        Request = 1,
        Reply = 2
    }
}
```

whitequark: Looks good otherwise, though.

phil-opp: That's useful, thanks!

phil-opp:
I implemented basic parsing and emitting of packets and DHCP options. I'll add a few more essential options, but I don't think that we need the rest right now.

phil-opp:
@whitequark I think the most important options are implemented now (message type, requested ip, server identifier). Are there any other options that we need? Other things that I should add/change?

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+        self.buffer
+    }
+
+    /// Returns wether this packet is a request (1) or a reply (2).
+    pub fn opcode(&self) -> u8 {
+        let data = self.buffer.as_ref();
+        data[field::OP]
+    }
+
+    pub fn hardware_address_type(&self) -> HardwareAddressType {
```
whitequark: The names of this and the following method as well as their values should be shared with ARP.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+        let field = &self.buffer.as_ref()[field::GIADDR];
+        Ipv4Address::from_bytes(field)
+    }
+
+    pub fn broadcast_flag(&self) -> Result<bool> {
+        let field = &self.buffer.as_ref()[field::FLAGS];
+        match NetworkEndian::read_u16(field) {
+            0 => Ok(false),
+            1 => Ok(true),
+            _ => Err(Error::Malformed),
```
whitequark: > The remaining bits of the flags field are reserved for future use.  They MUST be set to zero by clients and ignored by servers and relay agents.

Note `MUST be ignored`, not rejected.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+    /// Return a pointer to the options.
+    #[inline]
+    pub fn options_mut(&mut self) -> Result<&mut [u8]> {
+        let data = self.buffer.as_mut();
+        data.get_mut(field::OPTIONS).ok_or(Error::Truncated)
+    }
+}
+
+
+enum_with_unknown! {
```
whitequark: All enums generally go to the top of the file, for easier overview.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+        match *buffer.get(0).ok_or(Error::Truncated)? {
+            field::OPT_END => {
+                length = 1;
+                option = DhcpOption::EndOfList;
+            }
+            field::OPT_PAD => {
+                unimplemented!();
+                // “The pad option can be used to cause subsequent fields to align on word
+                // boundaries.”
+                // TODO: what does this mean? Align the next field or align all following fields?
```
whitequark: No, this just means that servers may insert pad options if they want, and clients should ignore them. It's for convenience of server implementers.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+                length = self.buffer_len();
+                buffer[1] = length as u8;
+                match self {
+                    &DhcpOption::EndOfList => unreachable!(),
+                    &DhcpOption::MessageType(value) => {
+                        buffer[0] = field::OPT_DHCP_MESSAGE_TYPE;
+                        buffer[2] = value.into();
+                    }
+                    &DhcpOption::RequestedIp(ip) | &DhcpOption::ServerIdentifier(ip) => {
+                        buffer[2..].copy_from_slice(ip.as_bytes());
```
whitequark: This will fail if `buffer[2..]` is longer than 4.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+                    }
+                    &DhcpOption::Other { kind, data: provided } => {
+                        buffer[0] = kind;
+                        buffer[2..].copy_from_slice(provided);
```
whitequark: Same as above.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub server_identifier: Option<Ipv4Address>,
+    pub parameter_request_list: Option<&'a [u8]>,
+}
+
+impl<'a> Repr<'a> {
+    /// Parse a DHCP packet and return a high-level representation.
+    pub fn parse<T>(packet: &Packet<&'a T>) -> Result<Self>
+            where T: AsRef<[u8]> + ?Sized {
+
+        packet.check_len()?;
```
whitequark: Useless check.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+                }
+                DhcpOption::Other {kind: field::OPT_PARAMETER_REQUEST_LIST, data} => {
+                    parameter_request_list = Some(data);
+                }
+                DhcpOption::Other {..} => {}
+            }
+            options = next_options;
+        }
+
+        // only ethernet is supported right now
```
whitequark: All validation should come before all further processing, if at all possible.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+            HardwareAddressType::Ethernet => {
+                if packet.hardware_address_len() != 6 {
+                    return Err(Error::Malformed);
+                }
+            }
+            _ => return Err(Error::Unrecognized), // unimplemented
+        }
+
+        // check magic number
+        if packet.magic_number() != 0x63825363 {
```
whitequark: Constant should be extracted. Also, the comment is trivial.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+
+        packet.check_len()?;
+        let transaction_id = packet.transaction_id();
+        let client_hardware_address = packet.client_hardware_address();
+        let client_ip = packet.client_ip();
+        let your_ip = packet.your_ip();
+        let server_ip = packet.server_ip();
+        let relay_agent_ip = packet.relay_agent_ip();
+
+        let mut message_type = None;
```
whitequark: `message_type = Err(Error::Malformed)`...

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+        let mut options = packet.options()?;
+        while options.len() > 0 {
+            let (next_options, option) = DhcpOption::parse(options)?;
+            match option {
+                DhcpOption::EndOfList => break,
+                DhcpOption::MessageType(value) => {
+                    if value.opcode()? != packet.opcode() {
+                        return Err(Error::Malformed);
+                    }
+                    message_type = Some(value);
```
whitequark: ... then set it to `Ok(value)` here.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,632 @@
...
+        {
+            let mut options = packet.options_mut()?;
+            let tmp = options; options = DhcpOption::MessageType(self.message_type).emit(tmp);
+            if let Some(list) = self.parameter_request_list {
+                let option = DhcpOption::Other{ kind: field::OPT_PARAMETER_REQUEST_LIST, data: list };
+                let tmp = options; options = option.emit(tmp);
+            }
+            if let Some(ip) = self.requested_ip {
+                let tmp = options; options = DhcpOption::RequestedIp(ip).emit(tmp);
+            }
```
whitequark: This doesn't emit server identifier, which makes fuzzing harder; can't rely on roundtrips.

phil-opp:
Thanks for the review! I think I addressed all your comments.

phil-opp:
@whitequark Is there anything I should improve to get this PR merged?

src/wire/dhcp.rs
```diff
@@ -0,0 +1,629 @@
...
+    pub server_identifier: Option<Ipv4Address>,
+    pub parameter_request_list: Option<&'a [u8]>,
+}
+
+impl<'a> Repr<'a> {
+    /// Parse a DHCP packet and return a high-level representation.
+    pub fn parse<T>(packet: &Packet<&'a T>) -> Result<Self>
+            where T: AsRef<[u8]> + ?Sized {
+
+        let transaction_id = packet.transaction_id();
```
dlrobertson: The length of the packet has not been checked before this point right? This could be out of bounds right?

src/wire/dhcp.rs
```diff
@@ -0,0 +1,629 @@
...
+        })
+    }
+
+    /// Emit a high-level representation into a Transmission Control Protocol packet.
+    pub fn emit<T>(&self, packet: &mut Packet<&mut T>) -> Result<()>
+            where T: AsRef<[u8]> + AsMut<[u8]> + ?Sized {
+        packet.set_opcode(self.message_type.opcode());
+        packet.set_hardware_type(Hardware::Ethernet);
+        packet.set_hardware_len(6);
+        packet.set_transaction_id(self.transaction_id);
```
dlrobertson: Same here. If we're writing to a short buffer, this could be out of bounds.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,629 @@
...
+                length = *buffer.get(1).ok_or(Error::Truncated)? as usize;
+                let data = buffer.get(2..(2+length)).ok_or(Error::Truncated)?;
+                match (kind, length) {
+                    (field::OPT_END, _) |
+                    (field::OPT_PAD, _) =>
+                        unreachable!(),
+                    (field::OPT_DHCP_MESSAGE_TYPE, 1) => {
+                        option = DhcpOption::MessageType(MessageType::from(data[0]));
+                    },
+                    (field::OPT_REQUESTED_IP, 4) => {
```
dlrobertson: We also need to verify that the length of `data` is 4 right? In theory `length` could be 4, but `data.len()` could be something else right?

src/wire/dhcp.rs
```diff
@@ -0,0 +1,629 @@
...
+            &DhcpOption::Pad => 1,
+            &DhcpOption::MessageType(_) => 3,
+            &DhcpOption::RequestedIp(ip) | &DhcpOption::ServerIdentifier(ip) => {
+                2 + ip.as_bytes().len()
+            },
+            &DhcpOption::Other { data, .. } => 2 + data.len()
+        }
+    }
+
+    pub fn emit<'b>(&self, buffer: &'b mut [u8]) -> &'b mut [u8] {
```
dlrobertson: No validation of the write buffers length here either.

src/wire/dhcp.rs
phil-opp: Thanks for the review! I'm a bit confused, since my original PR had such a check. It was called a "useless check" [in the previous review](https://github.com/m-labs/smoltcp/pull/75#discussion_r153168767), so I removed it. Should I add it again?

dlrobertson: When in doubt go with what @whitequark says. I am curious as to why that check is useless though. I would think the following would cause an out of bounds access, but I could be mistaken.

```rust
let bytes = [0x00; 4];
let packet = DhcpPacket::new(&bytes[..])
DhcpRepr::parse(packet)
```

src/wire/dhcp.rs
```diff
@@ -0,0 +1,629 @@
...
+            }
+            if let Some(ip) = self.requested_ip {
+                let tmp = options; options = DhcpOption::RequestedIp(ip).emit(tmp);
+            }
+            DhcpOption::EndOfList.emit(options);
+        }
+
+        Ok(())
+    }
+}
```
dlrobertson: The other `Repr` and `Packet` types in `wire` have tests. It might be good to add some basic unit tests.

src/wire/dhcp.rs
```diff
@@ -0,0 +1,629 @@
...
+    pub const OPT_PARAMETER_REQUEST_LIST: u8 = 55;
+    pub const OPT_MESSAGE: u8 = 56;
+    pub const OPT_MAX_DHCP_MESSAGE_SIZE: u8 = 57;
+    pub const OPT_RENEWAL_TIME_VALUE: u8 = 58;
+    pub const OPT_REBINDING_TIME_VALUE: u8 = 59;
+    pub const OPT_VENDOR_CLASS_ID: u8 = 60;
+    pub const OPT_CLIENT_ID: u8 = 61;
+}
+
+impl<T: AsRef<[u8]>> Packet<T> {
```
dlrobertson: Most the other `Packet` types document the methods. The `Repr` is documented well, but it would be nice if this was as well.

batonius:
@phil-opp @whitequark  What's the status of the PR? I could use a DHCP implementation :)

phil-opp:
I'm happy to update and adjust this PR as soon as there's desire to merge it. In https://github.com/m-labs/smoltcp/issues/63#issuecomment-339422852 @whitequark said that he wants to do the next step himself, maybe he wants to wait until the other parts are ready?

whitequark:
@phil-opp Can you add tests for parsing/emission and documentation for methods?

Also, I'm not a he.

phil-opp:
> Can you add tests for parsing/emission and documentation for methods?

Will do! What about the length checks (see https://github.com/m-labs/smoltcp/pull/75#discussion_r156649605)?

> Also, I'm not a he.

Sorry about that!

whitequark:
> What about the length checks (see #75 (comment))?

The length check should live in `Packet::check_len`, not `Repr::parse`, like it does everywhere else.

phil-opp:
Added tests and rebased on top of master. I try to add the method documentation soon.

whitequark:
Thanks! FYI the tests are failing.

dlrobertson:
Due to the addition of `proto-ipv6` since DHCP is IPv4 only the module and `pub use` should probably be `#[cfg(feature = "proto-ipv4")]`. Speaking of that, should the module be dhcpv4?

whitequark:
Yes to both.

phil-opp:
@whitequark I added the `cfg` gate, renamed the module, and documented the methods. I also rebased on master and squashed the commits.

whitequark:
This looks pretty good to me!
@phil-opp I think you didn't squash the commits after all.
@dlrobertson r? anything I've missed?

phil-opp:
Oh, I just thought that I leave the latest commits unsquashed to make the review easier.

I rebased and squashed again, but left @little-dude 's work (to keep author information) and the change to the `enum_with_unknown` macro (a change that has not much to do with dhcp) separate. I can squash them into a single commit if you like.

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit fe3f2d2 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit fe3f2d20f4d51e9070ea1b08361942a0ab750875 with merge 2582399191c2cc814e9c22da69acf9e52571dd56...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/352042815?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 2582399191c2cc814e9c22da69acf9e52571dd56 to master...

PR #78, opened 2017-11-12
Don't end ICMP packet processing with ICMP sockets. Ignore `Error::Unaddressable` for ICMP sockets.

batonius:
**Context:** I'm using ICMP sockets to implement `ping` in Redox.

**Problem:** Echo Requests to localhost fail because they are processed by the same ICMP socket which sent them in the first place, and `smoltcp` doesn't get a chance to respond to them with Echo Replies. The same problem could arise if another host sends Echo Request with an `ident` value we have an ICMP socket bound to.

**Solution:** Copy the raw sockets approach, whereby we dispatch a packet to all ICMP sockets and then let `smoltcp` handle it as well. Keep a flag to indicate whether a packet of unknown type has been processed by at least one ICMP socket, return `Error::Unrecognized` otherwise.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -227,11 +227,13 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
                     Socket::Icmp(ref mut socket) =>
                         socket.dispatch(&caps, |response| {
                             let tx_token = device.transmit().ok_or(Error::Exhausted)?;
-                            match response {
-                                (IpRepr::Ipv4(repr), icmp_repr) =>
-                                    inner.dispatch(tx_token, timestamp, Packet::Icmpv4((repr, icmp_repr))),
+                            device_result = match response {
+                                (IpRepr::Ipv4(ipv4_repr), icmpv4_repr) =>
+                                    inner.dispatch(tx_token, timestamp,
+                                                   Packet::Icmpv4((ipv4_repr, icmpv4_repr))),
                                 _ => Err(Error::Unaddressable),
-                            }
+                            };
+                            device_result
                         }),
                     #[cfg(feature = "socket-udp")]
                     Socket::Udp(ref mut socket) =>


src/iface/ethernet.rs
```diff
@@ -1049,34 +1063,38 @@ mod test {
...
         // socket.
         {
             assert!(!socket_set.get::<IcmpSocket>(socket_handle).can_recv());
         }
-        assert_eq!(iface.inner.process_icmpv4(&mut socket_set, ip_repr, data),
-                   Ok(Packet::None));
+
+        // Confirm we still get EchoReply from `smoltcp` even with the ICMP socket listening
+        let echo_reply = Icmpv4Repr::EchoReply{ ident, seq_no, data: echo_data };
+        let mut ipv4_reply = ipv4_repr.clone();
```
dlrobertson: Might be more readable to just write this out

src/iface/ethernet.rs
```diff
@@ -1049,34 +1063,38 @@ mod test {
...
+        assert_eq!(iface.inner.process_icmpv4(&mut socket_set, ip_repr, icmp_data),
+                   Ok(Packet::Icmpv4((ipv4_reply, echo_reply))));
+
         {
             let mut socket = socket_set.get::<IcmpSocket>(socket_handle);
             assert!(socket.can_recv());
             assert_eq!(socket.recv(),
-                       Ok((&data[..], IpAddress::Ipv4(Ipv4Address::new(0x7f, 0x00, 0x00, 0x02)))));
+                       Ok((&icmp_data[..],
+                           IpAddress::Ipv4(Ipv4Address::new(0x7f, 0x00, 0x00, 0x02)))));
```
dlrobertson: This IP is used multiple times. Might be worthwhile to make this a variable that is used.

src/iface/ethernet.rs
batonius: Agree.

batonius:
**Problem 2**: ICMP sockets don't update `device_result` in `socket_egress`, so their `Error::Unaddressable` isn't ignored, resulting in early returns from `socket_egress` and `poll`, preventing the second from processing ARP responses, which results in ARP flood until the socket is open.

**Solution 2**: Update `device_result` for ICMP sockets.

batonius:
@whitequark By the way, why do we break from the loop in `socket_egress` on `Error::Unaddressable`? There could be other sockets left with endpoints resolved, and by breaking here we prevent them from sending packets until all the sockets before them have their destinations resolved.

src/iface/ethernet.rs
```diff
@@ -241,11 +241,12 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
                     Socket::Icmp(ref mut socket) =>
                         socket.dispatch(&caps, |response| {
                             let tx_token = device.transmit().ok_or(Error::Exhausted)?;
-                            match response {
+                            device_result = match response {
```
dlrobertson: Good catch

src/iface/ethernet.rs
```diff
@@ -241,11 +241,12 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
...
                                 (IpRepr::Ipv4(repr), icmp_repr) =>
```
whitequark: Now that I look at it, can any of you explain what this `match` statement does?

whitequark:
> By the way, why do we break from the loop in socket_egress on Error::Unaddressable? There could be other sockets left with endpoints resolved, and by breaking here we prevent them from sending packets until all the sockets before them have their destinations resolved.

I'm currently working on fixing ARP floods, and once that's done I think there won't be any Error::Unaddressable; so disregard this for now.

src/iface/ethernet.rs
batonius: As far as I understand, since `ICMPv4` and `ICMPv6` are somewhat different (unlike TCP and UDP), this match guarantees we respond to an `ICMPv4` packet with `ICMPv4`. Specifically, we shouldn't return `Packet::Icmpv4` for IPv6 packets.

whitequark: Oh. Can you rename the captured variables to `ipv4_repr` and `icmpv4_repr`? That makes the intent obvious.

batonius: Done.

PR #79, opened 2017-11-12
Make socket errors in `socket_egress` non-critical.

batonius:
**Problem:** Right now a socket error in `EthernetInterface::socket_egress` results in both `socket_egress` and `poll` returning early, which prevents `socket_egress` from processing other sockets and `poll` from processing ingress packets. For example, if a socket tries to send a packet to an unresolved host, `inner.dispatch` returns `Error::Unaddressable`, which shortcuts `socket_egress` and `poll`, which prevents ingress sockets from being processed, and host's MAC from being resolved, while the socket exists.

**Solution**: Make socket errors non-critical, a socket error shouldn't affect other sockets or ingress packet processing. 

**Other:** I'm not sure if this is the best solution, but the problem is real, specifically the situation described above.


batonius:
False alarm, the issue is with ICMP sockets not setting `device_result` in `socket_egress`, `Error::Unaddressable` in `device_result` is already being ignored. Will add the fix to #78.

PR #80, opened 2017-11-13
Preserve retransmission timer for Dup-ACKs

pothos:
Duplicate ACKs should not replace the retransmission timer,
but if not in retransmission, still set the keep-alive timer as
normal ACKs do.


whitequark:
Thank you very much!

Issue #87, opened 2017-11-30
Closure based polling on ethernet interface


phil-opp:
At the moment, polling the ethernet interface works by passing in a `SocketSet`, which is passed down to the `process_*` methods (e.g. `process_udp`). The `process_*` methods then iterate over the socket set and pass the packet to all sockets that accept it.

The problem with this approach is that the socket interface requires buffering in the sockets and polling the sockets later, which is undesirable in some situations. For example, if we send a control command to a microcontroller, we want it to react as soon as it receives the packet and avoid any latency caused by buffering and polling.

Therefore I would propose to change the signature of the `EthernetInterface::poll` method, so that it takes a closure:

```rust
fn poll<F>(&mut self, f: F, timestamp: u64) -> Result<Option<u64>>
    where F: FnOnce(IpAddress, Port, Packet)
```

(The exact types for `Port` and  `Packet` still need to be determined.)

This closure can then be passed down to the `process_*` methods and invoked for each packet. This would allow the user to react immediately to certain packets and e.g. put the other ones in socket buffers. We could also add a `poll_sockets` method with the current behavior (which calls `poll` with a closure that iterates over a socket set).

What do you think about such a change?

cc @oli-obk

whitequark:
I don't think FnOnce works in this signature.

phil-opp:
> I don't think FnOnce works in this signature.

Why not? The function would only handle a single packet per call (like it does today), so it would only invoke the closure once.

> It's really unclear to me how this is supposed to work lifetime-wise too...

I try to implement a prototype. Where do you expect lifetime related problems?

whitequark:
> Why not? The function would only handle a single packet per call (like it does today), so it would only invoke the closure once.

How so? Right now poll() will pull all packets that are available from the device and process them.

phil-opp:
Ok, then I misunderstood the interface and we need `FnMut`.

PR #90, opened 2017-12-06
Prototype for closure based polling

phil-opp:
Changes the signature of `EthernetInterface::poll` to:

```rust
pub fn poll<F>(&mut self, mut f: F, timestamp: u64) -> Result<()>
        where F: for<'f> FnMut(Packet<'f>) -> Result<Packet<'f>>
{…}
```

This allows more flexible packet handling. The old socket based polling is still available as `poll_sockets` (under the hood it invokes `poll` with a closure that does the socket processing).

The idea was proposed in #87.

## Open Questions

- [ ] Reuse the `Packet` type for poll and make it public? Or introduce a separate type?
  - [ ] How to handle ICMP sockets with the packet type? ICMP sockets need access to the IP payload (see the `FIXME` in the PR), which isn't stored in the Packet::Icmpv4 variant.
- [ ] Are the socket_* features for the protocol parsing or just for enabling the socket types? E.g. should smoltcp still process TCP packets without the socket_tcp feature (and just don't include the tcp socket type)?
- [ ] Swap order of `f` and `timestamp` parameters for more ergonomic calling?

## Todo
- [ ] Fix tests
- [ ] Fix socket_* features
- [ ] Document `poll`

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -16,16 +16,10 @@ use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
 use wire::{UdpPacket, UdpRepr};
 #[cfg(feature = "socket-tcp")]
 use wire::{TcpPacket, TcpRepr, TcpControl};
-
-use socket::{Socket, SocketSet, AnySocket};
-#[cfg(feature = "socket-raw")]
-use socket::RawSocket;
-#[cfg(feature = "socket-icmp")]
-use socket::IcmpSocket;
-#[cfg(feature = "socket-udp")]
-use socket::UdpSocket;
 #[cfg(feature = "socket-tcp")]
 use socket::TcpSocket;
+
+use socket::{Socket, SocketSet};
 use super::{NeighborCache, NeighborAnswer};
 
 /// An Ethernet network interface.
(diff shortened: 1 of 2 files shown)


dlrobertson:
> How to handle ICMP sockets with the packet type? ICMP sockets need access to the IP payload (see the FIXME in the PR), which isn't stored in the Packet::Icmpv4 variant.

It would be very easy to update the signature of `IcmpSocket::process`. When I created ICMP sockets I didn't use the icmp repr in `process` because I wanted to avoid having to `emit` it to the `rx_buffer`. If it is agreed that the signature of `IcmpSocket::process` should change, I'd rather submit a PR to the upstream repo because it is a unrelated change that shouldn't take too long to merge.

whitequark:
I'm not sure how I feel about this change. It adds a lot of indirection and the benefit is not obvious. Perhaps you could present a concrete use case?

phil-opp:
The current socket API requires regular polling, which causes overhead (due to unnecessary polling when no new data is there) and latency (the time between data arrival and the next poll).

Ideally, we would like an interface that allows the following sequence:

1. Packets arrive
2. Hardware triggers interrupt
3. Interrupt service routine triggers task that polls the ethernet interface
4. The closure passed to `poll` reacts immediately to every received packet. It can continue exactly the tasks that are waiting for the packet. It can also react to control commands without latency (e.g. to a `reset` command).

whitequark:
> The closure passed to poll reacts immediately to every received packet. It can continue exactly the tasks that are waiting for the packet. It can also react to control commands without latency (e.g. to a reset command).

That looks like you don't want to use the socket part of `smoltcp` at all. This is not compatible with e.g. using TCP.

whitequark:
> It can continue exactly the tasks that are waiting for the packet.

Regarding this part, it seems to me that rather this sort of closure hack, you want to be able to pass a callback to SocketSet, so that when socket readiness changes, you can resume a task. Right?

phil-opp:
> That looks like you don't want to use the socket part of smoltcp at all. This is not compatible with e.g. using TCP.

Good point. I was only thinking about UDP. I refractored the code a bit, so that there's a `SocketSet::handle` method now. So you can do:

```rust
eth_interface.poll(|packet| match packet {
    Packet::Udp(_) if ... => // do something
    other => socket_set.handle(other,...)
},...)
```

> you want to be able to pass a callback to SocketSet, so that when socket readiness changes, you can resume a task. Right?

That would be useful, too. However, since the socket set lives for the whole time, we can't temporarily borrow variables in it. With the closure based approach, captured variables remain usable outside the `poll` call.

whitequark:
> Good point. I was only thinking about UDP.

I don't think I can merge a solution that changes (and likely pessimizes, via the additional indirection) all codepaths but only actually works for UDP. It sounds like you would be much better served by a middleware `Device` implementation that looks at packet headers and does something if it detects an UDP packet, and passes the packet through otherwise.

> However, since the socket set lives for the whole time, we can't temporarily borrow variables in it. With the closure based approach, captured variables remain usable outside the poll call.

Elaborate?

phil-opp:
The indirection is likely optimized away by the compiler, since everything is generic.

>  It sounds like you would be much better served by a middleware Device implementation that looks at packet headers and does something if it detects an UDP packet, and passes the packet through otherwise.

The problem with this approach is that we would need to parse every packet twice. A better point for this middleware would be between the `EthernetInterface` and the sockets, after the parsing happened.

> Elaborate?

For example, with the closure based interface we can do something like this:

```rust
let mut task_runner = …;
eth_interface.poll(|packet| match packet {
    Packet::Udp(_) if ... => task_runner.activate_task(…),
    other => socket_set.handle(other,...)
},...);
task_runner.run_next_task();
```

The task_runner is only borrowed for the `poll` call, so we can use it afterwards again. With a callback on the socket set, this wouldn't be possible:
```rust
let mut task_runner = …;
socket_set.set_callback(|packet| match packet {
    Packet::Udp(_) if ... => task_runner.activate_task(…),
    other => _
});
eth_interface.poll(socket_set ,...);
task_runner.run_next_task(); // compile error, since task_runner is still borrowed by the closure.
```

whitequark:
How about this interface instead?
```rust
trait PacketFilter {
  fn process_udp(UdpPacket) -> UdpPacket;
  fn process_tcp(TcpPacket) -> TcpPacket;
  // ...
}
impl PacketFilter for SocketSet {
  // ...
}
```
It's more clear how it works, it's more clear what the purpose is, there is no intermediate enum being created and matched on, and it is obvious that it will be optimized by the compiler.

phil-opp:
I need to think about it a bit more, but it seems like a good solution. It should allow us to use a closure based interface if we want to:

```rust
struct<F>Closure(F);
impl<F> PacketFilter for Closure<F> where F: FnMut(…) -> … {
    fn process_udp(UdpPacket) -> UdpPacket {
        if udp_port == x {
           (self.0)(…)
        } else { … }
    }
    …
}
```

whitequark:
Closing in favor of #94.

PR #91, opened 2017-12-07
Process the Icmpv4Repr in IcmpSocket::process

dlrobertson:
## Proposed change

Use the `Icmpv4Repr` in `IcmpSocket::process` instead of the raw IP payload (See #90).

## Drawbacks:

The ICMP packet needs to be parsed for `IcmpSocket::accept`, so the packet has already been parsed before `IcmpSocket::process` is called. As a result, most overhead incurred by such this change is due to the `Icmpv4Packet::emit` call. The main purpose of this PR is to aide with #90. When not considering #90, I'm not sure what I think about the updated `IcmpSocket::process`.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -6,6 +6,7 @@ use managed::ManagedSlice;
 
 use {Error, Result};
 use phy::{Device, DeviceCapabilities, RxToken, TxToken};
+use wire::pretty_print::PrettyPrinter;
 use wire::{EthernetAddress, EthernetProtocol, EthernetFrame};
 use wire::{Ipv4Address};
 use wire::{IpAddress, IpProtocol, IpRepr, IpCidr};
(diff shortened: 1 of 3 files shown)


whitequark:
This means you lose the ability to send ICMP packets smoltcp does not know about.

whitequark:
But in principle, we could just add an `Icmpv4Repr::Unknown(&[u8])`, I'll be fine with that.

dlrobertson:
> This means you lose the ability to send ICMP packets smoltcp does not know about.

AFAIK this is already the case. We currently cannot `send` a icmp message that isn't a variant of `Icmpv4Repr` due to the current implementation of [dispatch](https://github.com/m-labs/smoltcp/blob/master/src/socket/icmp.rs#L337). We cannot `recv` a message that is not a Echo Request/Reply or a Destination Unreachable due to the way [bind](https://github.com/m-labs/smoltcp/blob/master/src/socket/icmp.rs#L211-L221) and [accepts](https://github.com/m-labs/smoltcp/blob/master/src/socket/icmp.rs#L292-L314) works. Not saying this is the correct implementation, but I think it is the state of things currently.

> But in principle, we could just add an Icmpv4Repr::Unknown(&[u8]), I'll be fine with that.

Yeah, this would allow us to send arbitrary packets. We will need a generic `IcmpRepr` as well to encompass both ICMPv4 and ICMPv6 in the near future.

whitequark:
> AFAIK this is already the case. We currently cannot send a icmp message that isn't a variant of Icmpv4Repr due to the current implementation of dispatch.

You are of course correct.

Issue #47, opened 2017-10-03
Convert EthernetInterface to use a builder pattern


whitequark:
The type signature of `new` is getting _really_ unwieldy, and the breaking changes for every new features are undesirable.

dlrobertson:
This is going to be super important as we move towards adding IPv6 support. Especially once builds without the future `proto-ipv4` become a possibility.

dlrobertson:
In fact, this should probably be done before we add the `proto-ipv4` feature.

PR #92, opened 2017-12-11
Deprecate EthernetInterface::new for a builder

dlrobertson:
## Overview

Create a `EthernetInterfaceBuilder` structure and deprecate the use of
`EthernetInterface::new`.

This was the best interface I could come up with. Since we have members that
need to be consumed. I think `finalize` and the setters need to consume the
builder.

## Drawbacks

I didn't see a way to make this anything other than a consuming builder. I don't
think there will be any real drawbacks here since we are creating the builder to
allow more flexibility in specifying options instead of using the builder to create
a template which is used to create multiple similar interfaces.

Fixes: #47

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -193,6 +193,21 @@ enum Packet<'a> {
     Tcp((IpRepr, TcpRepr<'a>))
 }
 
+impl<'a> Packet<'a> {
+    fn neighbor_addr(&self) -> Option<IpAddress> {
+        match self {
+            &Packet::None | &Packet::Arp(_) => None,
+            &Packet::Icmpv4((ref ipv4_repr, _)) => Some(ipv4_repr.dst_addr.into()),
+            #[cfg(feature = "socket-raw")]
+            &Packet::Raw((ref ip_repr, _)) => Some(ip_repr.dst_addr()),
+            #[cfg(feature = "socket-udp")]
+            &Packet::Udp((ref ip_repr, _)) => Some(ip_repr.dst_addr()),
+            #[cfg(feature = "socket-tcp")]
+            &Packet::Tcp((ref ip_repr, _)) => Some(ip_repr.dst_addr())
+        }
+    }
+}
+
 impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
         where DeviceT: for<'d> Device<'d> {
     /// Get the Ethernet address of the interface.
@@ -266,7 +281,11 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
         if self.socket_ingress(sockets, timestamp)? {
             Ok(Some(0))
         } else {
-            Ok(sockets.iter().filter_map(|socket| socket.poll_at()).min())
+            Ok(sockets.iter().filter_map(|socket| {
+                let socket_poll_at = socket.poll_at();
+                socket.meta().poll_at(socket_poll_at, |ip_addr|
+                    self.inner.has_neighbor(&ip_addr, timestamp))
+            }).min())
         }
     }
 
@@ -302,12 +321,12 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
         caps.max_transmission_unit -= EthernetFrame::<&[u8]>::header_len();
 
         for mut socket in sockets.iter_mut() {
-            if let Some(hushed_until) = socket.meta().hushed_until {
-                if hushed_until > timestamp {
-                    continue
-                }
+            if !socket.meta_mut().egress_permitted(|ip_addr|
+                    self.inner.has_neighbor(&ip_addr, timestamp)) {
+                continue
             }
 
+            let mut neighbor_addr = None;
             let mut device_result = Ok(());
             let &mut Self { ref mut device, ref mut inner } = self;
             let socket_result =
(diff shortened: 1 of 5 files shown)


src/iface/ethernet.rs
```diff
@@ -74,6 +74,7 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
     /// # Panics
     /// See the restrictions on [set_hardware_addr](#method.set_hardware_addr)
     /// and [set_protocol_addrs](#method.set_protocol_addrs) functions.
+    #[deprecated(note = "please use `EthernetInterfaceBuilder` instead")]
```
whitequark: Just kill it, we don't have stable interfaces.

src/iface/mod.rs
```diff
@@ -5,8 +5,10 @@
 
 mod neighbor;
 mod ethernet;
+mod ethernetbuilder;
```
whitequark: Put it into `ethernet`, you don't need any of these `pub(super)` annotations.

src/iface/ethernetbuilder.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// Set the Neighbor Cache the interface will use.
+    pub fn neighbor_cache(mut self, neighbor_cache: NeighborCache<'b>) -> InterfaceBuilder<'b, 'c, DeviceT> {
+        self.neighbor_cache = Some(neighbor_cache);
+        self
+    }
+
+    /// Create a network interface using the previously provided configuration.
+    ///
+    /// Note: Returns `None` if a required option is not provided. Required
+    /// options include
```
whitequark: If a required option is not provided it's a programmer error, so it should panic.

src/iface/ethernetbuilder.rs
```diff
@@ -0,0 +1,154 @@
...
+            ethernet_addr:       None,
+            neighbor_cache:      None,
+            ip_addrs:            None,
+            ipv4_gateway:        None
+        }
+    }
+
+    /// Set the Ethernet address the interface will use.
+    ///
+    /// # Panics
```
whitequark: All these functions should probably reference the ones on `EthernetInterface` so that the docs don't go out of sync.

src/iface/ethernet.rs
```diff
@@ -759,11 +732,146 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+    /// [ipv4_gateway].
+    ///
+    /// # Panics
+    /// This function panics if the given address is not unicast.
+    ///
+    /// [ipv4_gateway]: struct.EthernetInterface.html#method.ipv4_gateway
+    pub fn ipv4_gateway<T>(mut self, gateway: T) -> InterfaceBuilder<'b, 'c, DeviceT>
+            where T: Into<Ipv4Address> {
+        let addr = gateway.into();
+        if !addr.is_unicast() {
```
whitequark: I think we should have `InterfaceInner::check_gateway_addr` instead of this.

src/iface/ethernet.rs
```diff
@@ -759,11 +732,146 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+    pub fn neighbor_cache(mut self, neighbor_cache: NeighborCache<'b>) -> InterfaceBuilder<'b, 'c, DeviceT> {
+        self.neighbor_cache = Some(neighbor_cache);
+        self
+    }
+
+    /// Create a network interface using the previously provided configuration.
+    ///
+    /// # Panics
+    /// If a required option is not provided, this function will panic. Required
+    /// options include.
```
whitequark: `Required options are:` (wording and typography)

src/iface/ethernet.rs
```diff
@@ -759,11 +732,146 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+    ///
+    /// - [ethernet_addr]
+    /// - [neighbor_cache]
+    /// - [ip_addrs]
+    ///
+    /// [ethernet_addr]: #method.ethernet_addr
+    /// [neighbor_cache]: #method.neighbor_cache
+    /// [ip_addrs]: #method.ip_addrs
+    pub fn finalize(self) -> Interface<'b, 'c, DeviceT> {
+        // TODO: Limit the number of required options.
```
whitequark: What do you mean by this?

src/iface/ethernet.rs
```diff
@@ -759,11 +732,146 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+    }
+}
+
+
 #[cfg(test)]
 mod test {
     use std::collections::BTreeMap;
     use {Result, Error};
 
+    use super::super::EthernetInterfaceBuilder;
```
whitequark: Just `super::InterfaceBuilder`, no?

src/iface/ethernet.rs
dlrobertson: For example, I wondered if it is maybe possible to `finalize` without `ip_addrs` so that.

```rust
EthernetInterfaceBuilder::new(Loopback::new())
        .ip_addrs([])
        ...
```

Becomes

```rust
EthernetInterfaceBuilder::new(Loopback::new())
        ...
```

whitequark: Ah sure, that works.

dlrobertson:
Updated

dlrobertson:
Thanks for the updates

PR #94, opened 2017-12-13
Add a PacketFilter trait to abstract over concrete SocketSet type

phil-opp:
Alternative to #90 (proposed in https://github.com/m-labs/smoltcp/pull/90#issuecomment-349711030)

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -317,7 +317,7 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
     pub fn poll_at<F>(&self, filter: &F, timestamp: u64) -> Option<u64>
         where F: PacketFilter
     {
-        filter.poll_at(timestamp)
+        filter.poll_at(timestamp, self)
     }
 
     /// Return an _advisory wait time_ for calling [poll] the next time.
@@ -374,31 +374,42 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
         filter.egress(&caps, timestamp, self)
     }
 
-    fn emit(&mut self, response: Packet, timestamp: u64) -> Result<()> {
+    fn emit(&mut self, response: Packet, timestamp: u64) -> Result<Option<IpAddress>> {
         let &mut Self { ref mut device, ref mut inner } = self;
         let neighbor_addr = response.neighbor_addr();
         let tx_token = device.transmit().ok_or(Error::Exhausted)?;
-        inner.dispatch(tx_token, timestamp, response)
+        inner.dispatch(tx_token, timestamp, response).map(|()| neighbor_addr)
     }
 }
 
 impl<'b, 'c, DeviceT> PacketEmitter for Interface<'b, 'c, DeviceT>
         where DeviceT: for<'d> Device<'d>
 {
-    fn emit_tcp(&mut self, response: (IpRepr, TcpRepr), timestamp: u64) -> Result<()> {
+    fn emit_tcp(&mut self, response: (IpRepr, TcpRepr), timestamp: u64) -> Result<Option<IpAddress>> {
         self.emit(Packet::Tcp(response), timestamp)
     }
 
-    fn emit_udp(&mut self, response: (IpRepr, UdpRepr), timestamp: u64) -> Result<()> {
+    fn emit_udp(&mut self, response: (IpRepr, UdpRepr), timestamp: u64) -> Result<Option<IpAddress>> {
         self.emit(Packet::Udp(response), timestamp)
     }
 
-    fn emit_icmpv4(&mut self, response: (Ipv4Repr, Icmpv4Repr), timestamp: u64) -> Result<()> {
+    fn emit_icmpv4(&mut self, response: (Ipv4Repr, Icmpv4Repr), timestamp: u64) -> Result<Option<IpAddress>> {
         self.emit(Packet::Icmpv4(response), timestamp)
     }
 
-    fn emit_raw(&mut self, response: (IpRepr, &[u8]), timestamp: u64) -> Result<()> {
+    fn emit_raw(&mut self, response: (IpRepr, &[u8]), timestamp: u64) -> Result<Option<IpAddress>> {
         self.emit(Packet::Raw(response), timestamp)
+    }
+
+    fn has_neighbor<'a>(&self, addr: &'a IpAddress, timestamp: u64) -> bool {
+        match self.inner.route(addr) {
+            Ok(routed_addr) => {
+                self.inner.neighbor_cache
+                    .lookup_pure(&routed_addr, timestamp)
+                    .is_some()
+            }
+            Err(_) => false
+        }
     }
 }
 
@@ -832,17 +843,6 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
                 None => Err(Error::Unaddressable),
             }
             _ => Err(Error::Unaddressable)
-        }
-    }
-
-    fn has_neighbor<'a>(&self, addr: &'a IpAddress, timestamp: u64) -> bool {
-        match self.route(addr) {
-            Ok(routed_addr) => {
-                self.neighbor_cache
-                    .lookup_pure(&routed_addr, timestamp)
-                    .is_some()
-            }
-            Err(_) => false
         }
     }
 
--- a/src/iface/mod.rs
+++ b/src/iface/mod.rs
@@ -15,7 +15,7 @@ pub use self::ethernet::{Interface as EthernetInterface,
 
 use Result;
 use phy::{DeviceCapabilities, ChecksumCapabilities};
-use wire::{IpRepr, Ipv4Repr, UdpRepr, TcpRepr, Icmpv4Repr};
+use wire::{IpAddress, IpRepr, Ipv4Repr, UdpRepr, TcpRepr, Icmpv4Repr};
 
 pub trait PacketFilter {
     fn process_udp(&mut self, ip_repr: &IpRepr, udp_repr: &UdpRepr) -> Result<()>;
@@ -27,12 +27,13 @@ pub trait PacketFilter {
     fn egress<E>(&mut self, caps: &DeviceCapabilities, timestamp: u64, emitter: &mut E) -> Result<bool>
         where E: PacketEmitter;
 
-    fn poll_at(&self, timestamp: u64) -> Option<u64> { None }
+    fn poll_at<E>(&self, _timestamp: u64, _emitter: &E) -> Option<u64> where E: PacketEmitter { None }
 }
 
 pub trait PacketEmitter {
-    fn emit_tcp(&mut self, (IpRepr, TcpRepr), timestamp: u64) -> Result<()>;
-    fn emit_udp(&mut self, (IpRepr, UdpRepr), timestamp: u64) -> Result<()>;
-    fn emit_icmpv4(&mut self, (Ipv4Repr, Icmpv4Repr), timestamp: u64) -> Result<()>;
-    fn emit_raw(&mut self, (IpRepr, &[u8]), timestamp: u64) -> Result<()>;
+    fn emit_tcp(&mut self, (IpRepr, TcpRepr), timestamp: u64) -> Result<Option<IpAddress>>;
+    fn emit_udp(&mut self, (IpRepr, UdpRepr), timestamp: u64) -> Result<Option<IpAddress>>;
+    fn emit_icmpv4(&mut self, (Ipv4Repr, Icmpv4Repr), timestamp: u64) -> Result<Option<IpAddress>>;
+    fn emit_raw(&mut self, (IpRepr, &[u8]), timestamp: u64) -> Result<Option<IpAddress>>;
+    fn has_neighbor<'a>(&self, addr: &'a IpAddress, timestamp: u64) -> bool;
 }
--- a/src/socket/set.rs
+++ b/src/socket/set.rs
@@ -260,7 +260,7 @@ impl<'a, 'b: 'a, 'c: 'a + 'b> PacketFilter for Set<'a, 'b, 'c> {
         let mut emitted_any = false;
         for mut socket in self.iter_mut() {
             if !socket.meta_mut().egress_permitted(|ip_addr|
-                    self.inner.has_neighbor(&ip_addr, timestamp)) {
+                    emitter.has_neighbor(&ip_addr, timestamp)) {
                 continue
             }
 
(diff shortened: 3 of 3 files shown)


src/iface/ethernet.rs
```diff
@@ -161,17 +154,21 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
     /// packets containing any unsupported protocol, option, or form, which is
     /// a very common occurrence and on a production system it should not even
     /// be logged.
-    pub fn poll(&mut self, sockets: &mut SocketSet, timestamp: u64) -> Result<Option<u64>> {
+    pub fn poll<P>(&mut self, sockets: &mut P, timestamp: u64) -> Result<Option<u64>>
```
whitequark: `filter: &mut F`

src/iface/ethernet.rs
```diff
@@ -161,17 +154,21 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
...
         if self.socket_ingress(sockets, timestamp)? {
             Ok(Some(0))
         } else {
-            Ok(sockets.iter().filter_map(|socket| socket.poll_at()).min())
+            Ok(sockets.poll_at())
         }
     }
 
-    fn socket_ingress(&mut self, sockets: &mut SocketSet, timestamp: u64) -> Result<bool> {
+    fn socket_ingress<P>(&mut self, sockets: &mut P, timestamp: u64) -> Result<bool>
```
whitequark: As above.

src/iface/ethernet.rs
```diff
@@ -198,89 +195,18 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
         Ok(processed_any)
     }
 
-    fn socket_egress(&mut self, sockets: &mut SocketSet, timestamp: u64) -> Result<()> {
+    fn socket_egress<P>(&mut self, sockets: &mut P, timestamp: u64) -> Result<()>
```
whitequark: As above.

src/iface/ethernet.rs
```diff
@@ -305,9 +231,10 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         self.ip_addrs.iter().any(|probe| probe.address() == addr)
     }
 
-    fn process_ethernet<'frame, T: AsRef<[u8]>>
-                       (&mut self, sockets: &mut SocketSet, timestamp: u64, frame: &'frame T) ->
+    fn process_ethernet<'frame, P, T: AsRef<[u8]>>
+                       (&mut self, sockets: &mut P, timestamp: u64, frame: &'frame T) ->
```
whitequark: As above.

src/iface/ethernet.rs
```diff
@@ -368,10 +295,11 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         }
     }
 
-    fn process_ipv4<'frame, T: AsRef<[u8]>>
-                   (&mut self, sockets: &mut SocketSet, timestamp: u64,
+    fn process_ipv4<'frame, P, T: AsRef<[u8]>>
+                   (&mut self, sockets: &mut P, timestamp: u64,
```
whitequark: As above.

src/iface/mod.rs
```diff
@@ -10,3 +10,21 @@ pub use self::neighbor::Neighbor as Neighbor;
...
 pub use self::ethernet::Interface as EthernetInterface;
+pub use self::ethernet::Packet as EthernetPacket;
+
+use Result;
+use wire::{IpRepr, UdpRepr, TcpRepr, Icmpv4Repr};
+use phy::{DeviceCapabilities, ChecksumCapabilities};
+
+pub trait PacketFilter {
+    fn process_udp(&mut self, ip_repr: &IpRepr, repr: &UdpRepr) -> Result<()>;
+    fn process_tcp<'frame>(&mut self, timestamp: u64, ip_repr: &IpRepr, repr: &TcpRepr<'frame>) ->
```
whitequark: This really should not return an `EthernetPacket`. We may not even be running over Ethernet!

src/iface/mod.rs
```diff
@@ -10,3 +10,21 @@ pub use self::neighbor::Neighbor as Neighbor;
...
+        Result<EthernetPacket<'frame>>;
+    fn process_icmpv4(&mut self, ip_repr: &IpRepr, icmp_repr: &Icmpv4Repr, cksum: &ChecksumCapabilities) -> Result<()>;
+    fn process_raw(&mut self, ip_repr: &IpRepr, payload: &[u8],checksum_caps: &ChecksumCapabilities) -> Result<()>;
+
+    fn egress<F>(&mut self, caps: &DeviceCapabilities, timestamp: u64, f: F) -> Result<()>
```
whitequark: This too.

dlrobertson:
Our current implementation and this implementation include `process_ethernet`, `process_ip`, and `process_udp` all in on `trait` or `impl`. At some point we're going to need the `process_ip` and `process_ethernet` implementations to be in different `trait`s or `impl`s right? We shouldn't have layer 2 processing and layer 3 processing in the same trait right?

A bit of context: I'm thinking of the `skb` in linux and the transmission path it goes through.

whitequark:
Yeah, there shouldn't be a `process_ethernet` in this trait either, I've overlooked this somehow.

As for `skb`, the comparison doesn't take hold because we form packets just-in-time. This is more complex, but ultimately far more efficient; have you seen how large `sk_buff` is?

dlrobertson:
> have you seen how large sk_buff is?

lol

phil-opp:
Sorry for the delay! I renamed `sockets` to `filter` and the generic parameter for `PacketFilter` from `P` to `F`, and I changed the return type of `process_tcp`.

I'm not sure what to do about the `EthernetPacket` in `egress`. We can't lock it down to a single type, since the function is invoked with UDP, TCP, ICMP, and raw packets. So we need some kind of packet enum. Instead of reusing `EthernetPacket`, we could move these four variants to a more general `Packet` type, but I'm not sure if it's worth it. What do you think, @whitequark?

src/socket/set.rs
```diff
@@ -174,6 +178,142 @@ impl<'a, 'b: 'a, 'c: 'a + 'b> Set<'a, 'b, 'c> {
...
+            match raw_socket.process(&ip_repr, ip_payload, &checksum_caps) {
+                // The packet is valid and handled by socket.
+                Ok(()) => handled = true,
+                // The socket buffer is full.
+                Err(Error::Exhausted) => (),
+                // Raw sockets don't validate the packets in any way.
+                Err(_) => unreachable!(),
+            }
+        }
+        if handled { Ok(()) } else { Err(Error::Dropped) }
```
whitequark: This is quite confusing. A packet not being accepted by any raw socket is not an error condition; this is how `process_raw` is different from e.g. `process_tcp`. Why not just use `Result<bool>` for the return value of this function? Then it'd just be `Ok(handled)`...

There are three different semantics here. A handler for raw sockets can:
  * take the packet and eat it (putting it into its own buffer), which counts as success;
  * ignore the packet and let the higher layer handle it, which also counts as success;
  * deliberately drop the packet (think of an application like a firewall).

Only the third one should result in an `Error::Dropped`, IMO.

src/socket/set.rs
```diff
@@ -174,6 +178,142 @@ impl<'a, 'b: 'a, 'c: 'a + 'b> Set<'a, 'b, 'c> {
...
+            match icmp_socket.process(&ip_repr, icmp_repr, checksum_caps) {
+                // The packet is valid and handled by socket.
+                Ok(()) => handled = true,
+                // The socket buffer is full.
+                Err(Error::Exhausted) => (),
+                // ICMP sockets don't validate the packets in any way.
+                Err(_) => unreachable!(),
+            }
+        }
+        if handled { Ok(()) } else { Err(Error::Dropped) }
```
whitequark: Same as for `process_raw`.

src/iface/mod.rs
```diff
@@ -10,3 +10,21 @@ pub use self::neighbor::Neighbor as Neighbor;
...
 pub use self::neighbor::Cache as NeighborCache;
 pub use self::ethernet::Interface as EthernetInterface;
+pub use self::ethernet::Packet as EthernetPacket;
+
+use Result;
+use wire::{IpRepr, UdpRepr, TcpRepr, Icmpv4Repr};
+use phy::{DeviceCapabilities, ChecksumCapabilities};
+
+pub trait PacketFilter {
+    fn process_udp(&mut self, ip_repr: &IpRepr, repr: &UdpRepr) -> Result<()>;
```
whitequark: We need to decide on one of the three possibilities here:
  * Methods of `PacketFilter` are `#[cfg]`d on the existing `socket-*` features. I am reluctant to go with this because you might want to e.g. set a raw packet filter without having the space overhead of having raw socket code in your firmware.
  * Methods of `PacketFilter` are `#[cfg]`d on new `proto-raw/tcp/udp/etc` features, and `EthernetInterface` is changed to also conditionally invoke these methods when the features are enabled. I am reluctant to go with this because of the combinatorial explosion we already have to deal with, but this will provide the most space savings all around.
  * Methods of `PacketFilter` are always present, and so are `EthernetInterface`'s internal code paths for dispatching to them. I am reluctant to go with this because this means we can't quite cut down e.g. TCP even when there's absolutely no need for it.

whitequark:
> I'm not sure what to do about the EthernetPacket in egress. We can't lock it down to a single type, since the function is invoked with UDP, TCP, ICMP, and raw packets. So we need some kind of packet enum. Instead of reusing EthernetPacket, we could move these four variants to a more general Packet type, but I'm not sure if it's worth it. What do you think, @whitequark?

How about adding a trait `PacketEmitter`, with four methods like `dispatch_tcp`, and pass a `&mut PacketEmitter` (really a `&mut EthernetInterfaceInner`) to `PacketFilter::egress`?

src/iface/mod.rs
phil-opp: Hmm… I don't think that variant 2 is worth the additional complexity. We can still implement it later if someone has a need for it. For me, variant 1 sounds most reasonable. However, my use case for smoltcp has all no_std features enabled, so I'm not in a good position to make this decision.

src/socket/set.rs
phil-opp: Makes sense. I'll change the return type to Result<bool>.

phil-opp:
I pushed a new version. Rebasing/merging led to a lot of conflicts, so I decided to implement it again instead. I added a `PacketEmitter` type and addressed your comments on `process_raw` and `process_icmpv4`.

I'm not sure about the code related to `has_neighbor` (see the second commit). For now I added the method to the emitter trait, but I'm not sure if this is the best solution.

src/iface/mod.rs
whitequark: Let's go with variant 1 for now, then. Since a `socket-udp` feature would always depend on `proto-udp` we can backwards compatibly update to 2 if it's ever needed.

whitequark:
> I'm not sure about the code related to has_neighbor (see the second commit). For now I added the method to the emitter trait, but I'm not sure if this is the best solution.

Now that I look through the design for this API again... I am having doubts that this is the best solution indeed. Having to proxy `poll_at` and `egress` through the `PacketFilter` trait should have been enough of an indication for me that this is trying to cut the abstractions that aren't really suited for this.

Let's try and design a simpler solution instead. How about this...

We do not touch `SocketSet` at all. `poll` still takes a `&mut SocketSet`. We add a new `PacketFilter` trait with `filter_raw`, `filter_icmpv4`, `filter_tcp` and `filter_udp` methods, which interpose at the corresponding stages in the packet processing; that is, `filter_ipv4` is called exactly at the moment when an IPv4 packet is parsed, `filter_tcp` is called after that once a TCP packet is parsed, and so on. The filter methods can return `Ok(())` to continue processing, or `Err(Error::Dropped)` to drop the packet.

There is no support for replying with another packet and aborting the processing. I've tried to design a flexible interface for that but gave up because of ownership issues. Do you need that?

The trait should look something like this:

```rust
trait PacketFilter {
  fn process_raw(&mut self, timestamp: u64, 
                 ip_repr: &IpRepr, payload: &[u8]) -> Result<()>;
  fn process_icmpv4(&mut self, timestamp: u64, 
                 ip_repr: &IpRepr, icmpv4_repr: &Icmpv4Repr) -> Result<()>;
  fn process_udp(&mut self, timestamp: u64, 
                 ip_repr: &IpRepr, udp_repr: &UdpRepr) -> Result<()>;
  fn process_tcp(&mut self, timestamp: u64, 
                 ip_repr: &IpRepr, tcp_repr: &TcpRepr) -> Result<()>;
}
```

phil-opp:
> Having to proxy poll_at and egress through the PacketFilter trait should have been enough of an indication for me that this is trying to cut the abstractions that aren't really suited for this.

Our use case requires more a custom SocketSet than a packet filter. I only chose the name because you proposed it as such. IMO, `PacketProcessor` would be a better name. Then the `egress` and `poll_at` methods make sense on this trait.

For example, imagine a custom packet processor that immediately reacts to certain UDP control messages and passes all other packets to an internal `SocketSet`. We want to override `poll_at` in this case to ensure that the reaction time to control packets stays below a certain limit.

Or maybe we don't want to use `SocketSet` at all and handle only some hardcoded types of UDP packets, because we have a device with very limited memory and want to avoid the packet buffer. With a custom `PacketProcessor`, this would be entirely possible.

So I don't think that this is cutting "abstractions that aren't really suited for this". It just allows users to use a custom packet processor instead of hardcoding the `SocketSet`.

whitequark:
> For example, imagine a custom packet processor that immediately reacts to certain UDP control messages and passes all other packets to an internal SocketSet.

Define "react"? If you don't need to reply then my proposed abstraction works.

> We want to override poll_at in this case to ensure that the reaction time to control packets stays below a certain limit.

`poll_at` serves exactly one purpose: an advisory value that can be used to let the device go into low-power mode (or a thread to sleep). You can simply ignore it and call `poll` as often as you want, right now.

phil-opp:
> Define "react"? If you don't need to reply then my proposed abstraction works.

For some commands a reply would make sense. For example "give me this sensor value" or "please send me all logged error messages".

> You can simply ignore it and call poll as often as you want, right now.

Yes, but why shouldn't we allow to override that advisory value to model custom demands?

I don't understand the problem with this PR. Is it just because of the `has_neighbor` related code? Or are there other issues? For me, this feels like a natural point to allow customization: Instead of allowing only the provided `SocketSet`, we make it possible to provide an own type with similar functionality, that can be specialized for different hardware, OSs, use cases, or other circumstances.

whitequark:
> I don't understand the problem with this PR. Is it just because of the has_neighbor related code? Or are there other issues?

That and multiple other issues. For example, it's not actually functional for your use case; you can see that there is no way to respond to UDP, ICMP or raw packets from a filter right now. Adding a return value like in `process_tcp` is not a solution because of lifetime issues: you cannot provide a payload unless that payload outlives the `process_tcp` frame, which makes being able to provide a payload almost useless. I tried to fix that but was not able to deal with the lifetime issues.

phil-opp:
Closing as stale

whitequark:
FYI I am still interested in getting something like this merged. I've looked at your code and I understand your use case much better now.

@astro @dlrobertson @phil-opp What do you think about meeting sometime on IRC and trying to brainstorm a good design for this? I'm afraid I'm running low on cycles here.

dlrobertson:
:+1: yeah that could be helpful

Issue #76, opened 2017-11-09
Add IPv6 to wire


dlrobertson:
- [x] Make IPv4 optional by adding a feature gating it. #98 
 - [x] Add Ipv6Address/Ipv6Cidr. #72 
 - [x] Add Ipv6Packet/Ipv6Repr. #89 

Metabug from https://github.com/m-labs/smoltcp/pull/72#issuecomment-342369380.

Related to #3

PR #98, opened 2017-12-22
Add the proto-ipv4 feature

dlrobertson:
Users should be able to choose to use smoltcp without IPv4 support.

Fixes: #76

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,22 +16,25 @@ matrix:
     - rust: nightly
       env: FEATURES='std phy-tap_interface proto-ipv6 socket-udp' MODE='test'
     - rust: nightly
-      env: FEATURES='std socket-raw proto-ipv4' MODE='test'
+      env: FEATURES='std proto-ipv4 socket-raw' MODE='test'
     - rust: nightly
-      env: FEATURES='std socket-udp proto-ipv6' MODE='test'
+      env: FEATURES='std proto-ipv6 socket-udp' MODE='test'
     - rust: nightly
-      env: FEATURES='std socket-tcp proto-ipv6' MODE='test'
+      env: FEATURES='std proto-ipv6 socket-tcp' MODE='test'
     - rust: nightly
-      env: FEATURES='std socket-icmp socket-tcp proto-ipv4' MODE='test'
+      env: FEATURES='std proto-ipv4 socket-icmp socket-tcp' MODE='test'
     - rust: nightly
-      env: FEATURES='std socket-icmp socket-tcp proto-ipv6' MODE='test'
+      env: FEATURES='std proto-ipv6 socket-icmp socket-tcp' MODE='test'
     ### Test select feature permutations, chosen to be as aggressive as possible
     - rust: nightly
-      env: FEATURES='socket-raw socket-udp socket-tcp socket-icmp proto-ipv4 std' MODE='test'
+      env: FEATURES='proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp std'
+        MODE='test'
     - rust: nightly
-      env: FEATURES='socket-raw socket-udp socket-tcp socket-icmp proto-ipv4 alloc' MODE='test'
+      env: FEATURES='proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp alloc'
+        MODE='test'
     - rust: nightly
-      env: FEATURES='socket-raw socket-udp socket-tcp socket-icmp' MODE='build'
+      env: FEATURES='proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp'
+        MODE='build'
 script:
   - cargo "$MODE" --no-default-features --features "$FEATURES"
 notifications:


src/socket/tcp.rs
```diff
@@ -1488,6 +1488,7 @@ impl<'a> fmt::Write for TcpSocket<'a> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "proto-ipv4")]
```
whitequark: Let's add `pub(crate) const MOCK_IP_ADDRESS_1` in `smoltcp::wire::ip::test` or something like this, it's not good to drop tests in IPv6...

src/socket/udp.rs
```diff
@@ -257,6 +257,7 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "proto-ipv4")]
```
whitequark: As mentioned elsewhere.

src/wire/tcp.rs
```diff
@@ -853,6 +853,7 @@ impl<T: AsRef<[u8]>> PrettyPrint for Packet<T> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "proto-ipv4")]
```
whitequark: As mentioned elsewhere.

src/wire/udp.rs
```diff
@@ -281,6 +282,7 @@ impl<T: AsRef<[u8]>> PrettyPrint for Packet<T> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "proto-ipv4")]
```
whitequark: As mentioned elsewhere.

src/iface/ethernet.rs
```diff
@@ -21,7 +25,7 @@ use wire::{TcpPacket, TcpRepr, TcpControl};
 use socket::{Socket, SocketSet, AnySocket};
 #[cfg(feature = "socket-raw")]
 use socket::RawSocket;
-#[cfg(feature = "socket-icmp")]
+#[cfg(all(feature = "socket-icmp", feature = "proto-ipv4"))]
```
whitequark: I don't think ICMP sockets should depend on `proto-ipv4`. Sure, right now, since we don't have ICMPv6 at all, they would do nothing, but they should eventually be available, so let's keep them in.

src/iface/neighbor.rs
```diff
@@ -156,6 +156,7 @@ impl<'a> Cache<'a> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "proto-ipv4")]
```
whitequark: As described elsewhere.

src/socket/raw.rs
```diff
@@ -240,6 +243,7 @@ impl<'a, 'b> RawSocket<'a, 'b> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "proto-ipv4")]
```
whitequark: As described elsewhere.

dlrobertson:
> I've revised required-features even further and added them upstream, so you can discard my commit here.

:+1: I'll rebase and force push

src/iface/ethernet.rs
dlrobertson: If we remove this that would mean a build without default features and without `proto-ipv4` but with `proto-ipv6` would fail. If you're okay with that, I can simply remove this. Otherwise we need to make a generic `IcmpRepr`/`IcmpPacket`

whitequark: Hmmm, I've looked at making ICMP sockets work with ICMPv6 but it looks involved, so let's disable them on non-IPv4 builds for now.

dlrobertson: Sounds good. I'll create an issue for this and https://github.com/dlrobertson/smoltcp/blob/proto-ipv4/src/lib.rs#L4-L5

src/socket/tcp.rs
dlrobertson: After looking at the tests, I'm not sure how a set of mock IP addresses would work. So I added the `ipv4` specific stuff in a `ipv4_locals` module. Would it be an public mock `IpAddress` be an `IPv6` variant if `proto-ipv6` was enabled and `IPv4` if not? Or would we have a public collection of mock IPv4 addresses and a public collection of IPv6 addresses.

dlrobertson:
Updated. I added the mock addrs to `wire::ip::test`. I did not update the `RawSocket` or `IcmpSocket` tests because we should have both IPv4 and IPv6 tests for `RawSocket`s and `IcmpSocket` does not support IPv6 at the moment.

BTW the `test_like_travis.rb` script is a lifesaver.

src/iface/ethernet.rs
```diff
@@ -921,8 +953,11 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         self.dispatch_ethernet(tx_token, timestamp, ip_repr.total_len(), |mut frame| {
             frame.set_dst_addr(dst_hardware_addr);
             match ip_repr {
+                #[cfg(feature = "proto-ipv4")]
                 IpRepr::Ipv4(_) => frame.set_ethertype(EthernetProtocol::Ipv4),
-                _ => unreachable!()
+                #[cfg(feature = "proto-ipv6")]
+                IpRepr::Ipv6(_) => frame.set_ethertype(EthernetProtocol::Ipv6),
+                _ => { return; }
```
whitequark: Nit: `_ => return`

src/lib.rs
```diff
@@ -1,6 +1,8 @@
 #![cfg_attr(feature = "alloc", feature(alloc))]
 #![no_std]
-#![deny(unsafe_code, unused)]
+#![deny(unsafe_code)]
+// TODO: Change this to enable deny(unused) if IPv6 or IPv4 are enabled
+#![cfg_attr(feature = "proto-ipv4", deny(unused))]
```
whitequark: What's holding up `#![deny(unused)]` for IPv6?

dlrobertson: Using it in `EthernetInterface`. Things like `Packet` and `cmp` are unused at the moment when only IPv6 is enabled

PR #100, opened 2017-12-23
Add faster TCP checksum implementation for x86_64

BurntPizza:
I was inspired/nerdsniped by your comment in the reddit thread on the TCP checksum being a bottleneck.
With this I'm getting a ~8-10% speed increase on the benchmark example.

It took some fiddling to get LLVM to vectorize satisfactorily, and it would still be possible to improve on the codegen using actual SIMD intrinsics. You can see the result [here](https://godbolt.org/g/9nZsd1). With `sse4.1` or `avx` target features the inner loop gets much tighter still.
One thing is that I only have a `x86_64` machine, so that's why it's the only arch enabled. I'm curious how it would do on other architectures that have vector instructions.

Please evaluate and benchmark (you can use my repo) and feel free to decline/request changes if it's not as good as I think it is.

### DIFF
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -613,10 +613,11 @@ pub mod checksum {
     pub fn data(mut data: &[u8]) -> u16 {
         let mut accum = 0;
 
-        // Operate on large chunks first to ease the job of the autovectorizer.
+        // For each 32-byte chunk...
         const CHUNK_SIZE: usize = 32;
         while data.len() >= CHUNK_SIZE {
             let mut d = &data[..CHUNK_SIZE];
+            // ... take by 2 bytes and sum them.
             while d.len() >= 2 {
                 accum += NetworkEndian::read_u16(d) as u32;
                 d = &d[2..];
@@ -625,11 +626,14 @@ pub mod checksum {
             data = &data[CHUNK_SIZE..];
         }
 
+        // Sum the rest that does not fit the last 32-byte chunk,
+        // taking by 2 bytes.
         while data.len() >= 2 {
             accum += NetworkEndian::read_u16(data) as u32;
             data = &data[2..];
         }
         
+        // Add the last remaining odd byte, if any.
         if let Some(&value) = data.first() {
             accum += (value as u32) << 8;
         }


kpp:
Why not use https://github.com/AdamNiederer/faster ? @AdamNiederer seems like a good challenge for you!

whitequark:
Sweet! Can you please add some testcases?

@kpp that doesn't work on `#![no_std]` unfortunately, which is smoltcp's primary deployment target.

AdamNiederer:
I don't think there's anything *preventing* it working with `no_std` - I'll see if I can hack something together.

BurntPizza:
I've revised to hopefully improve further upon  275eb90785a26ed1bd942f087c2d81647dc58f84, which was apparently the one simple formulation which works that I didn't try last night. :\

You'll want to bench this before accepting the PR, as it's pretty close.

What kind of testcases are you looking for?

whitequark:
> What kind of testcases are you looking for?

Not anymore, since the correctness is now pretty much obvious.

src/wire/ip.rs
```diff
@@ -611,15 +611,29 @@ pub mod checksum {
 
     /// Compute an RFC 1071 compliant checksum (without the final complement).
     pub fn data(mut data: &[u8]) -> u16 {
-        // See RFC 1071 section 4.1 for the original implementation.
-        let mut accum: u32 = 0;
+        let mut accum = 0;
+
+        // Operate on large chunks first to ease the job of the autovectorizer.
```
kpp: I would like to see comments like:

```
each 32-byte chunk
  take by 2 bytes as network order u16
  sum them
sum the rest that does not fit the last 32-byte chunk
```

whitequark: Better?

0x7CFE:
> that doesn't work on #![no_std] unfortunately, which is smoltcp's primary deployment target.

@whitequark, now it does! https://adamniederer.com/blog/faster-pr-2.html

AdamNiederer:
I'm seeing a 2x speedup on AVX with `faster`, and a 20-40% speedup on SSE using an unreleased verison of `faster`. I'll send a PR once 0.5.0 is out.

PR #102, opened 2017-12-24
Implement FromStr for IpEndpoint

### LINKED ISSUE #81: IpEndpoint should have a FromStr implementation
IpEndpoint does not currently have a `FromStr` implementation or `Parser::accept_ip_endpoint` implementation.

Per the suggestion by @batonius.
 
> ... it's trivial to split an IPv4 endpoint at ':', IPv6 endpoints are quite tricky: https://tools.ietf.org/html/rfc5952#section-6 . I think the recommended style would be enough, it's the only one std::net::SocketAddrV6 supports: https://doc.rust-lang.org/nightly/src/std/net/parser.rs.html#280

Original comment: https://github.com/m-labs/smoltcp/pull/72#issuecomment-344077253

hjr3:
Fixes #81

### DIFF
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -611,15 +611,33 @@ pub mod checksum {
 
     /// Compute an RFC 1071 compliant checksum (without the final complement).
     pub fn data(mut data: &[u8]) -> u16 {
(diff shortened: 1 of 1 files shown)


whitequark:
Thanks for the great PR!

PR #107, opened 2017-12-26
Ensure ICMPv4 error replies comply with size reqs

dlrobertson:
### Summary

While doing some reading over the weekend I realized I made a mistake
in #67. The size of the ICMPv4 replies generated by `icmpv4_reply` could
be too large.

### Changes

 - Add `MIN_MTU` constants to the IP version modules.
 - Ensure that ICMPv4 error replies comply with the size requirements
   specified in [RFC 1812 § 4.3.2.3].

[RFC 1812 § 4.3.2.3]: https://tools.ietf.org/html/rfc1812#section-4.3.2.3

### DIFF
--- a/examples/tcpdump.rs
+++ b/examples/tcpdump.rs
@@ -13,7 +13,7 @@ fn main() {
         phy_wait(socket.as_raw_fd(), None).unwrap();
         let (rx_token, _) = socket.receive().unwrap();
         rx_token.consume(/*timestamp = */ 0, |buffer| {
-            print!("{}", PrettyPrinter::<EthernetFrame<&[u8]>>::new("", &buffer));
+            println!("{}", PrettyPrinter::<EthernetFrame<&[u8]>>::new("", &buffer));
             Ok(())
         }).unwrap();
     }
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -382,52 +382,44 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
             let mut neighbor_addr = None;
             let mut device_result = Ok(());
             let &mut Self { ref mut device, ref mut inner } = self;
+
+            macro_rules! respond {
+                ($response:expr) => ({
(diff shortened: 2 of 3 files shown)


src/wire/ipv4.rs
```diff
@@ -7,6 +7,11 @@ use super::ip::{checksum, pretty_print_ip_payload};
 
 pub use super::IpProtocol as Protocol;
 
+/// Minimum MTU required of all links supporting IPv4. See [RFC 791 § 3.1]
+///
+/// [RFC 791 § 3.1]: https://tools.ietf.org/html/rfc791#section-3.1
+pub const MIN_MTU: usize = 576;
```
whitequark: No, this is actually not enough, 576 is the minimum datagram size. If you have e.g. an Ethernet header in front, then MTU of 576 is not sufficient.

dlrobertson: Right... my bad

dlrobertson: It is the minimum MTU for IPv6 https://tools.ietf.org/html/rfc8200#section-5. I think that is where I got confused. I'll have to think about how to best word this, since it is the minimum datagram size for IPv4 and the minimum MTU for IPv6. Maybe we end up with `IPV4_MIN_DATAGRAM_SIZE` and `IPV6_MIN_MTU`?

whitequark: Still not quite, it's very confusing. Take a look at https://en.wikipedia.org/wiki/Maximum_transmission_unit#MTUs_for_common_media. The minimum MTU on IPv4 is actually 68, but since we never fragment ourselves, the *effective* minimum MTU supported by smoltcp on IPv4 is 576. This needs to be documented, and the constant can remain the same.

whitequark:
@dlrobertson Oops, I've missed the changes in this PR again. It looks like GitHub doesn't send email notifications on rebases... care to leave a comment in place of that?

dlrobertson:
Yeah, I'll try to leave comments in the future.

Issue #35, opened 2017-08-29
Use a newtype instead of u64 for timestamps


whitequark:
It's getting embarassing.

whitequark:
@dlrobertson Could you take a look at this, please? It's been getting annoying.

dlrobertson:
Sounds good

PR #110, opened 2017-12-28
Add Timestamp type

dlrobertson:
We use timstamps often as function arguments. We should create a
Timestamp type that is used in these cases.

Fixes: #35

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -12,9 +12,6 @@ absolute and relative time.
 
 use core::{ops, fmt};
 
-#[cfg(feature = "std")]
-const NANOS_PER_MILLI: u32 = 1_000_000;
-
 /// A representation of an absolute time value.
 ///
 /// The `Instant` type is a wrapper around a `i64` value that
@@ -45,7 +42,7 @@ impl Instant {
     #[cfg(feature = "std")]
     pub fn from_system_time(time: ::std::time::SystemTime) -> Result<Instant, ::std::time::SystemTimeError> {
         let n = ::std::time::UNIX_EPOCH.duration_since(time)?;
-        Ok(Self::from_millis(n.as_secs() as i64 * 1000 + (n.subsec_nanos() / NANOS_PER_MILLI) as i64))
+        Ok(Self::from_millis(n.as_secs() as i64 * 1000 + (n.subsec_nanos() / 1_000_000) as i64))
     }
 
     /// The fractional number of milliseconds that have passed
@@ -192,7 +189,7 @@ impl ops::DivAssign<u32> for Duration {
 impl From<::std::time::Duration> for Duration {
     fn from(other: ::std::time::Duration) -> Duration {
         Duration::from_millis(
-            other.as_secs() * 1000 + (other.subsec_nanos() / NANOS_PER_MILLI) as u64
+            other.as_secs() * 1000 + (other.subsec_nanos() / 1_000_000) as u64
         )
     }
 }


src/timestamp.rs
```diff
@@ -0,0 +1,138 @@
...
+/// The number of milliseconds, monotonically increasing since an
+/// arbitrary moment in time, such as system startup.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Timestamp(pub u64);
+
+impl Timestamp {
+    /// Add a timestamp value or integer to the current timestamp
+    /// value. If the value is an integer the integer is assumed
+    /// to be a number of milliseconds.
+    pub fn advance<T: Into<Timestamp>>(&self, other: T) -> Timestamp {
```
whitequark: Ummmm, no. The *whole point* of having newtypes here is so that *timestamps* (absolute) and *intervals* (relative) cannot be accidentally mixed (which happened several times).

dlrobertson:
I wanted to get feedback on the actual `Timestamp` structure before substituting all of the `timestamp: u64` locations with it. I wasn't entirely sure what all of the requirements were for this type. At the moment the assumption is that it is a little more than a fancy integer.

src/timestamp.rs
whitequark: Look at how libstd implements this. We need more or less the same thing, but simpler (it only has to handle milliseconds, and it doesn't have to hide its internals quite as ardently).

dlrobertson: > Ummmm, no. The whole point of having newtypes here is so that timestamps (absolute) and intervals (relative)

So do we want both a relative and absolute time type?

dlrobertson: like libstd's `Duration` and `Timespec`

whitequark: Exactly.

dlrobertson:
Still need to fill out the `TimeInterval` type, but is this more like what we're looking for?

src/time.rs
```diff
@@ -0,0 +1,182 @@
...
+
+use std::{ops, fmt};
+
+static NANOS_PER_SEC: u32 = 1_000_000_000;
+
+/// A representation of an absolute time value.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TimeStamp {
+    secs: u64,
+    nanos: u32,
```
jordens: Are the ~600 years range of a plain `u64` nanoseconds not enough?

dlrobertson: My goal here was to implement something similar to `Timespec` and `Duration` from libstd. This is what `Timespec` and `Duration` do, so it is what I used. We certainly could use `TimeStamp(u64)`, but it wasn't that much work to add a seconds and nanoseconds member and most other implementations I found use something like this instead of `TimeStamp(u64)`.

whitequark: Please let's not hassle with this. `pub struct Timestamp { millis: u64 }` is *more than enough*, we are using it right now for a reason. Nanoseconds are completely unnecessary in a TCP/IP stack and calculations with them are a waste of time, and the additional API complexity is a nightmare.

dlrobertson:
@jordens and @whitequark updated to address your concerns

src/time.rs
```diff
@@ -0,0 +1,136 @@
...
+        TimeStamp { millis }
+    }
+
+    /// Add a [TimeInterval] to the current timestamp value and return a
+    /// `Option` containing a new [TimeStamp] of the result on success.
+    ///
+    /// [TimeInterval]: struct.TimeInterval.html
+    /// [TimeStamp]: struct.TimeStamp.html
+    pub fn advance(&self, rhs: TimeInterval) -> Option<TimeStamp> {
+        self.millis.checked_add(rhs.millis()).map(|x| TimeStamp::new(x))
```
whitequark:
```
> 2**64 milliseconds -> million years
approx. 584.5545 million tropicalyear (time)
```
Really? Caring about overflows here?

src/time.rs
```diff
@@ -0,0 +1,136 @@
...
+//! The `time` modules contains structures used to represent both
+//! absolute and relative time.
+
+use core::{ops, fmt};
+
+static MILLIS_PER_SEC: u64 = 1_000;
+
+/// A representation of an absolute time value.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TimeStamp {
```
whitequark: Why not use `Instant` and `Duration`, like libstd does?

src/time.rs
```diff
@@ -0,0 +1,136 @@
...
+        self.millis.checked_add(rhs.millis()).map(|x| TimeStamp::new(x))
+    }
+
+    /// Subtract a [TimeInterval] from the current timestamp value and
+    /// return a `Option` containing a new [TimeStamp] of the result on
+    /// success.
+    ///
+    /// [TimeInterval]: struct.TimeInterval.html
+    /// [TimeStamp]: struct.TimeStamp.html
+    pub fn sub(&self, rhs: TimeInterval) -> Option<TimeStamp> {
```
whitequark: Also, if we go for inherent method `advance`, then I think we should ditch inherent `sub` and go for just `difference` (renamed `saturating_sub`). It's enough to have `-` defined as `checked_sub`, I think...

src/time.rs
dlrobertson: Updated

src/time.rs
dlrobertson: Renamed `advance` to `checked_add` so anyone that needs to worry about overflows can (no idea who will).

src/time.rs
dlrobertson: kept `saturating_sub` and renamed `advance` for `checked_add` and `sub` for `checked_sub`.

src/time.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// The number of milliseconds that have passed since an arbitrary
+    /// point in time.
+    pub fn millis(&self) -> u64 {
+        self.millis
+    }
+
+    /// The number of seconds that have passed since an arbitrary
+    /// point in time.
+    pub fn secs(&self) -> u64 {
+        self.millis % MILLIS_PER_SEC
```
crawford: Shouldn't this be division?

src/time.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+impl fmt::Display for Instant {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}s{}ms", self.secs(), self.millis())
```
crawford: Why not "2.547s"?

src/time.rs
```diff
@@ -0,0 +1,136 @@
...
+impl Duration {
+    /// Create a new `Duration`.
+    pub fn new(millis: u64) -> Duration {
+        Duration { millis }
+    }
+
+    /// The relative number of seconds that have passed since an
+    /// arbitrary point in time.
+    pub fn secs(&self) -> u64 {
+        self.millis
```
crawford: Divide by 1000?

src/time.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /// The relative number of milliseconds that have passed since an
+    /// arbitrary point in time.
+    pub fn millis(&self) -> u64 {
+        self.millis % MILLIS_PER_SEC
```
crawford: It seems odd that Instant and Duration behave differently with regard to millis() and secs(). I think this should just return millis or Instant should be changed to return the modulus.

src/time.rs
whitequark: :+1:

src/time.rs
dlrobertson: Yup, same with the others. Looks like I forgot to update the getters when I moved from separate second and millisecond members to on millisecond member.

dlrobertson:
After the update `secs` will return `self.millis / MILLIS_PER_SEC`, `millis` will return `self.millis % MILLIS_PER_SEC`, and `elapsed` will return `self.millis`.

src/time.rs
```diff
@@ -0,0 +1,170 @@
+//! Time structures.
+//!
+//! The `time` modules contains structures used to represent both
+//! absolute and relative time.
+
+use core::{ops, fmt};
+
+static MILLIS_PER_SEC: u64 = 1_000;
```
whitequark: This would put the constant in `.rodata`. Use `const`.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// A representation of an absolute time value.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
```
whitequark: `Default` is semantically inapplicable. There are no default instants! Deriving `Default` for something else with an `Instant` in it is a logic error; it should use an `Option<Instant>` instead.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /// Add a [Duration] to the current timestamp value and return a
+    /// `Option` containing a new [Instant] of the result on success.
+    ///
+    /// [Duration]: struct.Duration.html
+    /// [Instant]: struct.Instant.html
+    pub fn checked_add(&self, rhs: Duration) -> Option<Instant> {
+        self.millis.checked_add(rhs.millis()).map(|x| Instant::new(x))
+    }
```
whitequark: Like I mentioned, there is no condition in which `checked_add` will ever return `None`. This function is meaningless.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+//! absolute and relative time.
+
+use core::{ops, fmt};
+
+static MILLIS_PER_SEC: u64 = 1_000;
+
+/// A representation of an absolute time value.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Instant {
+    millis: u64,
```
whitequark: On second thought, I think this should be an `i64`. Then we do not need `checked_sub` and `saturating_sub` at all! The zero instant would just be an arbitrary point in a (nearly) infinite timeline.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+        self.millis.checked_add(rhs.millis()).map(|x| Instant::new(x))
+    }
+
+    /// Subtract a [Duration] from the current timestamp value and
+    /// return a `Option` containing a new [Instant] of the result on
+    /// success.
+    ///
+    /// [Duration]: struct.Duration.html
+    /// [Instant]: struct.Instant.html
+    pub fn checked_sub(&self, rhs: Duration) -> Option<Instant> {
```
whitequark: Not needed per above.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+        self.millis.checked_sub(rhs.millis()).map(|x| Instant::new(x))
+    }
+
+    /// Perform a saturating subtraction with a [Duration] from
+    /// the current timestamp value and return a new [Instant] of
+    /// the result.
+    ///
+    /// [Duration]: struct.Duration.html
+    /// [Instant]: struct.Instant.html
+    pub fn saturating_sub(&self, interval: Duration) -> Instant {
```
whitequark: I am struggling to think of when this operation would have been semantically valid. It seems like a bug factory. Nevertheless, it is not needed as per above.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+        Instant::new(self.millis.saturating_sub(interval.millis()))
+    }
+
+
+    /// The number of milliseconds that have passed since an arbitrary
+    /// point in time.
```
whitequark: "an arbitrary point in time" → "the beginning of time", and the docstring for `Instant` itself should explain that the zero value of `Instant` is inherently arbitrary.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+impl Duration {
+    /// Create a new `Duration`.
+    pub fn new(millis: u64) -> Duration {
+        Duration { millis }
+    }
+
+    /// The relative number of milliseconds that have passed since an
+    /// arbitrary point in time.
```
whitequark: The docstrings are all wrong.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// The relative number of seconds that have passed since an
+    /// arbitrary point in time.
+    pub fn secs(&self) -> u64 {
+        self.millis / MILLIS_PER_SEC
+    }
+
+    /// The total number of milliseconds that have passed since an
+    /// arbitrary point in time.
+    pub fn elapsed(&self) -> u64 {
```
whitequark: `Duration::elapsed` doesn't make sense. `total_millis`?

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// A relative amount of time.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Duration {
+    millis: u64,
+}
+
+impl Duration {
+    /// Create a new `Duration`.
+    pub fn new(millis: u64) -> Duration {
```
whitequark: `from_millis`.

src/time.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}.{}s", self.secs(), self.millis())
+    }
+}
+
+impl ops::Add<Duration> for Instant {
+    type Output = Instant;
+
+    fn add(self, rhs: Duration) -> Instant {
+        Instant::new(self.millis + rhs.millis())
```
whitequark: `rhs.total_millis()` (neé `rhs.elapsed()`), not `rhs.millis()`. And this needs a test.

src/time.rs
dlrobertson: :+1: def more readable

src/time.rs
dlrobertson: :+1:

src/time.rs
dlrobertson: What would a negative value indicate?

whitequark: "Before startup", since 0 is usually "startup".

src/time.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    fn sub(self, rhs: Duration) -> Duration {
+        Duration::from_millis(self.millis - rhs.total_millis())
+    }
+}
+
+impl ops::Mul<Duration> for Duration {
+    type Output = Duration;
+
+    fn mul(self, rhs: Duration) -> Duration {
```
whitequark: This doesn't make sense... Consider the units. You're multiplying `self` (seconds) by `rhs` (also seconds), so you get seconds squared. But you return seconds!

Thus, `rhs` should be `u32`, just like in `std::time`.

src/time.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    fn sub(self, rhs: Duration) -> Instant {
+        Instant::from_millis(self.millis - rhs.total_millis())
+    }
+}
+
+/// A relative amount of time.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Duration {
+    millis: i64,
```
whitequark: No, _this_ should be an unsigned value. Consider a function `fn sleep(Duration)`; how can it possibly sleep for, say, minus one minute?

src/time.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn add(self, rhs: Duration) -> Duration {
+        Duration::from_millis(self.millis + rhs.total_millis())
+    }
+}
+
+impl ops::Sub<Duration> for Duration {
+    type Output = Duration;
+
+    fn sub(self, rhs: Duration) -> Duration {
+        Duration::from_millis(self.millis - rhs.total_millis())
```
whitequark: Needs to use `checked_sub` after `self.millis` becomes unsigned.

src/time.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    fn mul(self, rhs: Duration) -> Duration {
+        Duration::from_millis(self.millis * rhs.total_millis())
+    }
+}
+
+impl ops::Div<Duration> for Duration {
+    type Output = Duration;
+
+    fn div(self, rhs: Duration) -> Duration {
```
whitequark: You're making the same mistake here as with `Mul`, but it's more subtle. In principle, `Duration / Duration` is a meaningful operation (unlike `Duration * Duration`, which returns a physically nonsensical quantity, this returns the ratio of the two durations), but it's not *especially* useful. Instead, this should be an `impl ops::Div<u32> for Duration`.

src/time.rs
```diff
@@ -0,0 +1,189 @@
...
+        // std::ops::Sub
+        assert_eq!(Duration::from_millis(555) - Duration::from_millis(42), Duration::from_millis(513));
+        // std::ops::Mul
+        assert_eq!(Duration::from_millis(13) * Duration::from_millis(22), Duration::from_millis(286));
+        // std::ops::Div
+        assert_eq!(Duration::from_millis(53) / Duration::from_millis(4), Duration::from_millis(13));
+    }
+
+    #[test]
+    #[should_panic(expected = "attempt to multiply with overflow")]
```
whitequark: This test and the other overflow test above would fail on `cargo test --release` since that doesn't include overflow checks and you don't explicitly use checked operations. I suggest simply removing them, and letting the default Rust arithmetic rules apply to time manipulation, since 64-bit signed overflows and underflows are really not important in this context.

There is of course the important exception of `Duration - Duration`, which is an unsigned underflow, and should be tested.

src/time.rs
dlrobertson: Why would the units be seconds here when they're milliseconds everywhere else?

src/time.rs
dlrobertson: Same as above.

src/time.rs
dlrobertson: Right. I'll update this.

src/time.rs
whitequark: I'm talking in terms of [dimensional analysis](https://en.wikipedia.org/wiki/Dimensional_analysis) here. Seconds and milliseconds are the same in the context.

dlrobertson:
Updated. I used `checked_<op>` with an expect string in the operations because that is what is done in libstd.

I also don't see a way for us to implement `From` and `Into` for `SystemTime` and `Instant`. I don't see any sort of constructor for `SystemTime` other than `SystemTime::now`. There is `elapsed` which returns a `Result`, so I could implement `from_system_time` for `Instant`, but I don't see a way to impl `Into<::std::time::SystemTime>` for `Instant`.

whitequark:
> I don't see any sort of constructor for SystemTime other than SystemTime::now.

There is `std::time::UNIX_EPOCH` so you can do `UNIX_EPOCH + some_duration`.

src/time.rs
```diff
@@ -0,0 +1,281 @@
...
+ - [Instant] is used to represent absolute time.
+ - [Duration] is used to represet relative time.
+
+[Instant]: struct.Instant.html
+[Duration]: struct.Duration.html
+*/
+
+use core::{ops, fmt};
+
+const SIGNED_MILLIS_PER_SEC: i64 = 1_000;
```
whitequark: Can we just use `1000`? These constants are a net decrease in readability.

src/time.rs
```diff
@@ -0,0 +1,281 @@
...
+        write!(f, "{}.{}s", self.secs(), self.millis())
+    }
+}
+
+impl ops::Add<Duration> for Duration {
+    type Output = Duration;
+
+    fn add(self, rhs: Duration) -> Duration {
+        Duration::from_millis(
+            self.millis.checked_add(rhs.total_millis()).expect("overflow when adding durations"))
```
whitequark: I maintain my position that checking for unsigned overflow is a waste of time.

src/time.rs
```diff
@@ -0,0 +1,281 @@
...
+    fn mul_assign(&mut self, rhs: u32) {
+        self.millis = self.millis.checked_mul(rhs as u64).expect("overflow when multiplying by scalar");
+    }
+}
+
+impl ops::Div<u32> for Duration {
+    type Output = Duration;
+
+    fn div(self, rhs: u32) -> Duration {
+        Duration::from_millis(self.millis / rhs as u64)
```
whitequark: Conversely, what libstd does and what you forgot to do here is to check for division by zero.

src/time.rs
dlrobertson: Np, I'll revert it. Just wondered if that was enough to used the `checked` variant.

dlrobertson:
Based `SystemTime` conversions off of `UNIX_EPOCH`. I also added `Instant::now`, removed the use of `checked` functions, and removed the `MILLIS_PER_SEC` constants.

src/time.rs
```diff
@@ -0,0 +1,302 @@
...
+ - [Duration] is used to represet relative time.
+
+[Instant]: struct.Instant.html
+[Duration]: struct.Duration.html
+*/
+
+use core::{ops, fmt};
+
+#[cfg(feature = "std")]
+const NANOS_PER_MILLI: u32 = 1_000_000;
```
whitequark: One last nit: let's just embed this one too. Meanwhile let me fix the build failure.

dlrobertson:
Ugh, I accidentally dismissed a review. If you see a comment of yours missing, feel free to add it again...

src/time.rs
whitequark: I've released managed 0.5.1 which should fix the build failure.

dlrobertson: Is that what is happening in #116 as well?

whitequark: I think so

whitequark:
@dlrobertson Prefer a followup PR.

PR #113, opened 2018-01-04
Fix documentation warnings

dlrobertson:
There are several warnings that are thrown when running `cargo doc`. Fix
these warnings.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -9,16 +9,16 @@ use phy::{Device, DeviceCapabilities, RxToken, TxToken};
 use wire::pretty_print::PrettyPrinter;
 use wire::{EthernetAddress, EthernetProtocol, EthernetFrame};
 #[cfg(feature = "proto-ipv4")]
-use wire::{Ipv4Address};
+use wire::{Ipv4Address, Ipv4Packet, Ipv4Repr};
 use wire::{IpAddress, IpProtocol, IpRepr, IpCidr};
 #[cfg(feature = "proto-ipv4")]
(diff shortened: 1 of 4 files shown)


dlrobertson:
I'm okay with dropping the code block changes due to [rust-lang/rust#38739](https://github.com/rust-lang/rust/issues/38739)

src/lib.rs
```diff
@@ -10,16 +10,14 @@
 //! for writing applications but also to be a toolbox of networking primitives, so
 //! every layer is fully exposed and documented.
 //!
-//! When discussing networking stacks and layering, often the [OSI model][osi] is invoked.
+//! When discussing networking stacks and layering, often the [OSI model](osi) is invoked.
```
whitequark: Wait, what? How is this legal Markdown?

src/phy/mod.rs
```diff
@@ -17,73 +17,71 @@
 //! An implementation of the [Device](trait.Device.html) trait for a simple hardware
 //! Ethernet controller could look as follows:
 //!
-/*!
-```rust
```
whitequark: This change makes it a real pain in the ass to update large code blocks.

src/lib.rs
dlrobertson: Its not :(. I was playing around with some stuff and it got lumped in here.

src/phy/mod.rs
dlrobertson: Agreed. If we're okay with some extra warnings, I'd be okay with going back to `/*!`.

dlrobertson:
Updated to use `/*!` for the entire section.

PR #114, opened 2018-01-08
Clean up dispatch_socket!() syntax

crawford:
This modified syntax is meant to more closely resemble standard Rust.


whitequark:
Thanks!

PR #116, opened 2018-01-11
Handle IPv4 mapped IPv6 addresses

### LINKED ISSUE #86: Implement parsers and `Display` for IPv4 mapped IPv6 addresses
[`Display` for `IPv6Address`](https://github.com/m-labs/smoltcp/blob/master/src/wire/ipv6.rs#L131-L173) and the [IPv6 address parser](https://github.com/m-labs/smoltcp/blob/master/src/parsers.rs#L140-L206) should handle [IPv4-mapped IPv6 addresses](https://tools.ietf.org/html/rfc4291#section-2.5.5.2). which take on the form `::ffff:x.x.x.x`.

In addition `From<IPv4Address>` should be implemented for `IPv6Address`, converting the given IPv4 address into a IPv4-mapped IPv6 address if both of the features `proto-ipv4` and `proto-ipv6` are enabled.

Note: [IPv6 Compatable addresses](https://tools.ietf.org/html/rfc4291#section-2.5.5.1) have been deprecated. AFAIK there is no need to implement them.

hjr3:
* The IPv6 address parser now handles IPv4 mapped IPv6 addresses which
  take on the form ::ffff:x.x.x.x.
* Implement Display for IPv4 mapped IPv6 addresses
* Implement From<IPv4Address> for IPv6Address

Fixes #86

### DIFF
--- a/src/socket/raw.rs
+++ b/src/socket/raw.rs
@@ -6,6 +6,8 @@ use phy::ChecksumCapabilities;
 use wire::{IpVersion, IpRepr, IpProtocol};
 #[cfg(feature = "proto-ipv4")]
 use wire::{Ipv4Repr, Ipv4Packet};
+#[cfg(feature = "proto-ipv6")]
+use wire::{Ipv6Repr, Ipv6Packet};
 use socket::{Socket, SocketMeta, SocketHandle};
 use storage::{Resettable, RingBuffer};
 
@@ -205,7 +207,13 @@ impl<'a, 'b> RawSocket<'a, 'b> {
                     Ok((IpRepr::Ipv4(ipv4_repr), packet.payload()))
                 }
                 #[cfg(feature = "proto-ipv6")]
-                IpVersion::Ipv6 => Err(Error::Unrecognized),
+                IpVersion::Ipv6 => {
+                    let mut packet = Ipv6Packet::new_checked(buffer.as_mut())?;
+                    if packet.next_header() != protocol { return Err(Error::Unaddressable) }
+                    let packet = Ipv6Packet::new(&*packet.into_inner());
+                    let ipv6_repr = Ipv6Repr::parse(&packet)?;
+                    Ok((IpRepr::Ipv6(ipv6_repr), packet.payload()))
+                }
                 IpVersion::Unspecified => unreachable!(),
                 IpVersion::__Nonexhaustive => unreachable!()
             }
@@ -244,14 +252,17 @@ impl<'a, 'b> RawSocket<'a, 'b> {
 
 #[cfg(test)]
 mod test {
+    use wire::IpRepr;
     #[cfg(feature = "proto-ipv4")]
-    use wire::{Ipv4Address, IpRepr, Ipv4Repr};
+    use wire::{Ipv4Address, Ipv4Repr};
+    #[cfg(feature = "proto-ipv6")]
+    use wire::{Ipv6Address, Ipv6Repr};
     use super::*;
 
     fn buffer(packets: usize) -> SocketBuffer<'static, 'static> {
         let mut storage = vec![];
         for _ in 0..packets {
-            storage.push(PacketBuffer::new(vec![0; 24]))
+            storage.push(PacketBuffer::new(vec![0; 48]))
         }
         SocketBuffer::new(storage)
     }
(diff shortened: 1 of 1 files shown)


src/parsers.rs
```diff
@@ -164,6 +164,20 @@ impl<'a> Parser<'a> {
             }
         };
 
+        // Check for an IPv4 mapped address
+        if *head_idx == 6 || use_tail && *tail_idx == 1 {
```
dlrobertson: Shouldn't this also check that the current `u16` is `0xffff`?

src/parsers.rs
```diff
@@ -164,6 +164,20 @@ impl<'a> Parser<'a> {
...
+                    *tail_idx += 1;
+                    tail[*tail_idx] = ((octets[2] as u16) << 8) | (octets[3] as u16);
+                    *tail_idx += 1;
+                    return Ok(());
+                },
+                None => ()
+            }
+        }
+
         match self.try(|p| p.accept_number(4, 0x10000, true)) {
```
dlrobertson: IMO at a first glance, I'd say the parsing of the ipv4 address should be a match arm here if the number is is `0xffff` and `*head_idx == 6 || use_tail && *tail_idx == 1`. This would also increase readability in my opinion.

src/parsers.rs
```diff
@@ -510,6 +529,11 @@ mod test {
         // Long number
         assert_eq!(Ipv6Address::from_str("::000001"),
                    Err(()));
+        // IPv4-Mapped address
+        assert_eq!(Ipv6Address::from_str("::ffff:192.168.1.1"),
+                   Ok(Ipv6Address([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, 192, 168, 1, 1])));
+        assert_eq!(Ipv6Address::from_str("0:0:0:0:0:FFFF:192.168.1.1"),
+                   Ok(Ipv6Address([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, 192, 168, 1, 1])));
```
dlrobertson: Could you add some cases that fail? E.g. `::eeee:192.168.1.1`?

dlrobertson: I *think* `::1:fff:192.168.1.1` would also be accepted as a valid IPv4 mapped IPv6 address under the current parsing rules.

src/parsers.rs
hjr3: i thought about doing this, but it will cause a lot of change in the parser. before trying to parse a ipv4 address, i would have to `self.accept_char(b':')?;`. should the parser fail, the check down below looks ahead of `b':'` and errors if one is not present. i would also have to remove the `self.accept_char(b':')?;` in the `None` case for the double colon check.

all that being said, i see your point about readability. i will see what i can make better.

src/parsers.rs
hjr3: done

src/parsers.rs
hjr3: yes, fixed.

src/parsers.rs
```diff
@@ -169,12 +181,26 @@ impl<'a> Parser<'a> {
...
+                    });
+                }
                 Ok(())
             },
             Some(part) if *tail_idx < 6 => {
                 // Valid u16 to be added to the address
                 tail[*tail_idx] = part as u16;
                 *tail_idx += 1;
+
+                if part == 0xffff && use_tail && *tail_idx == 1 {
```
dlrobertson: AFAIK `use_tail` should be always true in this branch.

src/parsers.rs
hjr3: ignore my last comment. wrapping it in `self.try()` works fine.

i did not make it a separate match arm. i wrote that at first but that caused me to duplicate the logic of the next 3 match arms because i need to do something with `part` still. instead, i incorporated it into the match arms for head and tail. i think it reads better now. let me know what you think.

src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![deny(unsafe_code)]
 // TODO: Change this to enable deny(unused) if IPv6 or IPv4 are enabled
 #![cfg_attr(feature = "proto-ipv4", deny(unused))]
+#![feature(slice_patterns)]
```
dlrobertson: `slice_patterns` isn't available on stable, so if we're going to use slice patterns an additional note should probably be added to https://github.com/m-labs/smoltcp/blob/master/.travis.yml#L4-L9

src/wire/ipv6.rs
```diff
@@ -156,54 +156,72 @@ impl Address {
...
         }
         Ok(())
     }
 }
 
+#[cfg(feature = "proto-ipv4")]
+/// Convert the given IPv4 address into a IPv4-mapped IPv6 address
+impl From<::wire::ipv4::Address> for Address {
+    fn from(address: ::wire::ipv4::Address) -> Self {
+        Address::new(0, 0, 0, 0, 0, 0xffff,
```
dlrobertson: Might be more readable to just use `Address([values])` since you have to shift some bits. Feel free to ignore this comment. Just a nit.

src/parsers.rs
dlrobertson: Nice work. The update is much more readable

src/parsers.rs
dlrobertson: Actually my example was bad. `0:0:0:0:1:FFFF:192.168.1.1` is still parsed as a correctly formatted IPv4 mapped IPv6 address, but it shouldn't be. I think we need to check to ensure that the value of the head part contains only zero.

src/parsers.rs
hjr3: yes, removed

src/lib.rs
hjr3: added

src/parsers.rs
hjr3: Ah, yes. i updated the code to make sure that positions 0-4 of the head are 0. I also added some more test cases that fail.

hjr3:
One job in the build matrix is failing do to a dependency. I am not sure why. I will look into this later.

src/parsers.rs
```diff
@@ -182,7 +182,8 @@ impl<'a> Parser<'a> {
                 head[*head_idx] = part as u16;
                 *head_idx += 1;
 
-                if part == 0xffff && *head_idx == 6 {
+                if *head_idx == 6 && head[5] == 0xffff && head[4] == 0 && head[3] == 0
```
dlrobertson: nit: I like `head_idx == 6 && head[:head_idx] == [0, 0, 0, 0, 0, 0xffff]`

src/parsers.rs
```diff
@@ -195,7 +196,7 @@ impl<'a> Parser<'a> {
                 tail[*tail_idx] = part as u16;
                 *tail_idx += 1;
 
-                if part == 0xffff && use_tail && *tail_idx == 1 {
+                if *tail_idx == 1 && tail[0] == 0xffff {
```
dlrobertson: Also need to make sure that `head_idx` is 0, otherwise `1::ffff:<ip>` is valid.

src/parsers.rs
```diff
@@ -169,12 +181,27 @@ impl<'a> Parser<'a> {
                 // Valid u16 to be added to the address
                 head[*head_idx] = part as u16;
                 *head_idx += 1;
+
+                if *head_idx == 6 && head[5] == 0xffff && head[4] == 0 && head[3] == 0
```
whitequark: `head[0..head_idx] == [0, 0, 0, 0, 0, 0xffff]` ?

src/parsers.rs
hjr3: i was not aware of that syntax. thanks

src/parsers.rs
hjr3: per my reading of the spec, `0::ffff:<ip>` is valid though. i ended up checking that all of `head` is 0 in the tail match arm. tests added too.

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: rust
 matrix:
   include:
     ### Litmus check that we work on stable/beta
-    # (we don't, not until slice_rotate lands)
+    # (we don't, not until slice_rotate and slice_patterns land)
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6.rs
```diff
@@ -156,54 +156,72 @@ impl Address {
...
-                    State::TailBody
-                },
-                // Write the u16 with a leading colon when parsing a value
-                // that isn't the first in a section
-                (_, &State::HeadBody) | (_, &State::TailBody) => {
-                    write!(f, ":{:x}", word)?;
-                    state
+        match self.0 {
+            // IPv4-Mapped address
+            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, a, b, c, d] => {
```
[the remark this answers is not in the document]
hjr3: fixed

src/parsers.rs
hjr3: i was not aware of that syntax. thanks

hjr3:
> I would like to have Ipv6Address::is_ipv4_mapped and Ipv6Address::as_ipv4, too.

Added. This made all the changes more readable too!

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: rust
 matrix:
   include:
     ### Litmus check that we work on stable/beta
-    # (we don't, not until slice_rotate lands)
+    # (we don't, not until slice_rotate and lands)
```
dlrobertson: Unused and

hjr3: 😒 fixed

src/parsers.rs
dlrobertson: I appear to have incorrectly incorporated python syntax in my example :)

src/parsers.rs
dlrobertson: :+1:

hjr3:
@whitequark let me know if you need me to rebase/squash this into one commit

whitequark:
GitHub has a button for this n ow.

PR #117, opened 2018-01-12
Return `UdpSocket` from `UdpSocket::new` (instead of `Socket`)

phil-opp:
This makes it possible to call methods such as `bind` without downcasting.

### DIFF
--- a/src/socket/set.rs
+++ b/src/socket/set.rs
@@ -47,7 +47,9 @@ impl<'a, 'b: 'a, 'c: 'a + 'b> Set<'a, 'b, 'c> {
     ///
     /// # Panics
     /// This function panics if the storage is fixed-size (not a `Vec`) and is full.
-    pub fn add(&mut self, socket: Socket<'b, 'c>) -> Handle {
+    pub fn add<T>(&mut self, socket: T) -> Handle
+        where T: Into<Socket<'b, 'c>>
+    {
         fn put<'b, 'c>(index: usize, slot: &mut Option<Item<'b, 'c>>,
                        mut socket: Socket<'b, 'c>) -> Handle {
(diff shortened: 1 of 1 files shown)


whitequark:
1. You broke Travis.
2. You should update all of the other sockets too.
3. You should add an `Into` implementation or something for easy conversion to `Socket.

whitequark:
@phil-opp Can you update this PR? I'm OK with the concept, just needs the rest of the implementation.

whitequark:
And please add `impl Into<Socket> for UdpSocket` etc.

phil-opp:
Done. Sorry for the delay!

examples/loopback.rs
```diff
@@ -117,8 +117,8 @@ fn main() {
 
     let mut socket_set_entries: [_; 2] = Default::default();
     let mut socket_set = SocketSet::new(&mut socket_set_entries[..]);
-    let server_handle = socket_set.add(server_socket);
-    let client_handle = socket_set.add(client_socket);
+    let server_handle = socket_set.add(server_socket.into());
```
whitequark: To avoid this, `SocketSet::add` should have the signature `fn add<T: Into<Socket>>(socket: T)`.

phil-opp: Yes, I thought about this, but I wasn't sure if you're okay with the resulting increase in file size. But I assume that it's negligible.

whitequark: It might not be, but space profiling is something that should be done separately and preferably without degrading API ergonomics.

phil-opp:
I made SocketSet::add generic to avoid the `into` calls.

phil-opp:
Agreed!

PR #118, opened 2018-01-13
Support IPv6 raw sockets

dlrobertson:
- Add support for IPv6 raw sockets.
 - Update tests.

### DIFF
--- a/src/socket/raw.rs
+++ b/src/socket/raw.rs
@@ -6,6 +6,8 @@ use phy::ChecksumCapabilities;
 use wire::{IpVersion, IpRepr, IpProtocol};
 #[cfg(feature = "proto-ipv4")]
 use wire::{Ipv4Repr, Ipv4Packet};
+#[cfg(feature = "proto-ipv6")]
+use wire::{Ipv6Repr, Ipv6Packet};
 use socket::{Socket, SocketMeta, SocketHandle};
 use storage::{Resettable, RingBuffer};
 
@@ -205,7 +207,13 @@ impl<'a, 'b> RawSocket<'a, 'b> {
                     Ok((IpRepr::Ipv4(ipv4_repr), packet.payload()))
                 }
                 #[cfg(feature = "proto-ipv6")]
-                IpVersion::Ipv6 => Err(Error::Unrecognized),
+                IpVersion::Ipv6 => {
+                    let mut packet = Ipv6Packet::new_checked(buffer.as_mut())?;
+                    if packet.next_header() != protocol { return Err(Error::Unaddressable) }
+                    let packet = Ipv6Packet::new(&*packet.into_inner());
+                    let ipv6_repr = Ipv6Repr::parse(&packet)?;
+                    Ok((IpRepr::Ipv6(ipv6_repr), packet.payload()))
+                }
                 IpVersion::Unspecified => unreachable!(),
                 IpVersion::__Nonexhaustive => unreachable!()
             }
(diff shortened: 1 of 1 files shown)


dlrobertson:
> The very first PR should probably just handle IPv6 packets and forward them to the socket. The second should probably handle sending packets via RawSocket.

Flipped the order of these.

src/socket/raw.rs
```diff
@@ -205,7 +207,13 @@ impl<'a, 'b> RawSocket<'a, 'b> {
                     Ok((IpRepr::Ipv4(ipv4_repr), packet.payload()))
                 }
                 #[cfg(feature = "proto-ipv6")]
-                IpVersion::Ipv6 => Err(Error::Unrecognized),
+                IpVersion::Ipv6 => {
+                    let mut packet = Ipv6Packet::new_checked(buffer.as_mut())?;
+                    if packet.next_header() != protocol { return Err(Error::Unaddressable) }
+                    let packet = Ipv6Packet::new(&*packet.into_inner());
+                    let ipv4_repr = Ipv6Repr::parse(&packet)?;
```
whitequark: `ipv4_repr` → `ipv6_repr`

src/socket/raw.rs
```diff
@@ -205,7 +207,13 @@ impl<'a, 'b> RawSocket<'a, 'b> {
                     Ok((IpRepr::Ipv4(ipv4_repr), packet.payload()))
                 }
                 #[cfg(feature = "proto-ipv6")]
-                IpVersion::Ipv6 => Err(Error::Unrecognized),
+                IpVersion::Ipv6 => {
+                    let mut packet = Ipv6Packet::new_checked(buffer.as_mut())?;
+                    if packet.next_header() != protocol { return Err(Error::Unaddressable) }
```
whitequark: You should fill the checksum.

src/socket/raw.rs
```diff
@@ -291,123 +302,238 @@ mod test {
...
+        #[cfg(feature = "proto-ipv4")]
+        {
+            let socket = match RawSocket::new(IpVersion::Ipv4,
+                                              IpProtocol::Unknown(ipv4_locals::IP_PROTO+1),
+                                              buffer(1), buffer(1)) {
+                Socket::Raw(socket) => socket,
+                _ => unreachable!()
+            };
+            assert!(!socket.accepts(&ipv4_locals::HEADER_REPR));
+            #[cfg(feature = "proto-ipv6")]
```
whitequark: Useless `#[cfg]`.

src/socket/raw.rs
```diff
@@ -291,123 +302,238 @@ mod test {
...
+        #[cfg(feature = "proto-ipv6")]
+        {
+            let socket = match RawSocket::new(IpVersion::Ipv6,
+                                              IpProtocol::Unknown(ipv6_locals::IP_PROTO+1),
+                                              buffer(1), buffer(1)) {
+                Socket::Raw(socket) => socket,
+                _ => unreachable!()
+            };
+            assert!(!socket.accepts(&ipv6_locals::HEADER_REPR));
+            #[cfg(feature = "proto-ipv4")]
```
whitequark: As above.

src/socket/raw.rs
dlrobertson: IPv6 packets don't have a checksum. I assumed that the user takes care of the upper layer. Is that not correct?

src/socket/raw.rs
dlrobertson: I use an ipv6 specific variable below and this block is run if `proto-ipv4` is enabled, so if `proto-ipv4` is enabled but `proto-ipv6` is not you'll get a compile failure.

src/socket/raw.rs
dlrobertson: Same

src/socket/raw.rs
dlrobertson: Updated

src/socket/raw.rs
whitequark: Oh, I misread.

Issue #103, opened 2017-12-24
Processing of IPv6 packets


dlrobertson:
Now that #76 has been completed we can now move on to the handling of IPv6 packets. This metabug details the roadmap to add processing of IPv6 packets to smoltcp.
 
- Additions to `wire`
   - [x] [ICMPv6] (PR #128)
   - [x] [MLDv2] (PR #252)
   - [x] [NDISC] (PRs #167 #191 )
   - [x] Handle extension headers
       - [x] Add [IPv6] `Options` (PR #125)
       - [x] Add [IPv6] `Hop-by-Hop` extension header (PR #139)
       - [x] Add [IPv6] `Routing` extension header (PR #218)
       - [x] Add [IPv6] `Fragment` extension header (PR #168)
   - [x] Add [IPv6] `No Next` header
 - Updates to `socket`
   - [x] Update `IcmpSocket` to also handle [ICMPv6] packets (Issue #99 PR #205)
   - [x] Update `RawSocket` to also handle [IPv6] packets (PR #118) 
 - Updates to `iface`
   - [x] Enable `deny(unused)` when `proto-ipv6` is enabled, but `proto-ipv4` is not. (PR #121)
   - [x] Correctly handle [IPv6] packets, forwarding packets to the `accept`ing socket. (PR #121)
   - [x] Add Address Resolution via [NDISC] (PR #196)

Related to: #3

[ICMPv4]: https://tools.ietf.org/html/rfc792
[ICMPv6]: https://tools.ietf.org/html/rfc4443
[IPv6]: https://tools.ietf.org/html/rfc8200
[MLDv2]: https://tools.ietf.org/html/rfc3810
[NDISC]: https://tools.ietf.org/html/rfc4861\
[ADDRARCH 2.7.1]: https://tools.ietf.org/html/rfc4291#section-2.7.1

dlrobertson:
The very first PR should probably just handle IPv6 packets and forward them to the socket. The second should probably handle sending packets via `RawSocket`. After those are done then we can begin to tacke address resolution etc.
 
For a first pass at extension headers it may be worthwhile to add the extension headers to `wire` but only handle the `Pad1` and `PadN` options. Discarding packets with options according to the options type [RFC 8200 4.2]

[RFC 8200 4.2]: https://tools.ietf.org/html/rfc8200#section-4.2

whitequark:
:+1:

hjr3:
[...]

PR #125, opened 2018-01-17
Ipv6 extension option pad

hjr3:
This relates to https://github.com/m-labs/smoltcp/issues/103#issuecomment-357706855

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,11 @@ default = [
 ]
 
 [[example]]
+name = "packet2pcap"
+path = "utils/packet2pcap.rs"
+required-features = ["std"]
+
+[[example]]
 name = "tcpdump"
 required-features = ["std", "phy-raw_socket", "proto-ipv4"]
 
--- a/src/phy/pcap_writer.rs
+++ b/src/phy/pcap_writer.rs
@@ -92,15 +92,15 @@ impl<T: AsRef<PcapSink>> PcapSink for T {
 }
 
 #[cfg(feature = "std")]
-impl<T: AsMut<Write>> PcapSink for RefCell<T> {
+impl<T: Write> PcapSink for RefCell<T> {
     fn write(&self, data: &[u8]) {
-        self.borrow_mut().as_mut().write_all(data).expect("cannot write")
+        self.borrow_mut().write_all(data).expect("cannot write")
     }
 
     fn packet(&self, timestamp: u64, packet: &[u8]) {
         self.packet_header(timestamp, packet.len());
         PcapSink::write(self, packet);
-        self.borrow_mut().as_mut().flush().expect("cannot flush")
+        self.borrow_mut().flush().expect("cannot flush")
     }
 }
 
--- a/src/wire/ipv4.rs
+++ b/src/wire/ipv4.rs
@@ -184,6 +184,8 @@ impl<T: AsRef<[u8]>> Packet<T> {
 
     /// Ensure that no accessor method will panic if called.
     /// Returns `Err(Error::Truncated)` if the buffer is too short.
+    /// Returns `Err(Error::Malformed)` if the header length is greater
+    /// than total length.
     ///
     /// The result of this check is invalidated by calling [set_header_len]
     /// and [set_total_len].
@@ -196,6 +198,8 @@ impl<T: AsRef<[u8]>> Packet<T> {
             Err(Error::Truncated)
         } else if len < self.header_len() as usize {
             Err(Error::Truncated)
+        } else if self.header_len() as u16 > self.total_len() {
+            Err(Error::Malformed)
         } else if len < self.total_len() as usize {
             Err(Error::Truncated)
         } else {
@@ -737,6 +741,13 @@ mod test {
         packet.fill_checksum();
         let packet = Packet::new(&*packet.into_inner());
         assert_eq!(Repr::parse(&packet, &ChecksumCapabilities::default()), Err(Error::Malformed));
+    }
+
+    #[test]
+    fn test_parse_total_len_less_than_header_len() {
+        let mut bytes = vec![0; 40];
+        bytes[0] = 0x09;
+        assert_eq!(Packet::new_checked(&mut bytes), Err(Error::Malformed));
     }
 
     #[test]
--- a/utils/packet2pcap.rs
+++ b/utils/packet2pcap.rs
@@ -1,0 +1,71 @@
+extern crate smoltcp;
+extern crate getopts;
+
+use std::cell::RefCell;
+use std::io::{self, Read, Write};
+use std::path::Path;
+use std::fs::File;
+use std::env;
+use std::process::exit;
+use smoltcp::phy::{PcapLinkType, PcapSink};
+use getopts::Options;
+
+fn convert(packet_filename: &Path, pcap_filename: &Path, link_type: PcapLinkType)
+          -> io::Result<()> {
+    let mut packet_file = File::open(packet_filename)?;
+    let mut packet = Vec::new();
+    packet_file.read_to_end(&mut packet)?;
+
+    let pcap = RefCell::new(Vec::new());
+    PcapSink::global_header(&pcap, link_type);
+    PcapSink::packet(&pcap, 0, &packet[..]);
+
+    let mut pcap_file = File::create(pcap_filename)?;
+    pcap_file.write_all(&pcap.borrow()[..])?;
+
+    Ok(())
+}
+
+fn print_usage(program: &str, opts: Options) {
+    let brief = format!("Usage: {} [options] INPUT OUTPUT", program);
+    print!("{}", opts.usage(&brief));
+}
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let program = args[0].clone();
+
+    let mut opts = Options::new();
+    opts.optflag("h", "help", "print this help menu");
+    opts.optopt("t", "link-type", "set link type (one of: ethernet ip)", "TYPE");
+
+    let matches = match opts.parse(&args[1..]) {
+        Ok(m) => m,
+        Err(e) => {
+            eprintln!("{}", e);
+            return
+        }
+    };
+
+    let link_type =
+        match matches.opt_str("t").as_ref().map(|s| &s[..]) {
+            Some("ethernet") => Some(PcapLinkType::Ethernet),
+            Some("ip") => Some(PcapLinkType::Ip),
+            _ => None
+        };
+
+    if matches.opt_present("h") || matches.free.len() != 2 || link_type.is_none() {
+        print_usage(&program, opts);
+        return
+    }
+
+    match convert(Path::new(&matches.free[0]),
+                  Path::new(&matches.free[1]),
+                  link_type.unwrap()) {
+        Ok(()) => (),
+        Err(e) => {
+            eprintln!("Cannot convert packet to pcap: {}", e);
+            exit(1);
+        }
+    }
+}


hjr3:
@dlrobertson here is my PR for `Pad1` and `PadN`. If you see major problems, ping me and I can hop on IRC. I have two _TODO_ items to figure out, but they should not cause major change to the PR as-is.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub const TYPE:     usize = 0;
+    // 8-bit unsigned integer. Length of the DATA field of this option, in octets
+    pub const LENGTH:   usize = 1;
+	// Variable-length field. Option-Type-specific data.
+    pub const DATA:     Rest  = 2..;
+}
+
+impl<T: AsRef<[u8]>> Packet<T> {
+    /// Create a raw octet buffer with an IPv6 Extension Header Option packet structure.
+    pub fn new(buffer: T) -> Packet<T> {
```
dlrobertson: A `check_len` function should be added, and a `new_checked` function should be added. See some of the other implementations in `wire` for details. Make sure to test `check_len` well. It will be a little tricky since the length requirements vary based on the options type.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+        let data = self.buffer.as_ref();
+        data[field::LENGTH]
+    }
+
+    /// Return the option data.
+    ///
+    /// # Panics
+    /// The function panics if the type does not support this field
+    #[inline]
+    pub fn option_data(&self) -> &[u8] {
```
dlrobertson: See some of the other implementations of `payload` for `ipv4` etc and `data` for `icmpv4`.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+        let data = self.buffer.as_mut();
+        data[field::TYPE] = value.into();
+    }
+
+    /// Set the option data and data length
+    ///
+    /// # Panics
+    /// The function panics if the type does not support the length or data fields
+    #[inline]
+    pub fn set_option_data(&mut self, value: &[u8], len: u8) {
```
dlrobertson: `payload_mut` or `data_mut`. See some of the other implementations for details.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
+use {Error, Result};
+
+enum_with_unknown! {
+    /// IPv6 Extension Header Option type
+    pub doc enum OptionType(u8) {
```
dlrobertson: There should probably be a `OpFailureType` for the four on failure type defined in [RFC 8200 4.2]. Then a `From<OptionType> for OpFailureType`. E.g. something like

```
impl From<u8> for OpFailureType {
    fn from(other: u8) -> ExtOpFailureType {
        match other & 0b11000000 {
            0b00000000 => ExtOpFailureType::Skip,
            0b01000000 => ExtOpFailureType::Discard,
            0b10000000 => ExtOpFailureType::DiscardSendAll,
            0b11000000 => ExtOpFailureType::DiscardSendUnicast,
            _ => unreachable!()
        }
    }
}
```

[RFC 8200 4.2]: https://tools.ietf.org/html/rfc8200#section-4.2

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+            data[i] = value[(i-2)];
+        }
+    }
+}
+
+/// A high-level representation of an IPv6 Extension Header Option
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Repr<'a> {
+    Pad1 {
+        ident:  OptionType,
```
dlrobertson: I think `ident` is redundent. If you're using the `Pad1` `Repr` then `emit` knows that the option type should be that of `Pad1`.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+    },
+    PadN {
+        ident:  OptionType,
+        length: u8, // TODO -  should this value represent the total length or the data length?
```
dlrobertson: The length is just the length of the data.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+        // extra bytes in buffer
+        let bytes:  [u8; 10] = [0x1, 0x7, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xff];
+        let packet = Packet::new(&bytes);
+        assert_eq!(packet.option_type(), OptionType::PadN);
+        assert_eq!(packet.option_data_length(), 7);
+        assert_eq!(packet.option_data(), &[0, 0, 0, 0, 0, 0, 0]);
+
+        // unrecognized option actions
+        let bytes:  [u8; 1] = [0xff];
+        let packet = Packet::new(&bytes);
```
dlrobertson: Just FYI `Packet::new` bytes here should work, but `Packet::new_checked` here should fail.

src/wire/mod.rs
```diff
@@ -123,6 +125,10 @@ pub use self::ipv6::{Address as Ipv6Address,
                      Cidr as Ipv6Cidr,
                      MIN_MTU as IPV6_MIN_MTU};
 
+#[cfg(feature = "proto-ipv6")]
+pub use self::ipv6ext::{Packet as Ipv6OptionPacket,
```
dlrobertson: Don't forget the enums here.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// The function panics if the type does not support the length or data fields
+    #[inline]
+    pub fn set_option_data(&mut self, value: &[u8], len: u8) {
+        let data = self.buffer.as_mut();
+        data[field::LENGTH] = len;
+
+        let len = len as usize;
+
+        // TODO find idiomatic way of doing this
+        //data[field::DATA.start..len] = value[..len];
```
dlrobertson: The above will take care of this `TODO`

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /// Return the length of the data.
+    ///
+    /// # Panics
+    /// The function panics if the type does not support this field
+    #[inline]
+    pub fn option_data_length(&self) -> u8 {
+        let data = self.buffer.as_ref();
+        data[field::LENGTH]
```
dlrobertson: Create a `should_panic` test for this. Also I can't remember off the top of my head if an out of bounds index panics in release mode. If not, add an `assert!` checking the length.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,246 @@
...
+                    data: packet.option_data(),
+                })
+            }
+            _ => Err(Error::Unrecognized),
+        }
+    }
+
+    /// Return the length of a header that will be emitted from this high-level representation.
+    pub fn buffer_len(&self) -> usize {
+        // This function is not strictly necessary, but it can make client code more readable.
```
dlrobertson: I'd remove this comment. This is the case for headers like `ipv6` where the buffer length is static.

src/wire/ipv6ext.rs
hjr3: I did that at first, but removed it per:

> The three high-order bits described above are to be treated as part
   of the Option Type, not independent of the Option Type.  That is, a
   particular option is identified by a full 8-bit Option Type, not just
   the low-order 5 bits of an Option Type.

Thus, `0b00000005` and `0b00100005` were two distinct `OptionType`s.

dlrobertson: Yeah, you wouldn't change the current option type. I mean to add a new type that would be specifically for the following.

```
   The Option Type identifiers are internally encoded such that their
   highest-order 2 bits specify the action that must be taken if the
   processing IPv6 node does not recognize the Option Type:

      00 - skip over this option and continue processing the header.

      01 - discard the packet.

      10 - discard the packet and, regardless of whether or not the
           packet's Destination Address was a multicast address, send an
           ICMP Parameter Problem, Code 2, message to the packet's
           Source Address, pointing to the unrecognized Option Type.

      11 - discard the packet and, only if the packet's Destination
           Address was not a multicast address, send an ICMP Parameter
           Problem, Code 2, message to the packet's Source Address,
           pointing to the unrecognized Option Type.
```

This would essentially be used for something like

```rust
match packet.option_type() {
    ...
    OptionType::Unknown(type) {
        match OptFailureType::from(type) {
            OpFailureType::Skip => { skip the option }
            OpFailureType::Discard => { discard the packet }
            ...
        }
    }
}
```

I'm okay with tackling this in another issue and PR.

hjr3: I understand now., thank you. I will tackle that in a follow up PR.

dlrobertson: :+1: Sounds good

src/wire/ipv6ext.rs
hjr3: Added `check_len` and `new_checked`. I am not sure that I handled `Pad1 correctly. As is, calling `option_data_length` or `data` methods will panic if the type is `Pad1`.

src/wire/ipv6ext.rs
hjr3: added. based on some sample code, out of bounds index will panic in release mode

src/wire/ipv6ext.rs
hjr3: modified `option_data` to be `data` and conformed to the examples.

src/wire/ipv6ext.rs
hjr3: modified `option_data` to be `data` and conformed to the examples.

src/wire/ipv6ext.rs
hjr3: fixed

src/wire/ipv6ext.rs
hjr3: removed

src/wire/ipv6ext.rs
hjr3: 👍

src/wire/ipv6ext.rs
hjr3: removed

src/wire/ipv6ext.rs
hjr3: added test

src/wire/mod.rs
hjr3: exposed enums

hjr3:
I audited the other implementations. Only that I saw missing (besides CR comments) was `Display` implementations so I added them.

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,356 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            &OptionType::Pad1        => write!(f, "Pad1"),
+            &OptionType::PadN        => write!(f, "PadN"),
+            &OptionType::Unknown(id) => write!(f, "{}", id)
+        }
+    }
+}
+
+/// A read/write wrapper around a variable number of IPv6 Extension Header Options
```
dlrobertson: I don't think it is a variable number of IPv6 Extension Header Options. Each packet is only one Extension Header Option. E.g. if `bytes` contains the following

```
0x00, 0x01, 0x03, 0x00, 0x00, 0x00
```
This is a `Pad1` followed by a `PadN` with 3 bytes of padding, but when `packet = Packet::new(&bytes[..])` is run, `packet` is only the `Pad1` right?

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,356 @@
...
+//
+// See https://tools.ietf.org/html/rfc8200#section-4.2 for details.
+mod field {
+    use wire::field::*;
+
+    // 8-bit identifier of the type of option
+    pub const TYPE:     usize = 0;
+    // 8-bit unsigned integer. Length of the DATA field of this option, in octets
+    pub const LENGTH:   usize = 1;
+	// Variable-length field. Option-Type-specific data.
```
dlrobertson: indentation

src/wire/ipv6ext.rs
```diff
@@ -0,0 +1,356 @@
...
+
+        // unknown option type with truncated data
+        assert_eq!(Err(Error::Truncated), Packet::new(&PACKET_BYTES_UNKNOWN[..4]).check_len());
+        assert_eq!(Err(Error::Truncated), Packet::new(&PACKET_BYTES_UNKNOWN[..1]).check_len());
+        // unknown type
+        assert_eq!(Ok(()), Packet::new(&PACKET_BYTES_UNKNOWN).check_len());
+    }
+
+    #[test]
+    #[should_panic]
```
dlrobertson: nit: add an expect string `#[should_panic(expected = "<something>")]`

src/wire/mod.rs
```diff
@@ -85,6 +85,8 @@ pub(crate) mod ip;
 mod ipv4;
 #[cfg(feature = "proto-ipv6")]
 mod ipv6;
+#[cfg(feature = "proto-ipv6")]
+mod ipv6ext;
```
dlrobertson: nit: for other modules in `wire` if the module is `icmpv4` the exported names are `Icmpv4<something>. Perhaps `ipv6ops` or `ipv6option` would be more in-line with other implementations in `wire`.

src/wire/mod.rs
```diff
@@ -123,6 +125,11 @@ pub use self::ipv6::{Address as Ipv6Address,
                      Cidr as Ipv6Cidr,
                      MIN_MTU as IPV6_MIN_MTU};
 
+#[cfg(feature = "proto-ipv6")]
+pub use self::ipv6ext::{Packet as Ipv6OptionPacket,
+                        Repr as Ipv6OptionRepr,
+                        OptionType as IPv6OptionType};
```
dlrobertson: nit: `s/IPv6OptionType/Ipv6OptionType/`

src/wire/ipv6ext.rs
hjr3: Yes, that is correct. Fixed.

src/wire/ipv6ext.rs
hjr3: fixed

src/wire/ipv6ext.rs
hjr3: fixed

src/wire/mod.rs
hjr3: I chose `ipv6option`. I originally was going to put all of the extension headers in the `ipv6ext` module, but have a module for options seems fine to me.

src/wire/mod.rs
hjr3: fixed

dlrobertson:
IMO there are some doc strings that don't need a period. E.g. `IPv6 Extension Header Option Type`, but doc strings like `A read/write wrapper around a variable number of IPv6 Extension Header Options` probably should.

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            &OptionType::Pad1        => write!(f, "Pad1"),
+            &OptionType::PadN        => write!(f, "PadN"),
+            &OptionType::Unknown(id) => write!(f, "{}", id)
+        }
+    }
+}
+
+/// A read/write wrapper around an of IPv6 Extension Header Option
```
dlrobertson: nit: `s/of //`

src/wire/mod.rs
```diff
@@ -123,6 +125,11 @@ pub use self::ipv6::{Address as Ipv6Address,
                      Cidr as Ipv6Cidr,
                      MIN_MTU as IPV6_MIN_MTU};
 
+#[cfg(feature = "proto-ipv6")]
+pub use self::ipv6option::{Packet as Ipv6OptionPacket,
+                        Repr as Ipv6OptionRepr,
```
dlrobertson: nit: indentation

hjr3: hmm, looks like my vim settings are masking some indentation problems.

hjr3: n/m, that was simply my oversight.

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+    #[inline]
+    pub fn option_type(&self) -> OptionType {
+        let data = self.buffer.as_ref();
+        OptionType::from(data[field::TYPE])
+    }
+
+    /// Return the length of the data.
+    ///
+    /// # Panics
+    /// The function panics if the type does not support this field
```
dlrobertson: period

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+        assert_eq!(Ok(()), Packet::new(&PACKET_BYTES_PAD1).check_len());
+
+        // padn with truncated data
+        assert_eq!(Err(Error::Truncated), Packet::new(&PACKET_BYTES_PADN[..2]).check_len());
+        // padn
+        assert_eq!(Ok(()), Packet::new(&PACKET_BYTES_PADN).check_len());
+
+        // unknown option type with truncated data
+        assert_eq!(Err(Error::Truncated), Packet::new(&PACKET_BYTES_UNKNOWN[..4]).check_len());
+        assert_eq!(Err(Error::Truncated), Packet::new(&PACKET_BYTES_UNKNOWN[..1]).check_len());
```
dlrobertson: You could use `new_checked` here, but this works too.

src/wire/mod.rs
dlrobertson: +1 for vim

src/wire/ipv6option.rs
hjr3: fixed

src/wire/ipv6option.rs
hjr3: I decided to use `check_len` to ensure that it is working correctly independently of `new_checked`. I test `new_checked` in `test_option_deconstruct` too.

hjr3:
@dlrobertson Thank you for the reviews! I will make sure to follow the style more closely in the future.

src/wire/ipv6option.rs
dlrobertson: Ah, makes sense

src/wire/ipv6option.rs
dlrobertson: @hjr3 I don't see the updated version

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/// A high-level representation of an IPv6 Extension Header Option.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Repr<'a> {
+    Pad1,
+    PadN {
+        length: u8,
+        data:   &'a [u8]
+    },
```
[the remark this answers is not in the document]
hjr3: I imagine one would want to know the identifier of the unknown option too?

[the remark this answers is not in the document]
whitequark: Yeah, of course.

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
+use core::fmt;
+use {Error, Result};
+
+enum_with_unknown! {
+    /// IPv6 Extension Header Option Type
+    pub doc enum OptionType(u8) {
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+        /// Pad1 option
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+        Pad1 =  0,
+        /// PadN option
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+            &OptionType::Pad1        => write!(f, "Pad1"),
+            &OptionType::PadN        => write!(f, "PadN"),
+            &OptionType::Unknown(id) => write!(f, "{}", id)
+        }
+    }
+}
+
+/// A read/write wrapper around an IPv6 Extension Header Option.
+#[derive(Debug, PartialEq)]
+pub struct Packet<T: AsRef<[u8]>> {
```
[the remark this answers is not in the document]
hjr3: yup makes sense. fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+- - - - - - - - -
+// |  Option Type  |  Opt Data Len |  Option Data
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+- - - - - - - - -
+//
+//
+// See https://tools.ietf.org/html/rfc8200#section-4.2 for details.
+mod field {
+    use wire::field::*;
+
+    // 8-bit identifier of the type of option
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub const TYPE:     usize = 0;
+    // 8-bit unsigned integer. Length of the DATA field of this option, in octets
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+
+        if self.option_type() == OptionType::Pad1 {
+            return Ok(());
+        }
+
+        if len < field::DATA.start {
+            return Err(Error::Truncated);
+        }
+
+        if data[field::DATA].len() < (self.option_data_length() as usize) {
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+        let data = self.buffer.as_ref();
+        OptionType::from(data[field::TYPE])
+    }
+
+    /// Return the length of the data.
+    ///
+    /// # Panics
+    /// The function panics if the type does not support this field
+    #[inline]
+    pub fn option_data_length(&self) -> u8 {
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+        let data = self.buffer.as_ref();
+        data[field::LENGTH]
+    }
+
+    /// Return the start and end of option data.
```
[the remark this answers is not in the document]
hjr3: `field::DATA` is a function now.

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Set the option type.
+    #[inline]
+    pub fn set_option_type(&mut self, value: OptionType) {
+        let data = self.buffer.as_mut();
+        data[field::TYPE] = value.into();
+    }
+
+    /// Set the option data length.
+    #[inline]
+    pub fn set_option_data_length(&mut self, value: u8) {
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+}
+
+/// A high-level representation of an IPv6 Extension Header Option.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Repr<'a> {
+    Pad1,
+    PadN {
+        length: u8,
+        data:   &'a [u8]
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Emit a high-level representation into an IPv6 Extension Header Option packet.
+    pub fn emit<T: AsRef<[u8]> + AsMut<[u8]> + ?Sized>(&self, packet: &mut Packet<&'a mut T>) {
+        match self {
+            &Repr::Pad1 => {
+                packet.set_option_type(From::from(OptionType::Pad1));
+            }
+            &Repr::PadN{length, data} => {
+                let len = length as usize;
+                packet.set_option_type(From::from(OptionType::PadN));
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+
+            &Repr::__Nonexhaustive => unreachable!()
+        }
+    }
+
+    /// Emit a high-level representation into an IPv6 Extension Header Option packet.
+    pub fn emit<T: AsRef<[u8]> + AsMut<[u8]> + ?Sized>(&self, packet: &mut Packet<&'a mut T>) {
+        match self {
+            &Repr::Pad1 => {
+                packet.set_option_type(From::from(OptionType::Pad1));
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+            }
+
+            &Repr::__Nonexhaustive => unreachable!()
+        }
+    }
+}
+
+impl<'a> fmt::Display for Repr<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let name = "IPv6 Extension Option";
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,356 @@
...
+        match self {
+            &Repr::Pad1 => {
+                write!(f, "{} type={} ", &name, OptionType::Pad1)
```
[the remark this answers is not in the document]
hjr3: fixed

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,367 @@
...
+
+        if len < field::LENGTH {
+            return Err(Error::Truncated);
+        }
+
+        if self.option_type() == Type::Pad1 {
+            return Ok(());
+        }
+
+        let df = field::DATA(len as u8);
```
dlrobertson: The `len` used as input to `DATA` should be the result of `data[field::LENGTH]` instead of the buffer length. Also don't make the same mistake I did with ICMPv4 and make sure to check that the buffer is bigger than that before using that field.

src/wire/ipv6option.rs
```diff
@@ -0,0 +1,367 @@
...
+// See https://tools.ietf.org/html/rfc8200#section-4.2 for details.
+mod field {
+    use wire::field::*;
+
+    // 8-bit identifier of the type of option.
+    pub const TYPE:     usize = 0;
+    // 8-bit unsigned integer. Length of the DATA field of this option, in octets.
+    pub const LENGTH:   usize = 1;
+    // Variable-length field. Option-Type-specific data.
+    pub fn DATA(length: u8) -> Field {
```
dlrobertson: `allow(non_snake_case)` if we're sticking to `DATA`.

hjr3: i have this fixed in my latest changes

src/wire/ipv6option.rs
hjr3: i am fixing this right now

hjr3: this is fixed now. `check_len` is tricky because of the special `Pad1` case and the variable length field.

dlrobertson: Yup. Thanks for the update.

src/wire/ipv6option.rs
dlrobertson: :+1: Thanks! Just a side note. I personally I tend to prefer using

```rust
mod something {
    // other stuff in the module

    #[allow(non_snake_case)]
    fn EXCEPTIONTOTHERULE() {
    }
}
```

instead of

```rust
mod something {
    #![allow(non_snake_case)]

    // other stuff in the module

    fn EXCEPTIONTOTHERULE() {
    }
}
```

whitequark: No, this is fine in `mod fields`, since it has a specific structure.

hjr3: I was following the convention as I saw it here: https://github.com/m-labs/smoltcp/blob/c2d18ec071a2cd68aee2724ce13eefa6ccf1f0c1/src/wire/tcp.rs#L75

whitequark:
Thanks!

PR #126, opened 2018-01-18
Increase required size for ICMPv4 packets

dlrobertson:
- Increase the size required in Icmpv4Packet::check_len to 8 bytes
 - Add a test

No valid ICMPv4 packets are less than 8 bytes in length.


whitequark:
Wasn't there another length check somewhere?

dlrobertson:
There is another check just below this with `header_len`. `header_len` is still 4 for all unknown types though. Perhaps this change should be to `header_len`. Then we can simplify `check_len`.

dlrobertson:
Updated to be more reasonable.

whitequark:
Thanks!

PR #128, opened 2018-01-19
Add ICMPv6

dlrobertson:
Add support for ICMPv6 message types to the wire module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,11 @@ matrix:
         MODE='build'
     - rust: nightly
       env: MODE='fuzz run' ARGS='packet_parser -- -max_len=1536 -max_total_time=30'
+  allow_failures:
+    # something's screwy with Travis (as usual) and cargo-fuzz dies with a LeakSanitizer error
+    # even when it's successful. Keep this in .travis.yml in case it starts working some day.
+    - rust: nightly
+      env: MODE='fuzz run' ARGS='packet_parser -- -max_len=1536 -max_total_time=30'
 before_script:
   - if [ "$MODE" == "fuzz run" ]; then cargo install cargo-fuzz; fi
 script:


src/wire/icmpv6.rs
```diff
@@ -0,0 +1,690 @@
...
+        let packet = Self::new(buffer);
+        packet.check_len()?;
+        Ok(packet)
+    }
+
+    /// Ensure that no accessor method will panic if called.
+    /// Returns `Err(Error::Truncated)` if the buffer is too short.
+    pub fn check_len(&self) -> Result<()> {
+        let len = self.buffer.as_ref().len();
+        if len < self.header_len() {
```
dlrobertson: I'm not sure if this is correct. `header_len` returns results more similar to the older version of ICMPv4's `header_len` (can be 4 or 8). This is so that `payload` can be more useful for NDISC and MLDv2 etc that need the last 4 bytes of the header. That being said, no valid ICMPv6 message is less than 8 bytes in length.

whitequark: This will actually panic if you supply an empty buffer because `header_len` checks `msg_type`. If no valid ICMPv6 message is less than 8 bytes in length you should just use `field::UNUSED.end` (actually I prefer it to be called `field::HEADER_END`.

src/wire/icmpv6.rs
```diff
@@ -0,0 +1,690 @@
...
+    /// Return the checksum field.
+    #[inline]
+    pub fn checksum(&self) -> u16 {
+        let data = self.buffer.as_ref();
+        NetworkEndian::read_u16(&data[field::CHECKSUM])
+    }
+
+    /// Return the identifier field (for echo request and reply packets).
+    ///
+    /// # Panics
```
whitequark: None of these may panic if `check_len` verifies that the packet is bigger than 8 bytes, too.

src/wire/icmpv6.rs
```diff
@@ -0,0 +1,690 @@
...
+    #[inline]
+    pub fn echo_seq_no(&self) -> u16 {
+        let data = self.buffer.as_ref();
+        NetworkEndian::read_u16(&data[field::ECHO_SEQNO])
+    }
+
+    /// Return the MTU field (for packet too big messages).
+    ///
+    /// # Panics
+    /// This function may panic if this packet is not an packet too big packet.
```
whitequark: `a packet`, also `a parameter` below.

src/wire/icmpv6.rs
dlrobertson: :+1:

src/wire/icmpv6.rs
dlrobertson: True. Maybe this should be updated to something like

```
# Warning

This function may return invalid data if the packet is...
```

whitequark: That's obviously true from its name. The notice about panics was just documenting an implementation detail, really.

dlrobertson:
Thanks for the review and adding `cargo fuzz`, updated and rebased.

PR #133, opened 2018-01-26
Replace unspecified src IP addrs with the interface IP

### LINKED ISSUE #127: smoltcp doesn't replace unspecified src addresses with local addresses for UDP and TCP sockets
**Problem:** Right now binding a UDP or a TCP socket to `0.0.0.0:*` results in packets being sent with src IP `0.0.0.0`, not with the IP assigned to the interface, even though ICMP and Raw sockets do the replacement.

**Cause:** Both `UdpSocket::dispatch` and `TcpSocket::dispatch` use `IpRepr::Unspecified` and set `src_addr` to `local_endpoint`, wich is either `IpAddress::Ipv4` or `IpAddress::Ipv6`, even if the IP addr in question is itself unspecified, i.e. `IpAddress::Ipv4(0, 0, 0, 0)`. `IpRepr::lower` treat such combinations as not requiering the replacement: https://github.com/m-labs/smoltcp/blob/master/src/wire/ip.rs#L498 .

**Possible solutions:** 
1. Change `IpRepr::lower` to consider case `IpRepr::Unspecified(src_addr: IpAddr::Ipv*(src_addr), ..) if src_addr.is_unspecified()` as an equivalent to `IpRepr::Unspecified(src_addr: IpAddr::Unspecified, ..)`. I've done it in https://github.com/batonius/smoltcp/commit/8e4422cee017f60232cab7078efa683a70a733c9 , but it breaks `TcpSocket` since it uses `IpRepr::lower` with empty address list: https://github.com/m-labs/smoltcp/blob/master/src/socket/tcp.rs#L1315 . I'm not sure how to proceed from here - we can either provide some IPs to `IpRepr::lower` in `TcpSocket` or add a special case for empty `fallback_src_addrs` in `IpRepr::lower`.
2. Change `UdpSocket::dispatch` and `TcpSocket::dispatch` to check `src_addr` and replace it with `IpAddr::Unspecified`. This would fix the problem, but I'm not sure if there are other places where `IpRepr::Unspecified` could be used with `IpAddr::Ipv4`/`IpAddr::Ipv6` with an unspecified addr.

**Comment:** I was gonna fix it myself, but I'm not sure which of the options would be the optimal, so I'm creating the issue first.

batonius:
As discussed in #127, `IpRepr::lower` should treat `IpRepr::Unspecified(src_addr: IpAddr::Ipv*(src_addr), ..) if src_addr.is_unspecified()` as `IpRepr::Unspecified(src_addr: IpAddr::Unspecified, ..)`, i.e. try to replace `src_addr` with the concrete IP address assigned to the interface, unless `fallback_src_addrs` is empty, in which case it should keep the unspecified `src_addr`.

@dlrobertson I replaced the code for IPv6 case with an equivalent of the IPv4 one for uniformity and because the fix is more clear this way.

Closes #127


src/wire/ip.rs
```diff
@@ -496,75 +496,88 @@ impl Repr {
         match self {
             #[cfg(feature = "proto-ipv4")]
             &Repr::Unspecified {
-                src_addr: Address::Ipv4(src_addr),
+                src_addr: src_addr @ Address::Unspecified,
                 dst_addr: Address::Ipv4(dst_addr),
                 protocol, payload_len, hop_limit
-            } => {
+            } |
+            &Repr::Unspecified {
```
whitequark: Shouldn't you remove the `resolve_unspecified` macro above and the wildcard cases below? Or am I missing something?

batonius: No, it's used below for the `Repr::Ipv4` and `Repr::Ipv6` cases.

batonius: Since the branches in my case are equivalent, I could write a similar macro if you think that would be better. I didn't because `Ipv6Repr` and `Ipv4Repr` differ at `proto`/`next_header` so the macro would be somewhat messy.

PR #134, opened 2018-01-26
Fix icmpv4 check_len

dlrobertson:
- Ensure check_len does not require a packet to be of a certain length
 - Make icmpv4 check_len more similar to icmpv6 check_len


whitequark:
Hmmm, why wasn't this found by fuzzing?

dlrobertson:
Totally my bad. I just forgot to run it. Maybe it should be added to the ruby script?

whitequark:
> Maybe it should be added to the ruby script?

We can just run it on Travis, something like 30 seconds of fuzzing on each run.

dlrobertson:
:+1: yeah that seems reasonable

PR #136, opened 2018-01-28
rewrite ::phy::raw_socket

LuoZijun:
@whitequark 

Changes:

1. Add `cfg-if` crate.
2. Update `libc` version to latest.
3. Make examples `httpclient/ping/server/client/benchmark` only work on linux platform.
4. Update example `tcpdump`.
5. Add `phy::LinkLayer`.
6. Rewrite `phy::RawSocket` and `phy::sys::RawSocket`.

Problems:

1. BPF buffer reader ( see [code](https://github.com/m-labs/smoltcp/pull/136/files#diff-75b0666ac9f677ab0e26f16624beceb2R89) )


whitequark:
1. Please remove any changes related to mio. I do not want mio as a dependency of smoltcp, it is too large and complex.
2. Please do not remove tap_interface. As explained in README, on Linux tap interfaces can be used without root privileges (or CAP_NET_ADMIN). However, raw sockets cannot be.
3. You removed RxToken and TxToken implementations, and as a result raw sockets can not be used anymore to run the smoltcp stack against a raw socket. For example, it's not possible to do `cargo run --example server en0`. This needs to be fixed.
4. Why did you remove phy::pcap_writer and phy::tracer? These are not at all unused. They are a part of our public API and used in the examples. Put them back.

LuoZijun:
@whitequark 

1. i have not add `mio` as a dependency of smoltcp, i just said u can easy take `raw_fd` by `Rawsocket::as_raw_fd`.
then u can do everthing u want.
you can use that `raw_fd` with `mio` (linux/macos/bsd-like/windows) , 
you also can use that `raw_fd` with `libc::select` (on linux).
you have choice.

2. so this is good for linux platform, i'm happy make that back, but just on linux platform.

3. I think `RxToken` and `TxToken` implementations have a performance problem, u can see 
line: [src/phy/raw_socket.rs#L95](https://github.com/m-labs/smoltcp/blob/master/src/phy/raw_socket.rs#L95)
line: [src/phy/raw_socket.rs#L52](https://github.com/m-labs/smoltcp/blob/master/src/phy/raw_socket.rs#L52)
everytime they alloc one vec.
maybe we can redesign that ?

4. I have no problem with `phy::pcap_writer/phy::tracer` , disable that just for compiling with no error.

LuoZijun:
@whitequark 

i have an idea for question 2 ( linux tap_interface ): 

```rust
use libc;

struct RawSocket { }

impl RawSocket {
    pub fn with_ifname(ifname: &str) -> Result<RawSocket, io::Error> {
        if unsafe { libc::get_uid() != 0 } {
            libc::open("/dev/net/tun\0".as_ptr() as *const libc::c_char, libc::O_RDWR)
        } else {
            // root
            libc::socket(libc::AF_PACKET, libc::SOCK_RAW, protocol as i32)
        }
    }
}
```

does this work ?

whitequark:
No. The ioctls are different. Just leave the tap interface as it is.

whitequark:
> I think RxToken and TxToken implementations have a performance problem

Performance should be addressed in a different PR, not this one.

LuoZijun:
@whitequark 

make `bsd bpf` work with `RxToken` is not easy,

take a look here：

[bpf::recv](https://github.com/LuoZijun/smoltcp/blob/patch-3/src/phy/raw_socket.rs#L324-L335)
[bpf::packet_reader](https://github.com/LuoZijun/smoltcp/blob/patch-3/src/phy/raw_socket.rs#L35-L57)

Example:

```rust
let bpf_fd = BPF::open();
let buffer = [0u8; 70000];
let len = libc::read(bpf_fd, &mut buffer);

```

on linux, `&buffer[..len]` == one net packet,
on macos, not like linux, `&buffer[..len]` == one or more net packet.

so `fn consume<R, F>(self, timestamp: u64, f: FnOnce(&[u8]) -> Result<R>) -> Result<R>;` can not work on macos.

any suggestion ?

whitequark:
Put each packet into an individual Vec, then put all of them into VecDeque.

LuoZijun:
@whitequark 

that can't work with `phy::wait` .

whitequark:
We can place an additional condition that `receive` should have returned `None` before `phy::wait` can be run. What do you think?

LuoZijun:
@whitequark 

Okay, i will try it tonight.

LuoZijun:
@whitequark 

Changes:

1. Add `cfg-if` crate.
2. Update `libc` version to latest.
3. Make examples `httpclient/ping/server/client/benchmark` only work on linux platform.
4. Update example `tcpdump`.
5. Add `phy::LinkLayer`.
6. Rewrite `phy::RawSocket` and `phy::sys::RawSocket`.

Problems:

1. BPF buffer reader ( see [code](https://github.com/m-labs/smoltcp/pull/136/files#diff-75b0666ac9f677ab0e26f16624beceb2R89) )

src/phy/sys/raw_socket.rs
```diff
@@ -1,81 +1,310 @@
...
+impl RawSocket {
+    pub fn with_ifname(ifname: &str) -> Result<RawSocket, io::Error> {
+        let flags = sys::if_name_to_flags(ifname).unwrap();
+        let link_layer = 
+            if flags & sys::IFF_LOOPBACK != 0 {
+                LinkLayer::Eth
+            } else if flags & sys::IFF_BROADCAST != 0 {
+                LinkLayer::Eth
+            } else if flags & sys::IFF_POINTOPOINT != 0 {
+                if flags & sys::IFF_NO_PI as i32 != 0 {
```
LuoZijun: i'm not sure this, when we create one `tun` interface with `IFF_NO_PI` flag on linux.

src/phy/raw_socket.rs
```diff
@@ -64,16 +85,74 @@ impl<'a> Device<'a> for RawSocket {
         }
     }
 
+    #[cfg(any(target_os = "macos", target_os = "freebsd"))]
+    fn receive(&'a mut self) -> Option<(Self::RxToken, Self::TxToken)> {
+        let mut inner = self.inner.borrow_mut();
+
+        let mut buffer = vec![0; self.blen];
```
LuoZijun: BPF Reader problem.

examples/benchmark.rs
```diff
@@ -63,6 +64,15 @@ fn client(kind: Client) {
 
 static CLIENT_DONE: AtomicBool = ATOMIC_BOOL_INIT;
 
+
+#[cfg(not(all(feature = "phy-tap_interface", 
+          all(target_os = "linux", target_env = "gnu"))))]
+fn main() {
```
whitequark: Please do not do this. Instead, add the `phy-tap_interface` feature as a dependency in `Cargo.toml`.

Cargo.toml
```diff
@@ -42,6 +43,10 @@ default = [
   "socket-raw", "socket-icmp", "socket-udp", "socket-tcp"
 ]
 
+[[example]]
+name = "loopback"
+required-features = ["log", "proto-ipv4", "socket-tcp"]
+
```
whitequark: Please do not just move things around.

whitequark:
This PR is very complicated and nearly impossible to review. How about splitting it into multiple parts?

LuoZijun:
@whitequark 

now, history is rewrite, so if u still want multiple PRs, just desciption your want.

whitequark:
Why did you close this?

LuoZijun:
Oh, a mistake, i just delete all branch on my fork ...

how can i get back ?

whitequark:
https://github.com/m-labs/smoltcp/pull/136.diff

LuoZijun:
ok, i will copy that.

LuoZijun:
Back with `Restore branch` :)

m-labs-homu:
:umbrella: The latest upstream changes (presumably c418b60b5db93753999ae51d33ec4dc1d1631c69) made this pull request unmergeable. Please resolve the merge conflicts.

whitequark:
Thanks for your work. Ultimately, I have decided to merge PR #268 instead.

PR #140, opened 2018-01-31
Use separate metadata and payload buffers for UDP sockets

### LINKED ISSUE #138: UDP socket buffer wastes space
The rx buffer of `UdpSocket` being a `RingBuffer` of `PacketBuffer`s forces users to allocate much larger buffers than needed. If they want to be able to receive MTU sized packets, each `PacketBuffer` in the ring must have a MTU sized payload buffer. The problem is that a received UDP packet always fills a whole slot in the receive buffer, even if its payload is much smaller than the MTU. For example, if we allocate a two MTU sized packet buffers (so that we can receive two MTU sized packets), we can only receive two 10 byte UDP packets until the ring buffer is full. Most of the backing store is left unused.

A byte based buffer could utilize the available buffer space more effectively. For example, with a rx buffer size of 2*MTU we could either receive two MTU sized packets or hundreds of small packets. The `RingBuffer` already supports a byte based interface for tcp, so this part would be relatively easy.

The difficulty is that an UDP socket can receive data from multiple endpoints and should always return a the payload from a single packet on `recv` (instead of just a byte stream across packets like in TCP). My idea would be to use two ring buffers: one for the byte based backing storage and one for the metadata (PacketBuffer, packet size, packet start in byte buffer). This way, we could have separate limits for the amount of receivable bytes and the number of receivable packets. What do you think?

phil-opp:
Closes #138

This PR changes UDP sockets to use separate buffers for payloads and metadata. This allows a more flexible utilization of buffer space, since now small packets only occupy a small amount of the payload buffer. See #138 for more details.

A `UdpSocketBuffer` now has two arguments: A buffer used for storing metadata and a buffer used for storing the raw payload bytes. The former limits the number of packets in the buffer (since each packet needs a metadata entry) and the latter limits the accumulated packet size (since the payloads of each packet needs to be stored in it). The metadata buffer can be easily constructed via `Default::default`.

One complication is that we want contiguous slices for sending and receiving, but the `RingBuffer<u8>` wraps at its end. I worked around this problem by adding a dummy packet that occupies the remaining space up to the buffer end if a wrap around would occur. Thus, the buffer wraps around to the beginning so that we can insert our slice in a contiguous way. Unfortunately, this workaround makes the implementations of `send`, `recv`, `process`, and `dispatch` more complicated, but I think the better space utilization and the improved ergonomics are more than worth it.

### DIFF
--- a/src/socket/udp.rs
+++ b/src/socket/udp.rs
@@ -45,18 +45,15 @@ impl<'a, 'b> SocketBuffer<'a, 'b> {
         self.metadata_buffer.is_empty()
     }
 
-    fn check_capacity(&self, required_size: usize) -> Result<()> {
+    fn enqueue(&mut self, required_size: usize, endpoint: IpEndpoint) -> Result<&mut [u8]> {
+        let window = self.payload_buffer.window();
+        let contig_window = self.payload_buffer.contiguous_window();
+
         if self.metadata_buffer.is_full() || self.payload_buffer.window() < required_size {
-            Err(Error::Exhausted)
-        } else {
-            Ok(())
+            return Err(Error::Exhausted);
         }
-    }
 
-    fn prepare_for_insert(&mut self, required_size: usize) -> Result<()> {
-        self.check_capacity(required_size)?;
-
-        if self.payload_buffer.contiguous_window() < required_size {
+        if contig_window < required_size {
             // we reached the end of buffer, so the data does not fit without wrap-around
             // -> insert padding and try again
             self.payload_buffer.enqueue_many(required_size);
@@ -64,9 +61,17 @@ impl<'a, 'b> SocketBuffer<'a, 'b> {
             metadata_buf.padding = true;
             metadata_buf.size = required_size;
             metadata_buf.endpoint = IpEndpoint::default();
+            if window - contig_window < required_size {
+                return Err(Error::Exhausted);
+            }
         }
 
-        self.check_capacity(required_size)
+        let metadata_buf = self.metadata_buffer.enqueue_one()?;
+        metadata_buf.endpoint = endpoint;
+        metadata_buf.size = required_size;
+        metadata_buf.padding = false;
+
+        Ok(self.payload_buffer.enqueue_many(required_size))
     }
 }
 
@@ -178,18 +183,12 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
         if self.endpoint.port == 0 { return Err(Error::Unaddressable) }
         if !endpoint.is_specified() { return Err(Error::Unaddressable) }
 
-        self.tx_buffer.prepare_for_insert(size)?;
+        let payload_buf = self.tx_buffer.enqueue(size, endpoint)?;
 
-        let payload_buf = self.tx_buffer.payload_buffer.enqueue_many(size);
         debug_assert_eq!(payload_buf.len(), size);
 
-        let metadata_buf = self.tx_buffer.metadata_buffer.enqueue_one()?;
-        metadata_buf.endpoint = endpoint;
-        metadata_buf.size = size;
-        metadata_buf.padding = false;
-
         net_trace!("{}:{}:{}: buffer to send {} octets",
-                   self.meta.handle, self.endpoint, metadata_buf.endpoint, size);
+                   self.meta.handle, self.endpoint, endpoint, size);
         Ok(payload_buf)
     }
 
@@ -246,20 +245,15 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
         debug_assert!(self.accepts(ip_repr, repr));
 
         let size = repr.payload.len();
-        self.rx_buffer.prepare_for_insert(size)?;
 
-        let payload_buf = self.rx_buffer.payload_buffer.enqueue_many(size);
+        let endpoint = IpEndpoint { addr: ip_repr.src_addr(), port: repr.src_port };
+        let payload_buf = self.rx_buffer.enqueue(size, endpoint)?;
         assert_eq!(payload_buf.len(), size);
-
-        let metadata_buf = self.rx_buffer.metadata_buffer.enqueue_one()?;
-        metadata_buf.endpoint = IpEndpoint { addr: ip_repr.src_addr(), port: repr.src_port };
-        metadata_buf.size = size;
-        metadata_buf.padding = false;
         payload_buf.copy_from_slice(repr.payload);
 
         net_trace!("{}:{}:{}: receiving {} octets",
                    self.meta.handle, self.endpoint,
-                   metadata_buf.endpoint, metadata_buf.size);
+                   endpoint, size);
         Ok(())
     }
 


examples/server.rs
```diff
@@ -34,8 +34,8 @@ fn main() {
 
     let neighbor_cache = NeighborCache::new(BTreeMap::new());
 
-    let udp_rx_buffer = UdpSocketBuffer::new(vec![UdpPacketBuffer::new(vec![0; 64])]);
-    let udp_tx_buffer = UdpSocketBuffer::new(vec![UdpPacketBuffer::new(vec![0; 128])]);
+    let udp_rx_buffer = UdpSocketBuffer::new(vec![Default::default()], vec![0; 64]);
```
whitequark: I would prefer explicit initialization in *examples*, to make it clear what is being constructed.

src/socket/udp.rs
```diff
@@ -1,76 +1,68 @@
...
-            where T: Into<Managed<'a, [u8]>> {
-        PacketBuffer {
-            endpoint: IpEndpoint::default(),
-            size:     0,
-            payload:  payload.into()
-        }
-    }
+/// An UDP packet ring buffer.
+#[derive(Debug)]
+pub struct SocketBuffer<'a> {
```
whitequark: This has to have two lifetimes. If you think it can be one, try to construct multiple socket buffers without using `Vec`.

src/socket/udp.rs
```diff
@@ -156,18 +148,44 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
     /// to its payload.
     ///
     /// This function returns `Err(Error::Exhausted)` if the transmit buffer is full,
-    /// `Err(Error::Truncated)` if the requested size is larger than the packet buffer
+    /// `Err(Error::Truncated)` if the requested size is larger than the payload buffer
```
whitequark: It is not necessary to return `Error::Truncated` from this function anymore since there are no separate packet buffers.

src/socket/udp.rs
```diff
@@ -183,11 +201,21 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
...
     pub fn recv(&mut self) -> Result<(&[u8], IpEndpoint)> {
-        let packet_buf = self.rx_buffer.dequeue_one()?;
+        let mut metadata_buf = *self.rx_buffer.metadata_buffer.dequeue_one()?;
+        if metadata_buf.dummy {
+            // packet is dummy packet -> drop it and try again
+            self.rx_buffer.payload_buffer.dequeue_many(metadata_buf.payload_size);
+            metadata_buf = *self.rx_buffer.metadata_buffer.dequeue_one()?;
+        }
+
+        assert!(!metadata_buf.dummy);
```
whitequark: `debug_assert!`

src/socket/udp.rs
```diff
@@ -183,11 +201,21 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
...
+        let payload_buf = self.rx_buffer.payload_buffer.dequeue_many(metadata_buf.payload_size);
+        assert_eq!(metadata_buf.payload_size, payload_buf.len()); // ensured by inserting logic
```
whitequark: `debug_assert!`

src/socket/udp.rs
```diff
@@ -212,12 +240,40 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
     pub(crate) fn process(&mut self, ip_repr: &IpRepr, repr: &UdpRepr) -> Result<()> {
         debug_assert!(self.accepts(ip_repr, repr));
 
-        let packet_buf = self.rx_buffer.enqueue_one_with(|buf| buf.resize(repr.payload.len()))?;
-        packet_buf.as_mut().copy_from_slice(repr.payload);
-        packet_buf.endpoint = IpEndpoint { addr: ip_repr.src_addr(), port: repr.src_port };
+        let size = repr.payload.len();
+
+        if self.rx_buffer.payload_buffer.capacity() < size {
+            return Err(Error::Truncated);
```
whitequark: Same as above.

src/socket/udp.rs
```diff
@@ -212,12 +240,40 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
...
+        }
+
+        if self.rx_buffer.metadata_buffer.is_full() || self.rx_buffer.payload_buffer.window() < size {
```
whitequark: You should take this common logic and put it into `impl SocketBuffer`.

src/socket/udp.rs
```diff
@@ -226,24 +282,47 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
...
+            } else {
+                Err(Error::Exhausted) // don't dequeue metadata
+            }
+        });
+        if let Ok(size) = result {
+            payload_buffer.dequeue_many(size); // dequeue dummy payload
+        }
+
+        metadata_buffer.dequeue_one_with(move |metadata_buf| {
+            assert!(!metadata_buf.dummy);
```
whitequark: `debug_assert!`

src/storage/ring_buffer.rs
```diff
@@ -72,6 +72,11 @@ impl<'a, T: 'a> RingBuffer<'a, T> {
         self.capacity() - self.len()
     }
 
+    /// Return the number of contiguos elements that can be added to the ring buffer.
```
whitequark: This comment doesn't make sense.

    /// Return the largest number of elements that can be added to the buffer
    /// without wrapping around (i.e. in a single `enqueue_many` call).

phil-opp:
@whitequark I addressed your comments and added tests.

src/socket/udp.rs
```diff
@@ -1,58 +1,74 @@
...
-    fn as_ref<'b>(&'b self) -> &'b [u8] {
-        &self.payload[..self.size]
+    fn is_full(&self) -> bool {
+        self.metadata_buffer.is_full() || self.payload_buffer.is_full()
     }
 
-    fn as_mut<'b>(&'b mut self) -> &'b mut [u8] {
-        &mut self.payload[..self.size]
+    fn is_empty(&self) -> bool {
+        self.metadata_buffer.is_empty() || self.payload_buffer.is_empty()
```
whitequark: I don't think this function has the correct implementation. Shouldn't it be a logical AND? A buffer with a single zero-sized packet is not empty.

src/socket/udp.rs
```diff
@@ -1,58 +1,74 @@
...
-    fn resize<'b>(&'b mut self, size: usize) -> Result<&'b mut Self> {
-        if self.payload.len() >= size {
-            self.size = size;
-            Ok(self)
+    fn check_capacity(&self, required_size: usize) -> Result<()> {
+        if self.metadata_buffer.is_full() || self.payload_buffer.window() < required_size {
+            Err(Error::Exhausted)
         } else {
-            Err(Error::Truncated)
+                Ok(())
```
whitequark: Indentation.

src/socket/udp.rs
```diff
@@ -1,58 +1,74 @@
...
-#[derive(Debug)]
-pub struct PacketBuffer<'a> {
+// Endpoint and size of an UDP packet.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct PacketMetadata {
     endpoint: IpEndpoint,
-    size:     usize,
-    payload:  Managed<'a, [u8]>
+    payload_size: usize,
+    /// Dummy packets can be used to avoid wrap-arounds of packets in the payload buffer
```
whitequark: I would prefer a self-explanatory `padding` instead of non-informative `dummy`.

src/socket/udp.rs
```diff
@@ -1,58 +1,74 @@
...
         }
     }
-}
 
-impl<'a> Resettable for PacketBuffer<'a> {
-    fn reset(&mut self) {
-        self.endpoint = Default::default();
-        self.size = 0;
+    fn prepare_for_insert(&mut self, required_size: usize) -> Result<()> {
+        self.check_capacity(required_size)?;
```
whitequark: The control flow here is confusing. I think you can inline one `check_capacity` invocation, remove the second, and rewrite the condition to something like...
```
let window = self.payload_buffer.window();
let contig_window = self.payload_buffer.contiguous_window();
if window < size || window - contig_window < size { return Err(Error::Exhausted) }
```
etc.

src/socket/udp.rs
```diff
@@ -1,58 +1,74 @@
...
-/// A buffered UDP packet.
-#[derive(Debug)]
-pub struct PacketBuffer<'a> {
+// Endpoint and size of an UDP packet.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct PacketMetadata {
     endpoint: IpEndpoint,
-    size:     usize,
-    payload:  Managed<'a, [u8]>
+    payload_size: usize,
```
whitequark: Just `size`, it's implied that payloads are enqueued.

src/socket/udp.rs
whitequark: Also, I suggest calling this something like `enqueue`, moving the non-padding packet enqueueing code in, adding the `enqueue_many` call at the end, and making it return a `Result<&mut [u8]>` with the payload slice, this will reduce duplication below.

src/socket/udp.rs
```diff
@@ -155,19 +171,25 @@ impl<'a, 'b> UdpSocket<'a, 'b> {
...
     pub fn send(&mut self, size: usize, endpoint: IpEndpoint) -> Result<&mut [u8]> {
         if self.endpoint.port == 0 { return Err(Error::Unaddressable) }
         if !endpoint.is_specified() { return Err(Error::Unaddressable) }
 
-        let packet_buf = self.tx_buffer.enqueue_one_with(|buf| buf.resize(size))?;
-        packet_buf.endpoint = endpoint;
+        self.tx_buffer.prepare_for_insert(size)?;
+
+        let payload_buf = self.tx_buffer.payload_buffer.enqueue_many(size);
+        assert_eq!(payload_buf.len(), size);
```
whitequark: This should be a `debug_assert!` too.

m-labs-homu:
:umbrella: The latest upstream changes (presumably c418b60b5db93753999ae51d33ec4dc1d1631c69) made this pull request unmergeable. Please resolve the merge conflicts.

whitequark:
@phil-opp ping? I'd like to merge this.

phil-opp:
@whitequark I won't have time to work on this until Wednesday. Sorry for the delay!

phil-opp:
@whitequark I merged @dlrobertson 's improvements and rebased on master. I also squashed my commits. I left the new commits from @dlrobertson unsquashed to make the review easier and to keep the author information.

whitequark:
Thanks, this should be sufficient. Incidentally, you can make [commits with multiple authors](https://help.github.com/articles/creating-a-commit-with-multiple-authors/) now.

whitequark:
On reflection, I'm really not happy about this code. It's still too complex (there should have been a `dequeue` function as well), too coupled to the UDP socket (this is especially visible in the tests), and there is a bug where enqueueing a packet that's smaller than the window but larger than contiguous window will enqueue padding before bailing out, thus wasting space.

whitequark:
86c59fe is a *far* nicer implementation.

PR #141, opened 2018-02-01
Use the time types instead of a u64

dlrobertson:
Use the [Instant] and [Duration] types defined in #110

This isn't nearly done. I simply use `total_millis` where we should probably continue to pass the time type along. I'd like to get feedback and make sure I'm going the right direction before making all of the changes.

[Instant]: https://github.com/m-labs/smoltcp/blob/master/src/time.rs#L29
[Duration]: https://github.com/m-labs/smoltcp/blob/master/src/time.rs#L102

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -16,7 +16,7 @@ use smoltcp::iface::{NeighborCache, EthernetInterfaceBuilder};
 use smoltcp::socket::SocketSet;
 use smoltcp::socket::{UdpSocket, UdpSocketBuffer, UdpPacketBuffer};
 use smoltcp::socket::{TcpSocket, TcpSocketBuffer};
-use smoltcp::time::Instant;
+use smoltcp::time::{Duration, Instant};
 
 fn main() {
     utils::setup_logging("");
@@ -152,8 +152,8 @@ fn main() {
             let mut socket = sockets.get::<TcpSocket>(tcp3_handle);
             if !socket.is_open() {
                 socket.listen(6971).unwrap();
-                socket.set_keep_alive(Some(1000));
-                socket.set_timeout(Some(2000));
+                socket.set_keep_alive(Some(Duration::from_millis(1000)));
+                socket.set_timeout(Some(Duration::from_millis(2000)));
             }
 
             if socket.may_recv() {
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -342,7 +342,7 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
             let socket_poll_at = socket.poll_at();
             socket.meta().poll_at(socket_poll_at, |ip_addr|
                 self.inner.has_neighbor(&ip_addr, timestamp))
-        }).min().map(|x| Instant::from_millis(x as i64))
+        }).min()
     }
 
     /// Return an _advisory wait time_ for calling [poll] the next time.
@@ -438,7 +438,7 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
                             respond!(Packet::Udp(response))),
                     #[cfg(feature = "socket-tcp")]
                     Socket::Tcp(ref mut socket) =>
-                        socket.dispatch(timestamp.total_millis() as u64, &caps, |response|
+                        socket.dispatch(timestamp, &caps, |response|
                             respond!(Packet::Tcp(response))),
                     Socket::__Nonexhaustive(_) => unreachable!()
                 };
@@ -451,7 +451,7 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
                     // requests from the socket. However, without an additional rate limiting
                     // mechanism, we would spin on every socket that has yet to discover its
                     // neighboor.
-                    socket.meta_mut().neighbor_missing(timestamp.total_millis() as u64,
+                    socket.meta_mut().neighbor_missing(timestamp,
                         neighbor_addr.expect("non-IP response packet"));
                     break
                 }
@@ -625,7 +625,7 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
 
             #[cfg(feature = "socket-tcp")]
             IpProtocol::Tcp =>
-                self.process_tcp(sockets, timestamp.total_millis() as u64, ip_repr, ip_payload),
+                self.process_tcp(sockets, timestamp, ip_repr, ip_payload),
 
             #[cfg(feature = "socket-raw")]
             _ if handled_by_raw_socket =>
@@ -692,7 +692,7 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
 
             #[cfg(feature = "socket-tcp")]
             IpProtocol::Tcp =>
-                self.process_tcp(sockets, timestamp.total_millis() as u64, ip_repr, ip_payload),
+                self.process_tcp(sockets, timestamp, ip_repr, ip_payload),
 
             #[cfg(feature = "socket-raw")]
             _ if handled_by_raw_socket =>
@@ -887,7 +887,7 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
     }
 
     #[cfg(feature = "socket-tcp")]
-    fn process_tcp<'frame>(&self, sockets: &mut SocketSet, timestamp: u64,
+    fn process_tcp<'frame>(&self, sockets: &mut SocketSet, timestamp: Instant,
                            ip_repr: IpRepr, ip_payload: &'frame [u8]) ->
                           Result<Packet<'frame>>
     {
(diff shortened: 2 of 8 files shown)


src/time.rs
```diff
@@ -62,6 +67,13 @@ impl Instant {
     pub fn total_millis(&self) -> i64 {
         self.millis
     }
+
+    /// Return the [Duration] between two [Instant's].
+    ///
+    /// [Duration] struct.Duration.html
+    pub fn duration_since(&self, other: Instant) -> Duration {
```
whitequark: Shouldn't this be `impl Sub<Instant> for Instant`?

src/time.rs
```diff
@@ -85,6 +97,12 @@ impl ops::Add<Duration> for Instant {
     }
 }
 
+impl ops::AddAssign<Duration> for Instant {
```
whitequark: I think these should go in a separate commit.

src/time.rs
dlrobertson: 👍

src/time.rs
dlrobertson: :+1:

dlrobertson:
@whitequark would you like all the changes to the TCP timestamps etc to be in this PR too? If so, I'm assuming they should be in a separate commit

whitequark:
@dlrobertson Sure, separate commit seems fine.

examples/benchmark.rs
```diff
@@ -31,7 +31,7 @@ fn client(kind: Client) {
     let mut stream = TcpStream::connect(("192.168.69.1", port)).unwrap();
     let mut buffer = vec![0; 1_000_000];
 
-    let start = Instant::now();
+    let start = Instant::now().expect("Could not get the start time");
```
whitequark: Hang on. Why would this fail?

examples/utils.rs
```diff
@@ -18,14 +18,15 @@ use smoltcp::phy::{Device, EthernetTracer, FaultInjector};
...
 
 #[cfg(feature = "log")]
 pub fn setup_logging_with_clock<F>(filter: &str, since_startup: F)
-        where F: Fn() -> u64 + Send + Sync + 'static {
+        where F: Fn() -> Instant + Send + Sync + 'static {
     LogBuilder::new()
         .format(move |record: &LogRecord| {
             let elapsed = since_startup();
-            let timestamp = format!("[{:6}.{:03}s]", elapsed / 1000, elapsed % 1000);
+            let timestamp = format!("{}", elapsed);
```
whitequark: I think you forgot `[]`.

examples/benchmark.rs
dlrobertson: It makes use of [SystemTime::duration_since] which returns a result type. Now that I'm looking at it, would it make more sense if `Instant::from_system_time` used [SystemTIme::elapsed] instead of [SystemTime::duration_since] the [UNIX_EPOCH]? Or would it make sense to have a `Instant::elapsed()`?

[SystemTime::duration_since]: https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.duration_since
[SystemTime::elapsed]: https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.elapsed
[UNIX_EPOCH]: https://doc.rust-lang.org/std/time/constant.UNIX_EPOCH.html

whitequark: It should just use `duration_since(UNIX_EPOCH).unwrap()`. I am not interested in handling cases where the time is less than UNIX epoch. The only such case is badly misconfigured (i.e. non-UTC timezone) systems without an RTC battery or ntpd.

whitequark: Actually, nevermind, not even that can happen because UNIX_EPOCH includes the timezone. So just unwrap it.

dlrobertson: :+1:

dlrobertson:
Updated all the things!

Added commits that update the socket modules to use `time` etc.

src/time.rs
```diff
@@ -23,7 +23,7 @@ use core::{ops, fmt};
 ///   point.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Instant {
-    millis: i64,
+    pub millis: i64,
```
dlrobertson: Updated to `pub` so that constants can be used.

examples/benchmark.rs
```diff
@@ -52,11 +52,9 @@ fn client(kind: Client) {
 
     let end = Instant::now();
 
-    let elapsed = end - start;
-    let elapsed = elapsed.as_secs() as f64
-                + elapsed.subsec_nanos() as f64 * 1e-9;
+    let elapsed = (end - start).total_millis() as f64 / 1000.0;
 
-    println!("throughput: {:.3} Gbps", AMOUNT as f64 / elapsed / 0.125e9);
+    println!("throughput: [{:.3}] Gbps", AMOUNT as f64 / elapsed / 0.125e9);
```
whitequark: huh?

examples/loopback.rs
```diff
@@ -57,12 +60,12 @@ mod mock {
...
-        pub fn advance(&self, millis: u64) {
-            self.0.fetch_add(millis as usize, Ordering::SeqCst);
+        pub fn advance(&self, duration: Duration) {
+            self.0.fetch_add(duration.total_millis() as usize, Ordering::SeqCst);
         }
 
-        pub fn elapsed(&self) -> u64 {
-            self.0.load(Ordering::SeqCst) as u64
+        pub fn elapsed(&self) -> Instant {
+			Instant::from_millis(self.0.load(Ordering::SeqCst) as i64)
```
whitequark: something weird happened with indentation

examples/loopback.rs
```diff
@@ -164,13 +167,15 @@ fn main() {
         }
 
         match iface.poll_delay(&socket_set, clock.elapsed()) {
-            Some(0) =>
```
whitequark: now that `millis` is `pub`, I think you can match directly

examples/benchmark.rs
dlrobertson: updated the wrong line. fixed.

examples/loopback.rs
dlrobertson: That is strange... somehow I inserted a tab here. Not sure how that happened.

examples/loopback.rs
dlrobertson: Good point.

dlrobertson:
A few things before this should be merged.

 - `From<::std::time::Instant>` isn't implemented. Adding this now
 - `from_system_time` should be `From<::std::time::SystemTime>` now that it isn't a result type.
 - Question: Should `Instant::now()` be a "wrapper" for `::std::time::Instant::now()` or what is currently done with the `UNIX_EPOCH`?

dlrobertson:
Updated and I didn't change `Instant::now`. I think sticking with the current implementation makes more sense.

src/phy/fault_injector.rs
```diff
@@ -17,21 +18,35 @@ fn xorshift32(state: &mut u32) -> u32 {
...
     drop_pct:    u8,
     reorder_pct: u8,
     max_size:    usize,
     max_tx_rate: u64,
     max_rx_rate: u64,
-    interval:    u64,
+    interval:    Duration,
+}
+
+impl Default for Config {
```
whitequark: Why not `impl Default for Duration`? `std` does this.

src/time.rs
```diff
@@ -32,17 +32,15 @@ impl Instant {
...
-    #[cfg(feature = "std")]
-    pub fn now() -> Result<Instant, ::std::time::SystemTimeError> {
-        Self::from_system_time(::std::time::SystemTime::now())
+    /// Create a new `Instant` from a number of seconds.
+    pub fn from_secs(secs: i64) -> Instant {
+        Instant { millis: secs * 1000 }
     }
 
-    /// Create a new `Instant` from a `SystemTime`.
+    /// Create a new `Instant` from the current `SystemTime`.
```
whitequark: Nit: it's not clear from the signature and doc comment that this means `std::time::SystemTime`.

whitequark:
@m-labs-homu delegate+
@dlrobertson r=me once the comments above are fixed

dlrobertson:
@m-labs-homu r=whitequark

m-labs-homu:
:pushpin: Commit cbedd74 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit cbedd743db9bef88e4e0dd8e15bf042fa38c0ae4 with merge c418b60b5db93753999ae51d33ec4dc1d1631c69...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/342175985?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing c418b60b5db93753999ae51d33ec4dc1d1631c69 to master...

whitequark:
Thanks! Great work.

PR #143, opened 2018-02-02
Imp `From` trait on EthernetAddress

LuoZijun:


whitequark:
Why?

LuoZijun:
@whitequark 

just for friendly api.

and i also add `Hash` trait on `EthernetAddress` .

whitequark:
So what's "friendly" about it? I don't see the use case.

LuoZijun:
@whitequark 

```rust
let hwaddr = EthernetAddress::from([1,2,3,44,5,6]);

```

whitequark:
> let hwaddr = EthernetAddress::from([1,2,3,44,5,6]);

This is longer than using the constructor directly:

    let hwaddr = EthernetAddress([1,2,3,4,5,6]);

LuoZijun:
yeah, got it.

now just add `Hash` trait.

PR #145, opened 2018-02-04
remove linux only feature

LuoZijun:
remove that default features is good for a non-linux user.

### DIFF
--- a/examples/utils.rs
+++ b/examples/utils.rs
@@ -82,6 +82,7 @@ pub fn add_tap_options(_opts: &mut Options, free: &mut Vec<&str>) {
     free.push("INTERFACE");
 }
 
+#[cfg(feature = "phy-tap_interface")]
 pub fn parse_tap_options(matches: &mut Matches) -> TapInterface {
     let interface = matches.free.remove(0);
     TapInterface::new(&interface).unwrap()


whitequark:
I'm not so sure. On Windows we don't support either tap interfaces or raw sockets (and probably never will), so should remove both of these from default features forever? The primary platform for developing smoltcp is Linux, after all...

LuoZijun:
just remove `phy-tab_interface` from default features  forever.

in future, the `phy-raw_socket`  feature will available on windows/macos/linux,
i belive that. so at that moment, i think `phy-raw_socket` can back to default features.

does removing these features cause inconvenience to Linux users ?

i see every example section has mark required-features.
```
[[example]]
name = "httpclient"
required-features = ["std", "phy-tap_interface", "proto-ipv4", "socket-tcp"]
```

whitequark:
Yeah, this e.g breaks README:
```
$ cargo run --example server tap0
error: target `server` requires the features: `std`, `phy-tap_interface`, `proto-ipv4`, `socket-tcp`, `socket-udp`
Consider enabling them by passing e.g. `--features="std phy-tap_interface proto-ipv4 socket-tcp socket-udp"`
```
So I am not opposed to this PR but only if you update the README.

examples/utils.rs
```diff
@@ -15,7 +15,7 @@ use env_logger::LogBuilder;
 use getopts::{Options, Matches};
 
 use smoltcp::phy::{Device, EthernetTracer, FaultInjector};
-#[cfg(feature = "phy-tap_interface")]
+#[cfg(all(feature = "phy-tap_interface", target_os = "linux"))]
```
whitequark: I don't think `target_os = "linux"` is necessary.

LuoZijun:
ok, i did not know that problem.

i always think if u add `phy-tap_interface` to `required-features` array , that's wil be fine for linux user.

now i just add one line about examples/util.rs .

Issue #144, opened 2018-02-03
Add function `netmask()` and `broadcast` on Ipv4Cidr/Ipv6Cidr ?


LuoZijun:
Add function `netmask()` and `broadcast()` on Ipv4Cidr/Ipv6Cidr ?

```rust

impl Ipv4Cidr {
    pub fn netmask(&self) -> Address {
        let mut addr = [0u8; 4];
        if self.prefix_len >= 8 {
            for idx in 0..((self.prefix_len - self.prefix_len % 8) / 8) {
                addr[idx] = 255;
            }
        }

        Address(addr)
    }

    pub fn broadcast(&self) -> Address {
        let mut addr = [0u8; 4];
        let mask = self.netmask;
        for idx in 0..4 {
            if mask[idx] == 255 {
                addr[idx] = self.address.0[0];
            } else {
                addr[idx] = 255;
            }
        }
        Address(addr)
    }
}
```

whitequark:
Only on IPv4Cidr, IPv6 does not have netmasks and it only has a single broadcast address, `::`.

LuoZijun:
See (PR-146)[https://github.com/m-labs/smoltcp/pull/146]

PR #146, opened 2018-02-04
Add `netmask()/broadcast()/network()` method on IPv4Cidr #144

LuoZijun:
Add method:

*    netmask()
*    broadcast()
*    network()

### DIFF
--- a/src/wire/ipv4.rs
+++ b/src/wire/ipv4.rs
@@ -85,6 +85,20 @@ impl Address {
     }
 }
 
+#[cfg(feature = "std")]
+impl From<::std::net::Ipv4Addr> for Address {
+    fn from(x: ::std::net::Ipv4Addr) -> Address {
+        Address(x.octets())
+    }
+}
+
+#[cfg(feature = "std")]
+impl From<Address> for ::std::net::Ipv4Addr {
+    fn from(Address(x): Address) -> ::std::net::Ipv4Addr {
+        x.into()
+    }
+}
+
 impl fmt::Display for Address {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let bytes = self.0;
@@ -205,7 +219,7 @@ impl fmt::Display for Cidr {
 }
 
 /// A read/write wrapper around an Internet Protocol version 4 packet buffer.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct Packet<T: AsRef<[u8]>> {
     buffer: T
 }
--- a/src/wire/arp.rs
+++ b/src/wire/arp.rs
@@ -21,7 +21,7 @@ enum_with_unknown! {
 }
 
 /// A read/write wrapper around an Address Resolution Protocol packet buffer.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct Packet<T: AsRef<[u8]>> {
     buffer: T
 }
--- a/src/wire/icmpv4.rs
+++ b/src/wire/icmpv4.rs
@@ -166,7 +166,7 @@ enum_with_unknown! {
 }
 
 /// A read/write wrapper around an Internet Control Message Protocol version 4 packet buffer.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct Packet<T: AsRef<[u8]>> {
     buffer: T
 }
--- a/src/wire/icmpv6.rs
+++ b/src/wire/icmpv6.rs
@@ -94,7 +94,7 @@ impl fmt::Display for DstUnreachable {
 }
 
 /// A read/write wrapper around an Internet Control Message Protocol version 6 packet buffer.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct Packet<T: AsRef<[u8]>> {
     buffer: T
 }
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -160,6 +160,16 @@ impl Address {
     }
 }
 
+#[cfg(all(feature = "std", feature = "proto-ipv4", feature = "proto-ipv6"))]
+impl From<::std::net::IpAddr> for Address {
+    fn from(x: ::std::net::IpAddr) -> Address {
+        match x {
+            ::std::net::IpAddr::V4(ipv4) => Address::Ipv4(ipv4.into()),
+            ::std::net::IpAddr::V6(ipv6) => Address::Ipv6(ipv6.into()),
+        }
+    }
+}
+
 impl Default for Address {
     fn default() -> Address {
         Address::Unspecified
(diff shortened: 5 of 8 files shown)


src/wire/ipv4.rs
```diff
@@ -120,6 +120,45 @@ impl Cidr {
...
+                addr[idx] = self.address.0[idx];
+            } else {
+                addr[idx] = 255u8;
+            }
+        }
+        Address(addr)
+    }
+
+    /// Return the network CIDR of this IPv4 CIDR block.
+    pub fn network(&self) -> Cidr {
```
whitequark: This method doesn't make any sense to me. What is "network CIDR"? `self` is already a CIDR.

src/wire/ipv4.rs
```diff
@@ -120,6 +120,45 @@ impl Cidr {
         self.prefix_len
     }
 
+    /// Return the netmask address of this IPv4 CIDR block.
```
whitequark: It's not "netmask address", it's "network mask".

src/wire/ipv4.rs
```diff
@@ -120,6 +120,45 @@ impl Cidr {
...
+    pub fn netmask(&self) -> Address {
+        let mut addr = [0u8; 4];
+        if self.prefix_len >= 8 {
+            for idx in 0usize..((self.prefix_len - self.prefix_len % 8) / 8) as usize {
```
whitequark: Your computations are incorrect for prefix lengths that are not a multiple of 8. There is no requirement that prefix length be a multiple of 8.

src/wire/ipv4.rs
```diff
@@ -814,5 +853,36 @@ mod test {
 
         let cidr_without_prefix = Cidr::new(cidr.address(), 0);
         assert!(cidr_without_prefix.contains_addr(&Address::new(127, 0, 0, 1)));
+
+        // test_cidr_netmask
```
whitequark: These should be separate methods.

src/wire/ipv4.rs
LuoZijun: CIDR not eq network.
```
cirdr = 192.168.1.100/24
network = 192.168.1.0/24
```

src/wire/ipv4.rs
LuoZijun: just do XOR ?

src/wire/ipv4.rs
whitequark: OK, the documentation is confusing. How about:

    /// Return the network block of this IPv4 CIDR.

src/wire/ipv4.rs
```diff
@@ -814,5 +853,36 @@ mod test {
...
+        assert_eq!(Cidr::new(Address([127, 0, 0, 0]), 8).netmask(),
+                   Address([255, 0, 0, 0]));
+        assert_eq!(Cidr::new(Address([192, 168, 0, 0]), 16).netmask(),
+                   Address([255, 255, 0, 0]));
+        assert_eq!(Cidr::new(Address([192, 168, 1, 1]), 16).netmask(),
+                   Address([255, 255, 0, 0]));
+        assert_eq!(Cidr::new(Address([192, 168, 1, 1]), 17).netmask(),
+                   Address([255, 255, 0, 0]));
```
whitequark: This test is incorrect. It should be:

```
assert_eq!(Cidr::new(Address([192, 168, 1, 1]), 17).netmask(),
                     Address([255, 255, 128, 0]));
```

src/wire/ipv4.rs
whitequark: I don't see how this helps. I've commented below on a test to show how this method should work.

src/wire/ipv4.rs
LuoZijun: ok

src/wire/ipv4.rs
LuoZijun: now should be correct。

src/wire/ipv4.rs
LuoZijun: now should be correct。

LuoZijun:
does need a method `with_netmask()` ?

```
imp Cidr {
    
    pub fn with_netmask(addr: Address, netmask: Address) -> Cidr {
        let prefix_len = netmask.0[0].count_ones()
                         + netmask.0[1].count_ones() 
                         + netmask.0[2].count_ones()
                         + netmask.0[3].count_ones();
        Cidr { address: addr, prefix_len: prefix_len }
    }
}
```

whitequark:
To be honest, I don't know why you even bother with netmasks in the first place. Netmasks are an outdated concept and they shouldn't really be used today.

LuoZijun:
in real world, when you doing `unix network programming` , the `netmask` always follow you.

that's so sad.

LuoZijun:
so, if you have no objection, I will add ' with_netmask ' method.

that's very useful for system network programming.

src/wire/ipv4.rs
```diff
@@ -120,6 +120,45 @@ impl Cidr {
         self.prefix_len
     }
 
+    /// Return the network mask of this IPv4 CIDR block.
+    pub fn netmask(&self) -> Address {
+        let number = 0xffffffffu32 << (32 - self.prefix_len);
```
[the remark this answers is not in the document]
LuoZijun: ok, i will add more test.

src/wire/ipv4.rs
```diff
@@ -120,6 +120,45 @@ impl Cidr {
...
+        ];
+        Address(data)
+    }
+
+    /// Return the broadcast address of this IPv4 CIDR block.
+    pub fn broadcast(&self) -> Address {
+        let network = self.network();
+        let network_address = network.address.0;
+        let number = NetworkEndian::read_u32(&network_address[..]) 
+                    | (0xffffffffu32 << (32 - network.prefix_len) ^ 0xffffffff);
```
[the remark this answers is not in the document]
LuoZijun: 👍

src/wire/ipv4.rs
```diff
@@ -110,6 +110,19 @@ impl Cidr {
...
 
+    /// Create an IPv4 CIDR block from the given address and network mask.
+    ///
+    /// # Panics
+    /// This function panics if the prefix length is larger than 32.
+    pub fn with_netmask(addr: Address, netmask: Address) -> Cidr {
+        let prefix_len = netmask.0[0].count_ones()
+                         + netmask.0[1].count_ones() 
+                         + netmask.0[2].count_ones()
+                         + netmask.0[3].count_ones();
```
whitequark: I think this function must check that the ones in netmask are all consecutive and left-aligned, and return a `Result`.

src/wire/ipv4.rs
```diff
@@ -110,6 +110,19 @@ impl Cidr {
...
+        assert!(prefix_len <= 32);
```
whitequark: This assertion follows trivially from the line above, so it's not necessary.

src/wire/ipv4.rs
```diff
@@ -120,6 +133,53 @@ impl Cidr {
...
+            ((number >>  0) & 0xff) as u8,
+        ];
+        Address(data)
+    }
+
+    /// Return the broadcast address of this IPv4 CIDR.
+    pub fn broadcast(&self) -> Address {
+        let network = self.network();
+
+        if network.prefix_len == 32 {
```
whitequark: Actually, this is not right. A /31 subnet or /32 "subnet" [does not have a broadcast address](http://packetlife.net/blog/2008/jun/18/using-31-bit-subnets-on-point-point-links/). So, this function should return an `Option<Address>`.

src/wire/ipv4.rs
```diff
@@ -110,6 +110,19 @@ impl Cidr {
         Cidr { address, prefix_len }
     }
 
+    /// Create an IPv4 CIDR block from the given address and network mask.
+    ///
+    /// # Panics
+    /// This function panics if the prefix length is larger than 32.
```
whitequark: This function never panics (as written).

src/wire/ipv4.rs
LuoZijun: wow, you are very professional.

src/wire/ipv4.rs
LuoZijun: 👍

PR #149, opened 2018-02-05
Add conversions to/from std wire types

### LINKED ISSUE #147: Use standard library types where possible
I've noticed smoltcp defines its own `Ipv4Address` and `IpEndpoint` types instead of using the standard `Ipv4Addr` and `SocketAddr` types. Is there any good reason for this? It's kind of annoying to have to convert between the two.

canndrew:
Fixes #147

### DIFF
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -160,6 +160,16 @@ impl Address {
     }
 }
 
+#[cfg(all(feature = "std", feature = "proto-ipv4", feature = "proto-ipv6"))]
+impl From<::std::net::IpAddr> for Address {
+    fn from(x: ::std::net::IpAddr) -> Address {
+        match x {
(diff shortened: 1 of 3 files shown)


whitequark:
This broke CI.

canndrew:
CI passes now.

src/wire/ip.rs
```diff
@@ -1,5 +1,7 @@
 use core::fmt;
 use core::convert::From;
+#[cfg(all(feature = "std", feature = "proto-ipv4", feature = "proto-ipv6"))]
```
whitequark: I'd prefer if we had no conditional `use` statements.

canndrew: I've changed it to use global paths instead. eg. `::std::net::Ipv4Addr`.

whitequark:
Thanks!

PR #161, opened 2018-02-12
Small doc updates

dlrobertson:
- Update documentation about current support in the wire module
 - Ensure the possible panic is documented for `Ipv6Option::data_mut`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,8 +49,7 @@
 //! The wire layer APIs also provide _tcpdump_-like pretty printing.
 //!
 //! ## The representation layer
-//! The representation layer APIs are provided in the module [wire](wire/index.html); currently,
-//! Ethernet, ARP, generic IP, IPv4, ICMPv4, TCP and UDP packet representations are provided.
+//! The representation layer APIs are provided in the module [wire].
 //!
 //! The representation layer exists to reduce the state space of raw packets. Raw packets
 //! may be nonsensical in a multitude of ways: invalid checksums, impossible combinations of flags,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -49,24 +49,35 @@
...
 //! The packet layer exists to provide a more structured way to work with packets than
 //! treating them as sequences of octets. It makes no judgement as to content of the packets,
 //! except where necessary to provide safe access to fields, and strives to implement every
 //! feature ever defined, to ensure that, when the representation layer is unable to make sense
 //! of a packet, it is still logged correctly and in full.
 //!
+//! ## Packet and representation layer support
+//! The packet and representation layer currently support the following packet representations:
+//!
+//!  - Ethernet
```
whitequark: There's no EthernetRepr.

dlrobertson: Ah, good catch added a note here.

dlrobertson: Would a table be better here? Something like the following:

| Protocol | Packet | Representation |
|----------|--------|----------------|
| Ethernet | Yes    | No             |
| ARP      | Yes    | Yes            |
| IPv4     | Yes    | Yes            |
| ICMPv4   | Yes    | Yes            |
| IPv6     | Yes    | Yes            |
| ICMPv6   | Yes    | Yes            |
| TCP      | Yes    | Yes            |
| UDP      | Yes    | Yes            |

whitequark:
Implementing EthernetRepr would be better :)

dlrobertson:
> Implementing EthernetRepr would be better :)

Ah, didn't know if there was a reason behind the missing implementation. Did my best to take a crack at implementing it.

dlrobertson:
Weird... I guess I didn't push my changes that included the EthernetRepr

PR #165, opened 2018-02-15
Add `PrettyPrinter::print` convenience method

canndrew:
This is a much more convenient way of printing data during debugging because it saves me from having to reiterate the type of the packet. eg.

```rust
let frame = EthernetFrame::new(buffer);
trace!("{}", PrettyPrinter::print(&frame));
```

vs

```rust
let frame = EthernetFrame::new(buffer);
trace!("{}", PrettyPrinter::<EthernetFrame<&[u8]>>::new("", buffer));
```

The time saved quickly adds up when I'm trying to spray `trace` statements all over a piece of broken code.

Alternatively, or in addition, it might be a good idea to make the `Debug` impls pretty print like this. That way I could simply write `trace!("{:?}", frame)`. Thoughts?

### DIFF
--- a/src/wire/pretty_print.rs
+++ b/src/wire/pretty_print.rs
@@ -93,6 +93,17 @@ impl<'a, T: PrettyPrint> PrettyPrinter<'a, T> {
     }
 }
 
+impl<'a, T: PrettyPrint + AsRef<[u8]>> PrettyPrinter<'a, T> {
+    /// Create a `PrettyPrinter` which prints the given object.
+    pub fn print(printable: &'a T) -> PrettyPrinter<'a, T> {
+        PrettyPrinter {
+            prefix: "",
+            buffer: printable,
+            phantom: PhantomData,
+        }
+    }
+}
+
 impl<'a, T: PrettyPrint> fmt::Display for PrettyPrinter<'a, T> {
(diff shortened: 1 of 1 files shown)


whitequark:
> Alternatively, or in addition, it might be a good idea to make the Debug impls pretty print like this. That way I could simply write trace!("{:?}", frame). Thoughts?

If you have a packet inside some other structure, then this will wreck the flow, and the resulting text will not be valid Rust code, which is why I haven't already done that.

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 5745db4 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 5745db46873bbc6fbfc1084fa4da25e9279b0e60 with merge f1e5c73caec08b10a062392c1600ba4ccfd8d7da...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/341954748?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing f1e5c73caec08b10a062392c1600ba4ccfd8d7da to master...

Issue #159, opened 2018-02-08
Design: Implementing additional ICMP types


dlrobertson:
## Sumary

In order to implement [MLDv2], [NDISC], etc we need to implement additional ICMPv6 types (types that are not defined in [RFC 4443]. There are several reasonable paths forward, each with pros and cons. My plan it to go with Path 1. If there are other ideas or thoughts, please let me know.

## Path1

Create a new module in `wire` for each new protocol with the protocol specific parsing. New variants will need to be added to `Icmpv6Message` along with a generic variant of `Icmpv6Repr` for any message associated with the new protocol. The benefits of this approach include.

  - It will be much more ergonomic to include a separate `process_ndisc` function in `iface`.
  - There will not be as much clutter in the `icmpv6` module.

The drawbacks include

  - The API may look a little different than other protocols in `wire`, due to the message type identifier existing in the ICMPv6 message. This may cause `Packet::new(bytes)` in the new module to be `Packet::new(msg_type, bytes)`.
  - May be less efficient. Depending on the implementation, we might match on `msg_type` more than once to parse the message. Once to identify that it belongs to a "subprotocol", then again to parse the packet according to its type.

## Path 2

The easy path forward would just be to tack on additional variants to the `Icmpv6Repr` enum and add the necessary setters and getters to the `Icmpv6Packet` type. This is by far the simplest of the two, however, it introduces two key drawbacks.

  - There will be some setters and getters for `Icmpv6Packet` that would access fields beyond `HEADER_END`.
  - Eventually there would be a huge number of getters and setters for `Icmpv6Packet` and variants of `Icmpv6Repr`.

[RFC 4443]: https://tools.ietf.org/rfc/rfc4443
[NDISC]: https://tools.ietf.org/rfc/rfc4861
[MLDv2]: https://tools.ietf.org/rfc/rfc3810

whitequark:
Definitely path 2, I've already thought about it for ICMPv4 and decided on that. Did you know that you can add documentation to `impl` blocks? So you can have a separate `impl` block per message type.

dlrobertson:
> Definitely path 2

It is definitely a reasonable path, especially if the target user is a more advanced user. I'm nervous about users using NDISC or MLDv2 specific packet functions on packets that are not in the family. That is the main reason why I was leaning towards Path 1. But I guess such users probably will not be using the setters and getters in the `Packet` API?

> Did you know that you can add documentation to `impl` blocks?

[...]

PR #167, opened 2018-02-17
Add NDISC Option Type to wire

dlrobertson:
Add support for NDISC options parsing in wire.

See: https://tools.ietf.org/html/rfc4861#section-4.6

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ managed = { version = "0.5", default-features = false, features = ["map"] }
 byteorder = { version = "1.0", default-features = false }
 log = { version = "0.3", default-features = false, optional = true }
 libc = { version = "0.2.18", optional = true }
+bitflags = { version = "1.0", default-features = false }
 
 [dev-dependencies]
 env_logger = "0.4"
--- a/src/wire/mod.rs
+++ b/src/wire/mod.rs
@@ -97,6 +97,8 @@ mod icmpv4;
 mod icmpv6;
 #[cfg(feature = "proto-ipv4")]
 mod igmp;
+#[cfg(feature = "proto-ipv6")]
+mod ndiscoption;
 mod udp;
 mod tcp;
 #[cfg(feature = "proto-ipv4")]
@@ -170,6 +172,12 @@ pub use self::icmpv6::{Message as Icmpv6Message,
                        ParamProblem as Icmpv6ParamProblem,
                        Packet as Icmpv6Packet,
                        Repr as Icmpv6Repr};
+
+#[cfg(feature = "proto-ipv6")]
+pub use self::ndiscoption::{NdiscOption,
+                            Repr as NdiscOptionRepr,
+                            Type as NdiscOptionType,
+                            PrefixInfoFlags as NdiscPrefixInfoFlags};
 
 pub use self::udp::{Packet as UdpPacket,
                     Repr as UdpRepr};
--- a/src/wire/ndiscoption.rs
+++ b/src/wire/ndiscoption.rs
@@ -1,0 +1,704 @@
+use core::fmt;
+use byteorder::{NetworkEndian, ByteOrder};
+
+use {Error, Result};
+use time::Duration;
+use wire::{EthernetAddress, Ipv6Address, Ipv6Packet, Ipv6Repr};
+
+enum_with_unknown! {
+    /// NDISC Option Type
+    pub doc enum Type(u8) {
+        /// Source Link-layer Address
+        SourceLinkLayerAddr = 0x1,
+        /// Target Link-layer Address
+        TargetLinkLayerAddr = 0x2,
+        /// Prefix Information
+        PrefixInformation   = 0x3,
+        /// Redirected Header
+        RedirectedHeader    = 0x4,
+        /// MTU
+        Mtu                 = 0x5
+    }
+}
+
+impl fmt::Display for Type {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            &Type::SourceLinkLayerAddr => write!(f, "source link-layer address"),
+            &Type::TargetLinkLayerAddr => write!(f, "target link-layer address"),
+            &Type::PrefixInformation   => write!(f, "prefix information"),
+            &Type::RedirectedHeader    => write!(f, "redirected header"),
+            &Type::Mtu                 => write!(f, "mtu"),
+            &Type::Unknown(id) => write!(f, "{}", id)
+        }
+    }
+}
+
+bitflags! {
+    pub struct PrefixInfoFlags: u8 {
+        const ON_LINK  = 0b10000000;
+        const ADDRCONF = 0b01000000;
+    }
+}
+
+/// A read/write wrapper around an [NDISC Option].
+///
+/// [NDISC Option]: https://tools.ietf.org/html/rfc4861#section-4.6
+#[derive(Debug, PartialEq)]
+pub struct NdiscOption<T: AsRef<[u8]>> {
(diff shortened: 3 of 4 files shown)


src/wire/ndiscoption.rs
```diff
@@ -0,0 +1,703 @@
...
+            Err(Error::Truncated)
+        } else {
+            let data_range = field::DATA(data[field::LENGTH]);
+            if len < data_range.end {
+                Err(Error::Truncated)
+            } else {
+                match self.option_type() {
+                    Type::SourceLinkLayerAddr | Type::TargetLinkLayerAddr | Type::Mtu if data_range.end == 8 =>
+                        Ok(()),
+                    Type::PrefixInformation if data_range.end == 32 =>
```
dlrobertson: I was hesitant to include this, because we are beginning to test the packet is well formed if we use `==` here. We do need to do additional checks based on the type. Otherwise `check_len` will not return `Error::Truncated` for a option with `data[field::LENGTH] == 1` and `data[field::TYPE] == Type::PrefixInformation`, which could panic later if the prefix field were accessed.

src/wire/ndiscoption.rs
```diff
@@ -0,0 +1,703 @@
...
+// Format of an NDISC Option
+//
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |     Type      |    Length     |              ...              |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// ~                              ...                              ~
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+//
+// See https://tools.ietf.org/html/rfc4861#section-4.6 for details.
+mod field {
```
dlrobertson: Wasn't sure how to break up the groupings of constants. If this is not readable, I'm more than happy to iterate on this.

src/wire/ndiscoption.rs
```diff
@@ -0,0 +1,703 @@
...
+
+    /// Return the length of the data.
+    #[inline]
+    pub fn data_len(&self) -> u8 {
+        let data = self.buffer.as_ref();
+        data[field::LENGTH]
+    }
+}
+
+/// Getter methods only relevant for Source/Target Link-layer Address options.
```
dlrobertson: Used impl block docs :smile: As a whole I did not prefix type specific methods with the option type name. Let me know if this would be a good addition.

Cargo.toml
```diff
@@ -16,6 +16,7 @@ managed = { version = "0.5", default-features = false, features = ["map"] }
 byteorder = { version = "1.0", default-features = false }
 log = { version = "0.3", default-features = false, optional = true }
 libc = { version = "0.2.18", optional = true }
+bitflags = { version = "1.0", default-features = false, optional = true }
```
whitequark: I would prefer if we had `bitflags` non-optional for simplicity, since it has zero runtime cost.

src/wire/mod.rs
```diff
@@ -91,6 +91,8 @@ mod ipv6option;
 mod icmpv4;
 #[cfg(feature = "proto-ipv6")]
 mod icmpv6;
+#[cfg(feature = "proto-ipv6")]
+mod ndiscoption;
```
whitequark: Why not just `ndisc`? It doesn't matter that it's an option, the protocol is called Neighbor Discovery. Also, you're already dropping `Option` in type names like `NdiscPrefixInfoFlags`.

src/wire/ndiscoption.rs
whitequark: This seems fine to me, except that returning `Truncated` when the packet is *longer* than expected (i.e. `data_range.end > 32`) is really confusing.

Cargo.toml
dlrobertson: :+1:

src/wire/mod.rs
dlrobertson: True. This is just the Option parsing, but the actual NDISC parsing will be in `src/wire/icmpv6.rs` per the conversation for #159, so I'll rename it.

src/wire/ndiscoption.rs
dlrobertson: Yeah, what about updating the docs for `check_len` and returning `Malformed` here?

src/wire/mod.rs
whitequark: No, I misunderstood, if this file is just NDISC *option* parsing then this makes sense.

dlrobertson: :+1:

whitequark: We can actually pull out NDISC into its own `ndisc.rs` file if we mark `Icmpv6Packet::buffer` as `pub(crate)`.

src/wire/ndiscoption.rs
dlrobertson: Updated to *not* return `Truncated` when a packet is longer than expected.

dlrobertson:
Updated

src/wire/mod.rs
dlrobertson: So then `NdiscPacket::new` would be something like `NdiscPacket::new(packet: Icmpv6Packet)`? It gets a bit tricky since the type id is in the ICMP header.

whitequark: No, don't add an `NdiscPacket`, just move the `impl` blocks into a separate file.

dlrobertson:
@whitequark were there any outstanding review items that I missed. I'll add NDISC packet support in another PR.

m-labs-homu:
:umbrella: The latest upstream changes (presumably 53abe8c5202b6dda714a0f38e33faf3a49e1191d) made this pull request unmergeable. Please resolve the merge conflicts.

src/wire/ndiscoption.rs
```diff
@@ -0,0 +1,703 @@
...
+            Err(Error::Truncated)
+        } else {
+            let data_range = field::DATA(data[field::LENGTH]);
+            if len < data_range.end {
+                Err(Error::Truncated)
+            } else {
+                match self.option_type() {
+                    Type::SourceLinkLayerAddr | Type::TargetLinkLayerAddr | Type::Mtu =>
+                        Ok(()),
+                    Type::PrefixInformation if data_range.end > 31 =>
```
whitequark: Can you use constants from `field` here? It's not clear where 31 comes from.

whitequark:
Looks good other than the nit above. r=me after fixing that and rebasing.

dlrobertson:
@m-labs-homu r=whitequark

m-labs-homu:
@dlrobertson: :key: Insufficient privileges: Not in reviewers

dlrobertson:
@whitequark  :face_with_head_bandage: looks like I don't have the privileges for that.

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 999331e has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 999331efff8a197181da451cc3a494592a43e79b with merge 9b68b0866fcf79c1820db5a363166d26148ace37...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/352099633?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 9b68b0866fcf79c1820db5a363166d26148ace37 to master...

PR #176, opened 2018-03-05
IGMPv1/v2

astro:
This picks up the work of #60, trying to process all the feedback. I usually try to provide a cleanly rebased patchset but in this case the conflicts were too many. Apologies.

[ManagedMap.iter()](https://github.com/m-labs/rust-managed/pull/6) is required.

### DIFF
--- a/src/wire/igmpv2.rs
+++ b/src/wire/igmpv2.rs
@@ -1,15 +1,3 @@
-/// Internet Group Management Protocol v2
-/// defined in [RFC_2236]
-///
-///    0                   1                   2                   3
-///    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-///   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-///   |      Type     | Max Resp Time |           Checksum            |
-///   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-///   |                         Group Address                         |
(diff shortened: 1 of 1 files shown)


src/wire/igmpv2.rs
```diff
@@ -0,0 +1,420 @@
+/// Internet Group Management Protocol v2
+/// defined in [RFC_2236]
```
dlrobertson: - I don't see the href for this link.
 - AFAIK if this was intended to be module level docs it should be prefixed with `//!`.
 - Why was IGMPv2 (RFC 2236) implemented instead of IGMPv3 (RFC 3376)?

astro:
> Could this be split up into multiple PRs? E.g. could the changes requrired to add IGMP to wire be split out into a a separate PR?

Both the wire and the iface functionality is in the base PR #60 which I cannot rebase because it's too conflicty. Would you like to try yourself?

dlrobertson:
> Both the wire and the iface functionality is in the base PR #60

Ah, np

whitequark:
Can you just split it into two PRs? You can use [Co-authored-by:](https://blog.github.com/2018-01-29-commit-together-with-co-authors/) in commit messages to preserve authorship information.

src/wire/igmpv2.rs
astro: > I don't see the href for this link.

Thanks, I added it.

> AFAIK if this was intended to be module level docs it should be prefixed with //!.

Right! Only that the `igmpv2` mod isn't exported as-is but its individual exports are reexported by mod `wire`, therefore hiding any documentation at the `mod igmpv2` level. I decided to move this documentation chunk to the `IgmpPacket` struct.

> Why was IGMPv2 (RFC 2236) implemented instead of IGMPv3 (RFC 3376)?

I guess @podhrmic has done so for backwards compatibility. v3 can be added at any later time of course.

astro:
> Can you just split it into two PRs? You can use Co-authored-by: in commit messages to preserve authorship information.

Okay, please see #177 and #178.

PR #177, opened 2018-03-05
Implement IGMP packets

astro:
Retake on #60/#176 to clean up commit history.

### DIFF
--- a/src/wire/igmp.rs
+++ b/src/wire/igmp.rs
@@ -426,7 +426,7 @@ mod test {
         for i in 0..256usize {
             let time1 = i as u8;
             let duration = max_resp_code_to_duration(time1);
-            let time2 = duration_to_max_resp_code(&duration);
+            let time2 = duration_to_max_resp_code(duration);
             assert!(time1 == time2);
         }
     }


src/phy/mod.rs
```diff
@@ -161,6 +161,8 @@ pub struct ChecksumCapabilities {
     pub tcpv4: Checksum,
     #[cfg(feature = "proto-ipv4")]
     pub icmpv4: Checksum,
+    #[cfg(feature = "proto-ipv4")]
+    pub igmp: Checksum,
```
whitequark: Is there any hardware that computes IGMP checksums? If not I'd prefer to not have this.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,417 @@
...
+    pub fn set_msg_type(&mut self, value: Message) {
+        let data = self.buffer.as_mut();
+        data[field::TYPE] = value.into()
+    }
+
+    /// Set the Max Resp Time field.
+    #[inline]
+    pub fn set_max_resp_time(&mut self, value: u8) {
+        let data = self.buffer.as_mut();
+        data[field::MAX_RESP_TIME] = value
```
whitequark: Based on the algorithm in `max_resp_time()`, doing `x.set_max_resp_time(x.max_resp_time())` will sometimes change the packet. This is obviously wrong.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,417 @@
...
+            Message::MembershipQuery => {
+                let max_resp_time = packet.max_resp_time();
+                // See RFC 3376: 7.1. Query Version Distinctions
+                let version = if max_resp_time == 0 {
+                    IgmpVersion::Version1
+                } else {
+                    IgmpVersion::Version2
+                };
+                Ok(Repr::MembershipQuery {
+                       max_resp_time,
```
whitequark: Nit: this is indented a bit too much.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,417 @@
...
+                group_addr,
+                version,
+            } => {
+                write!(f,
+                       "IGMPv2 Membership report group_addr={} version={:?}",
+                       group_addr,
+                       version)
+            }
+            &Repr::LeaveGroup { group_addr } => {
+                write!(f, "IGMPv2 Leave Group group_addr={})", group_addr)
```
whitequark: All three branches have inconsistent capitalization. I prefer `membership query`, `membership report` and `leave group`.

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,417 @@
...
+    }
+}
+
+
+/// A high-level representation of an Internet Group Management Protocol v2 header.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Repr {
+    MembershipQuery {
+        /// Maximum Response Time in 1/10th second (100ms) units
+        max_resp_time: u16,
```
whitequark: Since not all `u16` values are valid for `max_resp_time` anyway, I think you should use `Duration` here, and silently clamp values to the closest valid ones.

src/phy/mod.rs
astro: Removed!

astro:
Thank you for the feedback.

Please forgive that I've cheated a bit in `set_max_resp_time()` but generating the exp/mant form is tricky while currently not needed. I added a comment in that place.

whitequark:
No, that's not acceptable. There shouldn't be any broken accessors.

You could make the Packet accessors read and write the raw value--the transformation to/from Duration shouldn't happen at that level anyway--and panic in `Repr::emit`, with documentation.

astro:
Thank you for the feedback. I hope the `max_resp_time` is now properly.

whitequark:
Thanks!
@dlrobertson r?

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,451 @@
...
+
+    /// Return the Max reponse time (in 100ms units)
+    ///
+    /// See [RFC 3376]: 4.1.1. Max Resp Code
+    ///
+    /// [RFC 3376]: https://tools.ietf.org/html/rfc3376
+    #[inline]
+    pub fn max_resp_time(&self) -> u8 {
+        let data = self.buffer.as_ref();
+        data[field::MAX_RESP_TIME].into()
```
dlrobertson: I don't think you need the `into` here if you're returning a `u8`. Did you mean to return a `Duration`?

src/wire/igmpv2.rs
```diff
@@ -0,0 +1,451 @@
...
+
+    #[test]
+    fn duration_to_max_resp_time_max() {
+        for duration in 31744..65536 {
+            let time = duration_to_max_resp_time(
+                &Duration::from_millis(duration * 100)
+            );
+            assert_eq!(time, 0xFF);
+        }
+    }
```
dlrobertson: +1 for the duration tests!

src/wire/mod.rs
```diff
@@ -153,6 +155,11 @@ pub use self::icmpv4::{Message as Icmpv4Message,
                        Packet as Icmpv4Packet,
                        Repr as Icmpv4Repr};
 
+#[cfg(feature = "proto-ipv4")]
+pub use self::igmpv2::{Packet as IgmpPacket,
```
dlrobertson: nit: seems strange to me to have the module named `igmpv2` but the exported structures and enums named `Igmp<Something>`. Would it make sense to just leave the module named `igmp` since we're not going to allow `Igmpv1Packet`, `Igmpv2Packet`, etc?

src/wire/igmpv2.rs
astro: Right, thanks for pointing out!

src/wire/mod.rs
astro: I completely agree. Renamed to `igmp`!

dlrobertson:
@whitequark looks good to me

src/wire/igmp.rs
```diff
@@ -8,7 +8,7 @@ use time::Duration;
 use wire::Ipv4Address;
 
 enum_with_unknown! {
-    /// Internet Group Management Protocol v2 message version/type.
+    /// Internet Group Management Protocol v1/v2 message version/type.
```
whitequark: This module handles both versions 1 and 2.

src/wire/igmp.rs
```diff
@@ -31,7 +31,7 @@ mod field {
     use wire::field::*;
 
     pub const TYPE: usize = 0;
-    pub const MAX_RESP_TIME: usize = 1;
+    pub const MAX_RESP_CODE: usize = 1;
```
whitequark: The field is called Max Resp Code in the RFC, and it's not (always) in centiseconds.

src/wire/igmp.rs
```diff
@@ -323,12 +316,11 @@ fn duration_to_max_resp_time(duration: &Duration) -> u8 {
     }
 }
 
-
 impl<'a, T: AsRef<[u8]> + ?Sized> fmt::Display for Packet<&'a T> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match Repr::parse(self) {
             Ok(repr) => write!(f, "{}", repr),
-            Err(err) => write!(f, "IGMPv2 ({})", err),
+            Err(err) => write!(f, "IGMP ({})", err),
```
whitequark: Again, this could be an IGMP packet of any version.

Issue #105, opened 2017-12-25
Implement support for jumbo frames


whitequark:
I'm not sure if this even needs any work besides testing that it actually works, but right now I don't know what's the status of them.

podhrmic:
Just tested with tap interface and the server example. Once the mtu on the interface is changed (let's say to 5000) I can send and receive 5000 bytes long frames. Do you require any additional testing?

whitequark:
What happens if you set the MTU to be over 9000? At some point it clashes with 802.2 LLC header.

podhrmic:
```
$ ping 192.168.69.1 -s 65000
```
works OK (MTU is set to 65k and I allocated appropriately sized buffers).

Interface settings:
```
16: tap1: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 65500 qdisc pfifo_fast state DOWN mode DEFAULT group default qlen 1000
```

When I send a UDP packet 65000 bytes long, it is both received by the server and returned. However, for some reason the socket buffer isn't cleared - because when I send a second packet of the same size, I get `buffer exhausted` error. But that is imho not related to this issue.

Haven't tested with UDP

whitequark:
Basically, the problem is: using jumbo frames with 802.3 or 802.1Q on the same medium is [ambiguous](https://en.wikipedia.org/wiki/EtherType), and it's not clear how this should be resolved. I'm leaning to rejecting all jumbo frames as invalid by default, unless explicitly enabled by a switch.

podhrmic:
I agree. I ll add an explicit feature switch to enable/disable frames with MTU > 1500

podhrmic:
Looking at it more:

the captured jumbo frames from wireshark have the Ethernet type set correctly (let's say Ipv4) - even for a frame 65k bytes long. 

Since smoltcp doesn't even try to interpred EtherType field as length, I don't think there is anything to be done here. If the hardware doesn't support MTU > 1500, then the packet is dropped, but if it does (like with the tap device) there is no harm in processing jumbo frames. Of course, large frames can be only received if the hardware supports it, and if it does we should process them.

If we add this "jumbo frame switch" we have to validate the payload length against some MTU specified for the ethernet interface and possibly throw an error if is exceeded (or fragment the frame).

What do you think?

podhrmic:
Ah, I guess mtu is already part of the device capabilities so disregard that. Still - is it necessary to drop large packets?

podhrmic:
[...]

Issue #52, opened 2017-10-06
Support IP fragmentation


podhrmic:
In a similar spirit, I would also like to implement support for IP fragmentation. The logic behind fragmentation is pretty straightforward, but the way to implement it less so.

So far it looks like it would need:
- a queue for the fragmented IP packets, where they can be stored before reassembling and passing them to the higher layers
- timers associated with each fragmented packet (similar to #25 ) so they can be dropped after a timeout

I think this means adding a queue to the https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs 

Thoughts?

whitequark:
Basically. What you need is...

  * one `storage::Assembler` per fragmented packet;
  * one `managed::ManagedSlice` per fragmented packet;
  * one `storage::RingBuffer` per interface, storing `FragmentedPacket`s in a FIFO queue;
  * one `managed::ManagedMap` per interface, mapping `Option<u16>` (ident) to `usize`s (positions in the queue).

Broadly, how it would work is:
  * every time a fragmented packet is received, its ident is looked up in the map and the corresponding FragmentedPacket is updated if it's still within the bounds;
  * if it's not within bounds, the entry is removed;
  * positions in the queue in the map and positions in the ring buffer are mapped to each other through an offset that is increased every time the queue has a packet removed from it.

whitequark:
The main issue here is that `managed::ManagedMap` doesn't exist. It should be a generalized version of `SliceArpCache`, using a slice on heap-free platforms and a BTreeMap otherwise.

whitequark:
#51 also needs a ManagedMap.

whitequark:
There is now a `managed::ManagedMap`. Also, scratch what I said about FIFO buffers, that doesn't sound like a very good idea on reflection. I'm not actually sure what sort of data structure can be used to implement reassembly without fixing MTU upfront or suffering internal fragmentation.

podhrmic:
After looking at [RFC 815](https://tools.ietf.org/html/rfc815) I have some ideas.

First, the buffer for each fragmented packet will have to be manually allocated (similar to defining buffers for sockets), so the number and size can be user-configured. 

Second, we need to check packet ID, as well as source and destination address. Something like this:

```Rust
struct FragmentedPacket {
	assebler: Assembler,
	rx_buffer: RingBuffer<'a>,
	src_addr: Ipv4Address,
	dst_addr: Ipv4Address,
	id: u16,
	// timeout?
}
[...]

PR #185, opened 2018-03-30
Initial commit for fragmentation support && Ethernet jumbo frames support

podhrmic:
I know, it should be two separate commits - but I noticed it too late. Anyway, since I expect changes to be requested, I will deal with that properly later.

Support for different ethernet frames: deals with #105 
I am testing if the frame is a valid Ethernet II type frame (the type/length >= 1536 see https://en.wikipedia.org/wiki/Ethernet_frame#Types ) so there is potential to process other types of Ethernet in the future. Also, added an explicit `cfg` to allow large frames, and fixed the interfaces to handle frames that are mtu size.

Fragmentation support: related to #52 
I rewrote the `src/iface/etherne.rs` to optionally handle fragmented packets (that is the major feature). The code that handles reassembly how handles only best case scenario, and shouldn't be considered final. The same for the fragments.rs - I don't know how to handle ManagedSlices properly (at least not yet), so for time being it is `std` only feature backed by a vec. 

In both cases I am looking for general feedback about the approach, the details can be polished later. Next I want to improve the fragments reassembly algorithm so it handles error cases and out-of-order fragments. After that perhaps fragmenting larger than MTU ougoing packets.

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -82,7 +82,7 @@ fn main() {
         
 
         // udp:6969: respond "hello"
-        /*{
+        {
             let mut socket = sockets.get::<UdpSocket>(udp_handle);
             if !socket.is_open() {
                 socket.bind(6969).unwrap()
@@ -102,7 +102,7 @@ fn main() {
                        str::from_utf8(data.as_ref()).unwrap());
                 socket.send_slice(data, endpoint).unwrap();
             }
-        }*/
+        }
                 // udp:6942: echo with reverse
         {
             let mut socket = sockets.get::<UdpSocket>(udp_handle);
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -667,41 +667,146 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         }
     }
 
-    #[cfg(feature = "proto-ipv4")]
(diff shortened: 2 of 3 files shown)


examples/server.rs
```diff
@@ -66,11 +66,20 @@ fn main() {
...
+    let mut fragments = FragmentsSet::new(vec![]);
+    let fragment = FragmentedPacket::new(vec![0; 65535]);
+    fragments.add(fragment);
+    
 
     let mut tcp_6970_active = false;
     loop {
         let timestamp = Instant::now();
-        iface.poll(&mut sockets, timestamp).expect("poll error");
+        match iface.poll(&mut sockets, &mut fragments, timestamp) {
```
podhrmic: This way the fragments are handled similarly to a socket set. Other option would be to have a fragment set attached to an interface, I am not sure which one is better.

examples/server.rs
```diff
@@ -94,6 +103,40 @@ fn main() {
                 socket.send_slice(data, endpoint).unwrap();
             }
         }
+                // udp:6942: echo with reverse
```
podhrmic: this is useful for testing

podhrmic:
@whitequark have you had a chance to look at this?

podhrmic:
Closing in favor of #190

PR #186, opened 2018-04-05
DHCPv4 client

astro:
Hi

This is an attempt at creating a DHCP client for IPv4. Please give feedback.

Cc: @phil-opp, this PR is uses the work merged in #75.

I've taken a completely different approach compared to #63. UDP sockets don't allow us to send and receive on unspecified/broadcast IPv4 addresses. Therefore IPv4/UDP is parsed/serialized on top of a `RawSocket`.

~~There are still two problems with `RawSocketBuffer`:~~
* ~~How do I let the user provide storage? I have tried but failed to accept that as parameters in `Client::new()`.~~ I've found the proper lifetime relation by now.
* ~~The `RawSocket` is getting stuck with `Error::Exhausted` when `contig_window` is too small for the DHCP egress packet. Somehow it appears to behave like sent packets don't get dequeued from the `tx_buffer`. Is that known? What is my code doing wrong regarding RawSocket?~~ See #187

PR #187, opened 2018-04-09
RingBuffer contiguous memory optimization

astro:
I've studied the code for a while and wondered why there is a check for `capacity > 0` at this location. `capacity` doesn't change. I believe `self.length` was actually intended here.

In the RawSocket use by #186 that's needed otherwise `PacketBuffer` tends to get a stuck `payload_ring` (`length=0` but `read_at` near `capacity`).

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -7,7 +7,7 @@ use core::{cmp, fmt};
 use {Error, Result};
 use phy::DeviceCapabilities;
 use time::{Duration, Instant};
-use socket::{Socket, SocketMeta, SocketHandle};
+use socket::{Socket, SocketMeta, SocketHandle, PollAt};
 use storage::{Assembler, RingBuffer};
 use wire::{IpProtocol, IpRepr, IpAddress, IpEndpoint, TcpSeqNumber, TcpRepr, TcpControl};
 
@@ -59,6 +59,7 @@ enum Timer {
         expires_at: Instant,
         delay:      Duration
     },
+    FastRetransmit,
     Close {
         expires_at: Instant
     }
@@ -89,7 +90,8 @@ impl Timer {
             Timer::Retransmit { expires_at, delay }
                     if timestamp >= expires_at => {
                 Some(timestamp - expires_at + delay)
-            }
+            },
+            Timer::FastRetransmit => Some(Duration::from_millis(0)),
             _ => None
         }
     }
@@ -104,11 +106,13 @@ impl Timer {
         }
     }
 
-    fn poll_at(&self) -> Option<Instant> {
+    fn poll_at(&self) -> PollAt {
         match *self {
-            Timer::Idle { keep_alive_at } => keep_alive_at,
-            Timer::Retransmit { expires_at, .. } => Some(expires_at),
-            Timer::Close { expires_at } => Some(expires_at),
+            Timer::Idle { keep_alive_at: Some(keep_alive_at) } => PollAt::Time(keep_alive_at),
+            Timer::Idle { keep_alive_at: None } => PollAt::Ingress,
+            Timer::Retransmit { expires_at, .. } => PollAt::Time(expires_at),
+            Timer::FastRetransmit => PollAt::Now,
+            Timer::Close { expires_at } => PollAt::Time(expires_at),
         }
     }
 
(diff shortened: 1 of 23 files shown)


astro:
The test strings were changed to make the CI pass.

I didn't encounter any division by zero errors. Should I try to solve my problem in a different way?

whitequark:
If you're changing tests "to make CI green" that means it's equally likely that there was a bug in code or a bug in tests or a bug in specs but you didn't distinguish these cases. Which one is it?

On April 10, 2018 10:17:36 PM GMT+08:00, Astro <notifications@github.com> wrote:
>The test strings were changed to make the CI pass.
>
>I didn't encounter any division by zero errors. Should I try to solve
>my problem in a different way?
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/m-labs/smoltcp/pull/187#issuecomment-380115827

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

astro:
I'm not 100% certain that this is the right approach. With the optimization, the buffer will be rewritten from start if it was empty after the previous `dequeue_many_with()`, hence the spec changes.

whitequark:
It seems we have two separate things at work here:
  1. A possible bug: "I've studied the code for a while and wondered why there is a check for capacity > 0 at this location. capacity doesn't change. I believe self.length was actually intended here."
  2. An optimization (?): "In the RawSocket use by #186 that's needed otherwise PacketBuffer tends to get a stuck payload_ring (length=0 but read_at near capacity)."

@astro, can you please provide separate test cases for (1) and (2)? If the check is genuinely dead, then let's first remove it, and then separately examine the optimization (2); I don't fully understand why it is necessary yet.

astro:
Ok, I reordered this PR with the updated tests first and had CI trigger on them.

Now I've put the optimization into `enqueue_many_with()` which is the method that benefits from this. Subsequently I had to fundamentally touch `test_buffer_wraparound_tx()` to trigger the intended check.

src/socket/tcp.rs
```diff
@@ -3656,33 +3656,25 @@ mod test {
...
-        send!(s, TcpRepr {
-            seq_number: REMOTE_SEQ + 1,
-            ack_number: Some(LOCAL_SEQ + 1 + 3),
-            ..SEND_TEMPL
-        });
-        assert_eq!(s.send_slice(b"defghi"), Ok(6));
         recv!(s, Ok(TcpRepr {
-            seq_number: LOCAL_SEQ + 1 + 3,
+            seq_number: LOCAL_SEQ + 1 + 6,
             ack_number: Some(REMOTE_SEQ + 1),
```
astro: TcpSocket is acking the same seq_number for a second time. Is that another bug?

src/socket/tcp.rs
```diff
@@ -3656,33 +3656,25 @@ mod test {
     #[test]
     fn test_buffer_wraparound_tx() {
         let mut s = socket_established();
-        s.tx_buffer = SocketBuffer::new(vec![0; 6]);
-        assert_eq!(s.send_slice(b"abc"), Ok(3));
+        s.tx_buffer = SocketBuffer::new(vec![b'.'; 9]);
+        assert_eq!(s.tx_buffer.enqueue_many(6).len(), 6);
+        assert_eq!(s.tx_buffer.dequeue_many(3).len(), 3);
```
whitequark: Why are you manipulating the buffer directly in this test and not via the socket API? This defeats the purpose of testing the socket.

astro: Because this is just preparing the buffer. Anyway, fixed it.

src/socket/tcp.rs
```diff
@@ -3656,33 +3656,25 @@ mod test {
     #[test]
     fn test_buffer_wraparound_tx() {
         let mut s = socket_established();
-        s.tx_buffer = SocketBuffer::new(vec![0; 6]);
-        assert_eq!(s.send_slice(b"abc"), Ok(3));
+        s.tx_buffer = SocketBuffer::new(vec![b'.'; 9]);
+        assert_eq!(s.send_slice(b"xxxyyy"), Ok(6));
+        assert_eq!(s.tx_buffer.dequeue_many(3), &b"xxx"[..]);
```
whitequark: This is not any better...

astro: Why touch the internal `tx_buffer` you ask? The `read_at` pointer must be initialized so that the check still makes sense when the optimization is merged.

whitequark:
@astro Thanks, and sorry for taking so much time to merge this.

PR #188, opened 2018-04-09
Fix DhcpOption::Other serialization.

astro:
`data` is a `&'a [u8]` and therefore has no fixed length of 4 bytes.


whitequark:
Could you add a test?

astro:
Test added.

PR #190, opened 2018-04-10
Support for receiving fragmented IPv4 packets

podhrmic:
Implemented as an optional feature. Provided tests and expanded `server.rs` example. The code should be self-explanatory. Tested with UDP and ICMP packets. Replaces #185

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -53,7 +53,11 @@ fn main() {
     let tcp4_socket = TcpSocket::new(tcp4_rx_buffer, tcp4_tx_buffer);
 
     let ethernet_addr = EthernetAddress([0x02, 0x00, 0x00, 0x00, 0x00, 0x01]);
-    let ip_addrs = [IpCidr::new(IpAddress::v4(192, 168, 69, 1), 24)];
+    let ip_addrs = [
+        IpCidr::new(IpAddress::v4(192, 168, 69, 1), 24),
+        IpCidr::new(IpAddress::v6(0xfdbe, 0x0000, 0x0000, 0x0000,
+                                  0x0000, 0x0000, 0x0000, 0x002a), 64)
+    ];
 
     let mut iface = EthernetInterfaceBuilder::new(device)
             .ethernet_addr(ethernet_addr)
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -20,6 +20,10 @@ use wire::{ArpPacket, ArpRepr, ArpOperation};
 use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
 #[cfg(feature = "proto-ipv6")]
 use wire::{Icmpv6Packet, Icmpv6Repr, Icmpv6ParamProblem};
+#[cfg(all(feature = "socket-icmp", any(feature = "proto-ipv4", feature = "proto-ipv6")))]
+use wire::IcmpRepr;
+#[cfg(feature = "proto-ipv6")]
+use wire::{NdiscNeighborFlags, NdiscRepr};
 #[cfg(all(feature = "proto-ipv6", feature = "socket-udp"))]
 use wire::Icmpv6DstUnreachable;
 #[cfg(feature = "socket-udp")]
@@ -30,7 +34,7 @@ use wire::{TcpPacket, TcpRepr, TcpControl};
 use socket::{Socket, SocketSet, AnySocket};
 #[cfg(feature = "socket-raw")]
 use socket::RawSocket;
-#[cfg(all(feature = "socket-icmp", feature = "proto-ipv4"))]
+#[cfg(all(feature = "socket-icmp", any(feature = "proto-ipv4", feature = "proto-ipv6")))]
 use socket::IcmpSocket;
 #[cfg(feature = "socket-udp")]
 use socket::UdpSocket;
@@ -62,6 +66,8 @@ struct InterfaceInner<'b, 'c, 'e> {
     ip_addrs:               ManagedSlice<'c, IpCidr>,
     #[cfg(feature = "proto-ipv4")]
     ipv4_gateway:           Option<Ipv4Address>,
+    #[cfg(feature = "proto-ipv6")]
+    ipv6_gateway:           Option<Ipv6Address>,
     device_capabilities:    DeviceCapabilities,
     fragments:              Option<FragmentSet<'e>>,
 }
@@ -73,9 +79,11 @@ pub struct InterfaceBuilder <'b, 'c, 'e, DeviceT: for<'d> Device<'d>> {
     ethernet_addr:       Option<EthernetAddress>,
     neighbor_cache:      Option<NeighborCache<'b>>,
     ip_addrs:            ManagedSlice<'c, IpCidr>,
+    fragments:           Option<FragmentSet<'e>>,
     #[cfg(feature = "proto-ipv4")]
     ipv4_gateway:        Option<Ipv4Address>,
-    fragments:           Option<FragmentSet<'e>>,
+    #[cfg(feature = "proto-ipv6")]
+    ipv6_gateway:        Option<Ipv6Address>,
 }
 
 impl<'b, 'c, 'e, DeviceT> InterfaceBuilder<'b, 'c, 'e, DeviceT>
@@ -114,6 +122,8 @@ impl<'b, 'c, 'e, DeviceT> InterfaceBuilder<'b, 'c, 'e, DeviceT>
             #[cfg(feature = "proto-ipv4")]
             ipv4_gateway:        None,
             fragments:        None,
+            #[cfg(feature = "proto-ipv6")]
+            ipv6_gateway:        None,
         }
     }
 
@@ -158,8 +168,25 @@ impl<'b, 'c, 'e, DeviceT> InterfaceBuilder<'b, 'c, 'e, DeviceT>
         where T: Into<Ipv4Address>
     {
         let addr = gateway.into();
-        InterfaceInner::check_gateway_addr(&addr);
+        InterfaceInner::check_ipv4_gateway_addr(&addr);
         self.ipv4_gateway = Some(addr);
+        self
+    }
+
+    /// Set the IPv6 gateway the interface will use. See also
+    /// [ipv6_gateway].
+    ///
+    /// # Panics
+    /// This function panics if the given address is not unicast.
+    ///
+    /// [ipv6_gateway]: struct.EthernetInterface.html#method.ipv6_gateway
+    #[cfg(feature = "proto-ipv6")]
+    pub fn ipv6_gateway<T>(mut self, gateway: T) -> InterfaceBuilder<'b, 'c, DeviceT>
+        where T: Into<Ipv6Address>
+    {
+        let addr = gateway.into();
+        InterfaceInner::check_ipv6_gateway_addr(&addr);
+        self.ipv6_gateway = Some(addr);
         self
     }
 
@@ -201,6 +228,8 @@ impl<'b, 'c, 'e, DeviceT> InterfaceBuilder<'b, 'c, 'e, DeviceT>
                         #[cfg(feature = "proto-ipv4")]
                         ipv4_gateway: self.ipv4_gateway,
                         fragments: self.fragments,
+                        #[cfg(feature = "proto-ipv6")]
+                        ipv6_gateway: self.ipv6_gateway,
                     }
                 }
             },
@@ -280,25 +309,6 @@ impl<'b, 'c, 'e, DeviceT> Interface<'b, 'c, 'e, DeviceT>
         self.inner.ip_addrs.as_ref()
     }
 
-    /// Determine if the given `Ipv6Address` is the solicited node
-    /// multicast address for a IPv6 addresses assigned to the interface.
-    /// See [RFC 4291 § 2.7.1] for more details.
-    ///
-    /// [RFC 4291 § 2.7.1]: https://tools.ietf.org/html/rfc4291#section-2.7.1
-    #[cfg(feature = "proto-ipv6")]
-    pub fn has_solicited_node(&self, addr: Ipv6Address) -> bool {
-        self.inner.ip_addrs.iter().find(|cidr| {
-            match *cidr {
-                &IpCidr::Ipv6(cidr) if cidr.address() != Ipv6Address::LOOPBACK=> {
-                    // Take the lower order 24 bits of the IPv6 address and
-                    // append those bits to FF02:0:0:0:0:1:FF00::/104.
-                    addr.as_bytes()[14..] == cidr.address().as_bytes()[14..]
-                }
-                _ => false,
-            }
-        }).is_some()
-    }
-
     /// Update the IP addresses of the interface.
     ///
     /// # Panics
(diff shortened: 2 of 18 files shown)


src/iface/ethernet.rs
```diff
@@ -667,14 +684,93 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+                    }
+                    fragment.reset();
+                    return Ok(Some(Ipv4Packet::new_checked(&fragment.rx_buffer[0..front])?));
+                }
+
+                // not the last fragment
+                return Ok(None);
+        	},
+        	None => {
+        		return Ok(None); // TODO: throw an exception?
```
whitequark: Also log it, with a different message indicating that fragmentation is not supported.

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
+use storage::Assembler;
+use time::Instant;
+use managed::ManagedSlice;
+
+/// Default timeout duration
+const FRAGMENTATION_TIMEOUT_MS: i64 = 500;
+
+#[allow(dead_code)]
```
whitequark: Better not include the module at all if built without the fragmentation feature.

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
...
+pub struct Packet<'a, T: 'a> {
+    assembler: Assembler,
+    pub rx_buffer: ManagedSlice<'a, T>,
+    id: u16,
```
whitequark: `ident`

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
...
+    last_used: Instant,
+    header: bool,
```
whitequark: `has_header`

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
...
+use time::Instant;
+use managed::ManagedSlice;
+
+/// Default timeout duration
+const FRAGMENTATION_TIMEOUT_MS: i64 = 500;
+
+#[allow(dead_code)]
+pub struct Packet<'a, T: 'a> {
+    assembler: Assembler,
+    pub rx_buffer: ManagedSlice<'a, T>,
```
whitequark: Why is this `pub`?

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
+use storage::Assembler;
+use time::Instant;
+use managed::ManagedSlice;
+
+/// Default timeout duration
+const FRAGMENTATION_TIMEOUT_MS: i64 = 500;
+
+#[allow(dead_code)]
+pub struct Packet<'a, T: 'a> {
```
whitequark: Why is it generic over `T`? We'll never want to defragment anything except `u8`. This just results in silly `T: Clone` bounds on functions.

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
...
+        let range = offset+self.header_len..offset+payload_len+self.header_len;
+        self.rx_buffer[range].clone_from_slice(&data[header_len..]);
+        Ok(())
+    }
+
+    /// Return `true` if the packet has all fragments,
+    /// and can be reassmbled, and `false` otherwise
+    pub fn check_contig_range(&self) -> bool {
+        if self.total_len != 0 {
+            match self.assembler.peek_front() {
```
whitequark: `if let Some(front) = self.assembler.peek_front()`

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
...
+            id: 0,
+            last_used: Instant::from_millis(0),
+            header: false,
+            header_len: 0,
+            total_len: 0,
+        }
+    }
+
+    /// Reset packet
+    pub fn reset(&mut self) {
```
whitequark: I think `Packet` would be much cleaner if instead of these empty values and `set_id` it was defined like...
```
struct Packet {
  rx_buffer: ManagedSlice<'a, T>,
  state: PacketState  
}

enum PacketState {
  Empty,
  Assembling {
    assembler: ...,
    has_header: ...,
    etc
  }
}
```
and then instead of methods like `set_id` you would have methods for transition between states, i.e. `start(ident: u16)`, `add_fragment(...)`, `extract_packet()`, and calling a method in the state that isn't correct for it would panic. Currently, this class of logic errors is not guarded against at all.

src/iface/fragments.rs
```diff
@@ -0,0 +1,170 @@
...
+		        }		
+    		},
+    		#[cfg(any(feature = "std", feature = "alloc"))]
+    		ManagedSlice::Owned(ref mut packets) => {
+    			packets.push(new_packet);
+    		}
+    	}
+    }
+
+    /// Check if the set contains a packet with given `id`
```
whitequark: This isn't enough. You need to at least match the ident field with the source IP, because otherwise you're introducing a vulnerability: anyone who knows that there is a flow of fragmented packets has an 1/65536 chance of arbitrarily altering their contents. This is a 100% viable attack vector and very similar attacks (utilizing the port number, but the concept is the same) have been demonstrated against DNS two decades ago.

whitequark: Also, you probably should use a ManagedMap here instead of hardcoding a linear search.

src/phy/mod.rs
```diff
@@ -265,3 +265,8 @@ pub trait TxToken {
     fn consume<R, F>(self, timestamp: Instant, len: usize, f: F) -> Result<R>
         where F: FnOnce(&mut [u8]) -> Result<R>;
 }
+
+
+/// Ethernet header length to be added to the MTU, since the header is not counted into MTU
+/// 6 bytes src addr, 6 bytes dst addr, 2 bytes EthType, (optional) 4 bytes IEEE 802.1Q Header
+pub const ETHERNET_HAEDER_MAX_LEN: usize = 18;
```
whitequark: This doesn't belong in this pull request. Please submit it separately and we'll have a discussion about it; in particular, you're counting 802.1Q, but not QinQ or QinQinQ.

whitequark:
Please *please* fix indentation. It makes the code very hard to read.

examples/server.rs
```diff
@@ -17,6 +17,8 @@ use smoltcp::socket::SocketSet;
 use smoltcp::socket::{UdpSocket, UdpSocketBuffer, UdpPacketMetadata};
 use smoltcp::socket::{TcpSocket, TcpSocketBuffer};
 use smoltcp::time::{Duration, Instant};
+#[cfg(feature = "fragmentation-ipv4")]
+use smoltcp::iface::{FragmentsSet, FragmentedPacket};
```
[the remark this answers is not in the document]
podhrmic: Done. On a side note, fragmentation is currently a part of the default features. If I remove it, it makes building the server_fragmented example really verbose (you have to enumerate all features). Is there some better way how to do that?

src/iface/ethernet.rs
```diff
@@ -55,27 +56,29 @@ pub struct Interface<'b, 'c, DeviceT: for<'d> Device<'d>> {
...
 /// exclusively). However, it is still possible to call methods on its `inner` field.
-struct InterfaceInner<'b, 'c> {
+struct InterfaceInner<'b, 'c, 'e> {
     neighbor_cache:         NeighborCache<'b>,
     ethernet_addr:          EthernetAddress,
     ip_addrs:               ManagedSlice<'c, IpCidr>,
     #[cfg(feature = "proto-ipv4")]
     ipv4_gateway:           Option<Ipv4Address>,
     device_capabilities:    DeviceCapabilities,
+    fragments:				Option<FragmentsSet<'e>>,
```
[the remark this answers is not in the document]
podhrmic: Fixed.

src/iface/fragments.rs
podhrmic: But I need to include `FragmentSet` because interface now contains `fragments:              Option<FragmentSet<'e>>`, except without enabled fragmentation it compiles away as it is always None

src/iface/fragments.rs
podhrmic: because `process_ipv4_fragment` returns `return Ok(Some(Ipv4Packet::new_checked(&fragment.rx_buffer[0..front])?));`

So having a getter function that returns a slice from a given range seemed unnecessary. But I am happy to add it.

src/iface/ethernet.rs
```diff
@@ -667,14 +684,93 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+				    	fragments: &'r mut Option<FragmentsSet<'e>>
+				    ) ->
+                     Result<Option<Ipv4Packet<&'r [u8]>>> 
+    {
+        match *fragments {
+        	Some(ref mut fragments) => {
+                // get an existing fragment or attempt to get a new one
+				let fragment = match fragments.get_packet(ipv4_packet.ident(), timestamp) {
+				    Some(frag) => frag,
+				    None => return Ok(None), // TODO: throw an exception?
```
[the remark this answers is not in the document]
podhrmic: I meant return an error. Fixed.

src/iface/ethernet.rs
podhrmic: Fixed

src/iface/fragments.rs
podhrmic: should be fixed

src/iface/fragments.rs
podhrmic: fixed

src/iface/fragments.rs
podhrmic: Fixed.

src/phy/mod.rs
podhrmic: This is needed to get the fragmentation to work. In Linux MTU refers to the max length of the ethernet payload. But smoltcp counts MTU as max length of the ethernet frame. Hence without this added ethernet length, all fragmented packets are truncated. I am happy to fix is more meaningfully later, but without this fix the fragmentation reassembly doesn't work.

podhrmic:
Work in progress - will ping you once I have all your comments implemented.

examples/server.rs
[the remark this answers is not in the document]
whitequark: I think it's fine to have fragmentation in default features as it's required in IPv4. API consumers can always not initialize the FragmentSet (which will happen by default), so this would only impact code size, and anyone conscious about code size should already turn off unused features.

src/iface/ethernet.rs
```diff
@@ -393,7 +406,8 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
                 Some(tokens) => tokens,
             };
             rx_token.consume(timestamp, |frame| {
-                inner.process_ethernet(sockets, timestamp, &frame).map_err(|err| {
+            	let mut fragments = ::std::mem::replace(&mut inner.fragments, None);
```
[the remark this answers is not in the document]
podhrmic: Currently requires `std`. See my comments in the updated version with proposed solutions.

src/iface/fragments.rs
podhrmic: Done.

src/iface/fragments.rs
podhrmic: Now I am checking the id and src & dst addresses. 
ManagedMap is not possible here because each packet needs an underlying storage. So I cannot simply add a new packet to the map, unless I explicitly provide a storage for that packet.

podhrmic:
@whitequark please see the new commit. The main issue I found out is that currently, fragmentation requires `std` for the memory swap. The proposed solution is in `ethernet.rs` at line 410.

whitequark:
> The main issue I found out is that currently, fragmentation requires std for the memory swap.

Huh? [core::mem::replace](https://doc.rust-lang.org/nightly/core/mem/fn.replace.html) exists. But I strongly dislike this solution. The code in EthernetInterface is already awful, let's not make it worse.

src/iface/fragments.rs
whitequark: Of course it is possible, you use a map between ident/addresses and the *index* into the packet storage.

src/phy/mod.rs
whitequark: Can you extract it into a separate PR that we merge before this one?

podhrmic:
> Huh? core::mem::replace exists. But I strongly dislike this solution. The code in EthernetInterface is already awful, let's not make it worse.

Nice! I missed that one. Updated and tested. I am curious if you find a better solution, as I said, I spent quite some time navigating between lifetimes and borrows on this one.

podhrmic:
And one more question - in my `server_fragmented` example, I have the following rx buffer for UDP data:
```Rust
let udp_rx_buffer = UdpSocketBuffer::new(vec![UdpPacketMetadata::EMPTY], vec![0; 9000]);
```

I can receive 2 reassembled packets, each 4096 bytes long, but the third one triggers this error:
```
[1523987506.48s] (iface::ethernet): cannot process ingress packet: buffer space exhausted
```

Am I missing something? Shouldn't the buffer be cleared once the packet is processed?A similar thing happens to the regular server example.

whitequark:
Might be related to https://github.com/m-labs/smoltcp/pull/187 cc @astro

whitequark:
Can you please write a test case for this, on top of `master`? I would really appreciate that.

podhrmic:
OK, will give it a try.

podhrmic:
Another thing - I am not sure how to handle the CI build fails. In short, fragmentation currently works only for ipv4, but Ethernet Interface needs FragmentsSet even when it is not used. Hence the build fails when not using ipv4...:-/

whitequark:
You could conditionally include the entire module.
```
mod fragments;
```

src/wire/ipv4.rs
```diff
@@ -273,7 +273,7 @@ impl<T: AsRef<[u8]>> Packet<T> {
             Err(Error::Truncated)
         } else if self.header_len() as u16 > self.total_len() {
             Err(Error::Malformed)
-        } else if len < self.total_len() as usize {
+        } else if len < self.total_len() as usize && self.dont_frag() {
```
whitequark: This needs an added test.

src/iface/fragments.rs
```diff
@@ -0,0 +1,278 @@
...
+            },
+            PacketState::Assembling{..} => {
+                &mut self.rx_buffer[start..end]
+            }
+        }
+    }
+}
+
+pub struct Set<'a> {
+	packets: ManagedSlice<'a, Packet<'a>>,
```
whitequark: I would like this to use a ManagedMap as I have described previously.

src/wire/ipv4.rs
podhrmic: Can you elaborate? If the packet is not fragmented, the check goes on as before. If it is fragmented, I skip this check (and verify the length upon assembly).

src/iface/fragments.rs
podhrmic: With ManagedMap and std it goes as follows:

1. new fragment arrives
2. fragment is not in the FragmentSet, create a new FragmentedPacket (its storage backed by a `vec!`) and insert it with the key to the FragmentSet

Now, with `no_std`, you cannot create a new FragmentedPacket with a `vec!` - you need an array, and you cannot allocate it from within the `process_ipv4_fragment` function because of lifetimes. So all FragmentedPackets have to be initialized prior polling (hence ManagedSlice).

Does it make sense?

m-labs-homu:
:umbrella: The latest upstream changes (presumably 6c363762845e69dd316617f7ae0a083b4e534d5b) made this pull request unmergeable. Please resolve the merge conflicts.

m-labs-homu:
:umbrella: The latest upstream changes (presumably 2363f97834b3d529c6344f1840ab001645e853d2) made this pull request unmergeable. Please resolve the merge conflicts.

podhrmic:
Closing in favor of #236

src/phy/mod.rs
podhrmic: See #237

PR #191, opened 2018-04-12
Add Repr support for NDISC packets

dlrobertson:
Add Representation layer support for NDISC packets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,15 +336,15 @@ that do. Because of this, only one such example is provided.
 ### examples/loopback.rs
 
 _examples/loopback.rs_ sets up _smoltcp_ to talk with itself via a loopback interface.
-Although it does not require `std`, this example still requires the `alloc` feature to run.
+Although it does not require `std`, this example still requires the `alloc` feature to run, as well as `log`, `proto-ipv4` and `socket-tcp`.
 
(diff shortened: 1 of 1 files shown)


dlrobertson:
After this we can start on address resolution on IPv6!

whitequark:
Thanks for the PR! Since `icmpv6.rs` contains only "base" ICMPv6 packets right now, with `ndisc.rs` containing NDISC packet code, do you think you can move out `Repr` (which of course cannot be easily split) into `icmpv6repr.rs`?

dlrobertson:
You mean have a separate module with all of `Icmpv6Repr` in it or a separate module with the NDISC variants.

whitequark:
The former--I don't think the latter is possible in Rust.

whitequark:
Actually, on second thought, that will cause issues with tests. Let me think a bit more about the best course of action here.

dlrobertson:
What about adding a `Ndisc(NdiscRepr)` variant to `Icmpv6Repr`?

whitequark:
Let's do that.

dlrobertson:
:+1:

dlrobertson:
Updated to use a `Ndisc(NdiscRepr<'a>)` variant.

whitequark:
Thanks! This is very nice.

PR #193, opened 2018-04-17
Resource exhaustion test

podhrmic:
This is an attempt to recreate a simple resource exhaustion test.
The intent is to test whether the received data are dropped once they are processed (i.e. the buffer should never be full unless a single packet is larger than the backing up buffer). Indeed, this test fails and I am not sure why.

The real-life test is to run `server.rs` and send UDP packets at port 6969 until the buffer is exhausted.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -1419,6 +1419,49 @@ mod test {
 
     #[test]
     #[cfg(all(feature = "socket-udp", feature = "proto-ipv4"))]
+    fn test_udp_buffer_exhaustion() {
+        use socket::{UdpSocket, UdpSocketBuffer, UdpPacketMetadata};
+        use wire::IpEndpoint;
+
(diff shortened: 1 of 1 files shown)


astro:
Thank you for reporting this issue. Great that I'm not the only one having this problem! It is solved by PR #187.

Please note that your test does not terminate when there is no more problem.

podhrmic:
Thanks for fixing it. Do you want me to add some termination criteria to the test (assuming we want to keep it)?

whitequark:
So, the underlying cause is that PacketBuffer may use one entry in the metadata ring buffer for the padding, which causes buffer exhaustion when the actual metadata entry is inserted.

I confirm that #187 is the correct way to fix this.

podhrmic:
Fixed with https://github.com/m-labs/smoltcp/pull/187/commits/6e2da5e809f669cc59d080f4206c6f1f90628e02

PR #197, opened 2018-04-28
Do not export phy::wait on non-Unix platforms

podhrmic:
Necessary to use smoltcp on seL4 (and perhaps other platforms).

By @aisamanra


whitequark:
Thanks!

PR #198, opened 2018-04-30
Support for seL4 interface

podhrmic:
Following #197 this PR adds support for seL4 interfaces. seL4 provides network driver, which has two functions 'ethdriver_rx()' and 'ethdriver_tx' - the Sel4Device is basically a wrapper around those.


dlrobertson:
Normally a device or OS specific driver is implemented outside of this repo. E.g. the [RedoxOS driver](https://github.com/redox-os/netstack/blob/57cb6cbbefc2a5c87c9fc01bb284b8d4d212a82f/src/smolnetd/device.rs)

whitequark:
I am concerned about merging this. Specifically, there is no way I can test this code. (Indeed, it fails on Travis right now.)

podhrmic:
@dlrobertson thanks for the pointer, that indeed makes more sense. I will keep it separate.

PR #200, opened 2018-05-07
Update remote_last_win for all ACKs

### LINKED ISSUE #199: Stalled TCP connection because window update is not announced after reading
Only ACKs generated by `dispatch` are updating `remote_last_win` but not the ACKs generated by  `process`, failing to keep track of the last announced receive window. This leaves `window_to_update` to return false in `dispatch` after the RX buffer was emptied, thus not announcing that the receive window is not zero anymore after reading from the RX buffer → Connection is stalled if keep-alive is off.

I will submit a patch now.

Actually I also want to use this issue to ask if _Probing Zero Windows_ should be added to the list as **not supported**: https://tools.ietf.org/html/rfc1122#page-92 (This requires a timer for a sender which got a zero window announced from the receiver.)

pothos:
Only ACKs generated by `dispatch` were updating
`remote_last_win` but not the ACKs generated by
`process`, failing to keep track of the last
announced receive window. This left
`window_to_update` to return false in `dispatch`
after the RX buffer was emptied, thus not announcing
that the receive window is not zero anymore.

The change updates `remote_last_win` for immediate
ACKs generated by `process`.

Fixes #199

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -782,7 +782,7 @@ impl<'a> TcpSocket<'a> {
         (ip_reply_repr, reply_repr)
     }
 
-    fn ack_reply(&self, ip_repr: &IpRepr, repr: &TcpRepr) -> (IpRepr, TcpRepr<'static>) {
+    fn ack_reply(&mut self, ip_repr: &IpRepr, repr: &TcpRepr) -> (IpRepr, TcpRepr<'static>) {
         let (ip_reply_repr, mut reply_repr) = Self::reply(ip_repr, repr);
 
         // From RFC 793:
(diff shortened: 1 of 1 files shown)


whitequark:
Excellent work, thanks!
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 0ad249c has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 0ad249cdeaba3cd42c296769f39d3c3891a9ef0f with merge 7935f0d0b5692114ef8ca740d92f385c3393c837...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/375998726?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 7935f0d0b5692114ef8ca740d92f385c3393c837 to master...

Issue #99, opened 2017-12-23
ICMP sockets should not use ICMPv4 specific structures.


dlrobertson:
Currently the implementation of `IcmpSocket` is only enabled if both `socket-icmp` and `proto-ipv4` is enabled because it directly uses `Icmpv4Repr`. A generic `IcmpRepr` type should be created.

dlrobertson:
Fixed by #205

PR #205, opened 2018-05-09
Add support for ICMPv6 to ICMP sockets

progval:
Hi,

I started adding support for ICMPv6 to ICMP sockets (see #99 and #103).

Do you have some feedback on my changes (attached to this PR) before I implement it further?

Thanks!

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -20,6 +20,8 @@ use wire::{ArpPacket, ArpRepr, ArpOperation};
 use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
 #[cfg(feature = "proto-ipv6")]
 use wire::{Icmpv6Packet, Icmpv6Repr, Icmpv6ParamProblem};
+#[cfg(all(feature = "socket-icmp", any(feature = "proto-ipv4", feature = "proto-ipv6")))]
+use wire::IcmpRepr;
 #[cfg(feature = "proto-ipv6")]
 use wire::{NdiscNeighborFlags, NdiscRepr};
 #[cfg(all(feature = "proto-ipv6", feature = "socket-udp"))]
@@ -32,7 +34,7 @@ use wire::{TcpPacket, TcpRepr, TcpControl};
 use socket::{Socket, SocketSet, AnySocket};
 #[cfg(feature = "socket-raw")]
 use socket::RawSocket;
-#[cfg(all(feature = "socket-icmp", feature = "proto-ipv4"))]
+#[cfg(all(feature = "socket-icmp", any(feature = "proto-ipv4", feature = "proto-ipv6")))]
 use socket::IcmpSocket;
 #[cfg(feature = "socket-udp")]
 use socket::UdpSocket;
@@ -424,13 +426,16 @@ impl<'b, 'c, DeviceT> Interface<'b, 'c, DeviceT>
                     Socket::Raw(ref mut socket) =>
                         socket.dispatch(&caps.checksum, |response|
                             respond!(Packet::Raw(response))),
-                    #[cfg(all(feature = "socket-icmp", feature = "proto-ipv4"))]
+                    #[cfg(all(feature = "socket-icmp", any(feature = "proto-ipv4", feature = "proto-ipv6")))]
                     Socket::Icmp(ref mut socket) =>
                         socket.dispatch(&caps, |response| {
                             match response {
                                 #[cfg(feature = "proto-ipv4")]
-                                (IpRepr::Ipv4(ipv4_repr), icmpv4_repr) =>
+                                (IpRepr::Ipv4(ipv4_repr), IcmpRepr::Ipv4(icmpv4_repr)) =>
                                     respond!(Packet::Icmpv4((ipv4_repr, icmpv4_repr))),
+                                #[cfg(feature = "proto-ipv6")]
+                                (IpRepr::Ipv6(ipv6_repr), IcmpRepr::Ipv6(icmpv6_repr)) =>
+                                    respond!(Packet::Icmpv6((ipv6_repr, icmpv6_repr))),
                                 _ => Err(Error::Unaddressable)
                             }
                         }),
@@ -746,6 +751,23 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         let icmp_repr = Icmpv6Repr::parse(&ip_repr.src_addr(), &ip_repr.dst_addr(),
                                           &icmp_packet, &checksum_caps)?;
 
+        #[cfg(feature = "socket-icmp")]
+        let mut handled_by_icmp_socket = false;
+
+        #[cfg(all(feature = "socket-icmp", feature = "proto-ipv6"))]
+        for mut icmp_socket in _sockets.iter_mut().filter_map(IcmpSocket::downcast) {
+            if !icmp_socket.accepts(&ip_repr, &icmp_repr.into(), &checksum_caps) { continue }
+
+            match icmp_socket.process(&ip_repr, &icmp_repr.into(), &checksum_caps) {
+                // The packet is valid and handled by socket.
+                Ok(()) => handled_by_icmp_socket = true,
+                // The socket buffer is full.
+                Err(Error::Exhausted) => (),
+                // ICMP sockets don't validate the packets in any way.
+                Err(_) => unreachable!(),
+            }
+        }
+
         match icmp_repr {
             // Respond to echo requests.
             Icmpv6Repr::EchoRequest { ident, seq_no, data } => {
@@ -770,6 +792,11 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
                 IpRepr::Ipv6(ipv6_repr) => self.process_ndisc(timestamp, ipv6_repr, repr),
                 _ => Ok(Packet::None)
             },
+
+            // Don't report an error if a packet with unknown type
+            // has been handled by an ICMP socket
+            #[cfg(feature = "socket-icmp")]
+            _ if handled_by_icmp_socket => Ok(Packet::None),
 
             // FIXME: do something correct here?
             _ => Err(Error::Unrecognized),
@@ -839,9 +866,9 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
 
         #[cfg(all(feature = "socket-icmp", feature = "proto-ipv4"))]
         for mut icmp_socket in _sockets.iter_mut().filter_map(IcmpSocket::downcast) {
-            if !icmp_socket.accepts(&ip_repr, &icmp_repr, &checksum_caps) { continue }
+            if !icmp_socket.accepts(&ip_repr, &icmp_repr.into(), &checksum_caps) { continue }
 
-            match icmp_socket.process(&ip_repr, &icmp_repr, &checksum_caps) {
+            match icmp_socket.process(&ip_repr, &icmp_repr.into(), &checksum_caps) {
                 // The packet is valid and handled by socket.
                 Ok(()) => handled_by_icmp_socket = true,
                 // The socket buffer is full.
@@ -853,6 +880,7 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
 
         match icmp_repr {
             // Respond to echo requests.
+            #[cfg(feature = "proto-ipv4")]
             Icmpv4Repr::EchoRequest { ident, seq_no, data } => {
                 let icmp_reply_repr = Icmpv4Repr::EchoReply {
                     ident:  ident,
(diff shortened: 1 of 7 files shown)


src/wire/mod.rs
```diff
@@ -174,6 +176,13 @@ pub use self::icmpv6::{Message as Icmpv6Message,
                        ParamProblem as Icmpv6ParamProblem,
                        Packet as Icmpv6Packet,
                        Repr as Icmpv6Repr};
+#[cfg(any(feature = "proto-ipv4", feature = "proto-ipv6"))]
+pub use self::icmp::{Message as IcmpMessage,
+                     DstUnreachable as IcmpDstUnreachable,
+                     TimeExceeded as IcmpTimeExceeded,
+                     ParamProblem as IcmpParamProblem,
+                     Packet as IcmpPacket,
+                     Repr as IcmpRepr};
```
whitequark: Are you planning to somehow unify the ICMPv4 and ICMPv6 `Message`, `DstUnreachable`, `TimeExceeded` and `ParamProblem` enums? This doesn't seem like the right way to go to me. Why is this necessary?

whitequark:
Other than the comment above, looks good!

src/iface/ethernet.rs
```diff
@@ -826,13 +831,13 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         packet
     }
 
-    #[cfg(feature = "proto-ipv4")]
-    fn process_icmpv4<'frame>(&self, _sockets: &mut SocketSet, ip_repr: IpRepr,
-                              ip_payload: &'frame [u8]) -> Result<Packet<'frame>>
+    #[cfg(any(feature = "proto-ipv4", feature = "proto-ipv6"))]
+    fn process_icmp<'frame>(&self, _sockets: &mut SocketSet, ip_repr: IpRepr,
```
dlrobertson: ICMPv6 packets will be processed by [`process_icmpv6`](https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs#L741). It has a different [IP protocol number](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml).

src/wire/mod.rs
```diff
@@ -174,6 +176,13 @@ pub use self::icmpv6::{Message as Icmpv6Message,
                        ParamProblem as Icmpv6ParamProblem,
                        Packet as Icmpv6Packet,
                        Repr as Icmpv6Repr};
+#[cfg(any(feature = "proto-ipv4", feature = "proto-ipv6"))]
```
dlrobertson: Same

src/wire/mod.rs
```diff
@@ -95,6 +95,8 @@ mod ipv6fragment;
 mod icmpv4;
 #[cfg(feature = "proto-ipv6")]
 mod icmpv6;
+#[cfg(all(feature = "proto-ipv4", feature = "proto-ipv6"))]
```
dlrobertson: Rely on the `icmpv6` module for `proto-ipv6`.

src/wire/mod.rs
ProgVal: Indeed, it isn't

src/wire/mod.rs
dlrobertson: The comment by @whitequark [here](https://github.com/m-labs/smoltcp/pull/205/files#r187041666) was kind of what I was getting at and is a much clearer. If we get an ICMPv6 response we shouldn't be using the definitions in `wire::icmpv4`.

Note: my assumption was that `wire::icmp` was the same thing as `wire::icmpv4`. If your intent was to make `wire::icmp` a module that encapsulates ICMPv4 and ICMPv4 (similar to `wire::ip` for IPv4 and IPv6), then your current approach is correct.

whitequark:
OK, now that I see `icmp.rs`, I think this is definitely not the right approach. `IcmpRepr` makes sense to have, but the other enums aren't used anywhere and it's not clear to me that they are clearly useful as defined.

ProgVal:
Yeah, I started by implementing all the enums before looking at the existing code. I'll remove them

src/wire/mod.rs
ProgVal: Hmm, except that it should be `any()` and not `all()`

dlrobertson: True.

ProgVal:
I'm getting three test failures:

```
---- socket::icmp::test::test_accepts_udp_v6 stdout ----
	thread 'socket::icmp::test::test_accepts_udp_v6' panicked at 'assertion failed: `(left == right)`
  left: `Err(Truncated)`,
 right: `Ok(())`', src/socket/icmp.rs:745:9

---- socket::icmp::test::test_send_dispatch_v6 stdout ----
	thread 'socket::icmp::test::test_send_dispatch_v6' panicked at 'assertion failed: `(left == right)`
  left: `Err(Checksum)`,
 right: `Err(Unaddressable)`', src/socket/icmp.rs:499:9

---- socket::icmp::test::test_set_hop_limit_v6 stdout ----
	thread 'socket::icmp::test::test_set_hop_limit_v6' panicked at 'assertion failed: `(left == right)`
  left: `Err(Checksum)`,
 right: `Ok(())`', src/socket/icmp.rs:552:9
```

My guess is that the two latter are because the checksum is never computed (which is not an issue for ICMPv4 tests, because ICMPv4 checksums can be blank).

I tried solving them, with no success. Any idea?

whitequark:
> ---- socket::icmp::test::test_accepts_udp_v6 stdout ----
> 	thread 'socket::icmp::test::test_accepts_udp_v6' panicked at 'assertion failed: `(left == right)`
>   left: `Err(Truncated)`,
>  right: `Ok(())`', src/socket/icmp.rs:745:9

Do you know why this results in a Truncated error?

ProgVal:
No. I added a `println!` in front of every initialization of `Truncated` I could find in ICMP- and IPv6-related files, none of them were called.

src/iface/ethernet.rs
```diff
@@ -828,11 +833,11 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
 
     #[cfg(feature = "proto-ipv4")]
     fn process_icmpv4<'frame>(&self, _sockets: &mut SocketSet, ip_repr: IpRepr,
-                              ip_payload: &'frame [u8]) -> Result<Packet<'frame>>
+                            ip_payload: &'frame [u8]) -> Result<Packet<'frame>>
```
dlrobertson: nit: whitespace

src/iface/ethernet.rs
```diff
@@ -828,11 +833,11 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
     {
         let icmp_packet = Icmpv4Packet::new_checked(ip_payload)?;
         let checksum_caps = self.device_capabilities.checksum.clone();
-        let icmp_repr = Icmpv4Repr::parse(&icmp_packet, &checksum_caps)?;
+        let icmp_repr = Icmpv4Repr::parse(&icmp_packet, &checksum_caps)?.into();
```
dlrobertson: The new `IcmpRepr` really shouldn't be used in the `process_icmpv4` or `process_icmpv6` functions. AFAIK it is really only needed for getting data from the socket to the interface, but I don't think it is needed for parsing data from the outside world.

src/socket/icmp.rs
```diff
@@ -297,7 +313,20 @@ impl<'a, 'b> IcmpSocket<'a, 'b> {
                         payload_len: repr.buffer_len(),
                         hop_limit:   hop_limit,
                     });
-                    emit((ip_repr, repr))
+                    emit((ip_repr, IcmpRepr::Ipv4(repr)))
+                },
+                IpAddress::Ipv6(ipv6_addr) => {
+                    let packet = Icmpv6Packet::new(&*packet_buf);
+                    let src_addr = Ipv6Address::default();
+                    let repr = Icmpv6Repr::parse(&src_addr.into(), &ipv6_addr.into(), &packet, &ChecksumCapabilities::default())?;
```
dlrobertson: Should we expect users to fill in the checksum? Or should we just parse it with `ChecksumCapabilities::ignored()`?

whitequark: We shouldn't expect the users to fill in checksum.

dlrobertson: @ProgVal you might update the corresponding line for ICMPv4 as well https://github.com/m-labs/smoltcp/pull/205/files#diff-7dff2f1b7b74bbfcd78f956578d349d1R308

ProgVal: I don't really understand what `ChecksumCapabilities` mean (should the user of smoltcp verify them? or smoltcp itself? or the hardware?), but using `ChecksumCapabilities::ignored` fixed two of the issues, thanks!

src/iface/ethernet.rs
ProgVal: It's needed for `IcmpSocket::send`.

ProgVal:
Ha, I found why it's truncated, it's because the buffer allocated on the first line of `test_accepts_udp_v6` only supports 46-bytes packets.

src/iface/ethernet.rs
dlrobertson: yeah this is more or less a nit. It might be cleaner to add a line like `let socket_repr = IcmpRepr::from(icmp_repr)` then use `socket_repr` for the `IcmpSocket` functions and then you can still `match icmp_repr` and you don't have the extra `IcmpRepr::Ipv4(...)`, but this is just a nit.

src/socket/icmp.rs
dlrobertson: It is essentially used to determine if the parse/emit function should verify/fill the checksum. It is typically used in the context of a device. For example, a device that we can offload IPv4 checksumming on transmit to would have a `ChecksumCapabilities` structure with `ipv4` equal to [TX](https://github.com/m-labs/smoltcp/blob/master/src/phy/mod.rs#L127). However, in this case we don't want to evaluate the checksum so using `ChecksumCapabilities::ignored` makes that happen.

src/iface/ethernet.rs
dlrobertson: You also need to add similar code to `process_icmpv6` to ensure any ICMPv6 sockets actually receive data.

src/wire/icmp.rs
```diff
@@ -0,0 +1,58 @@
...
+                $class::Ipv4(s)
+            }
+        }
+        impl From<icmpv6::$class> for $class {
+            fn from(s: icmpv6::$class) -> Self {
+                $class::Ipv6(s)
+            }
+        }
+    };
+    ( $class:ident < $typearg:tt > ) => {
```
dlrobertson: This is the only matching arm that is used right? Or am I missing something?

ProgVal: Yes, now it is. I think I'll remove the macro altogether, as there is only one class left.

ProgVal:
(I just squashed some commits to make the git history cleaner)

Looks like everything is working now, thanks for your help!

src/socket/icmp.rs
```diff
@@ -244,13 +248,23 @@ impl<'a, 'b> IcmpSocket<'a, 'b> {
...
+            #[cfg(feature = "proto-ipv4")]
+            (&Endpoint::Udp(endpoint), &IcmpRepr::Ipv4(Icmpv4Repr::DstUnreachable { data, .. }))
+                    if endpoint.addr.is_unspecified() || endpoint.addr == ip_repr.dst_addr() => {
+                let packet = UdpPacket::new(data);
+                match UdpRepr::parse(&packet, &ip_repr.src_addr(), &ip_repr.dst_addr(), cksum) {
+                    Ok(repr) => endpoint.port == repr.src_port,
+                    Err(_) => false,
+                }
+            }
+            #[cfg(feature = "proto-ipv6")]
```
whitequark: I feel like these two cases ought to be unified, but not sure if this can be done cleanly.

whitequark:
Looks good! Please squash all commits. After that, r=me.

src/socket/icmp.rs
ProgVal: I could add a function to factorize the code, but I'm not sure it improves readability

whitequark: I was thinking of something like
```rust
match x {
  #[cfg(foo)] Y | #[cfg(bar)] Z => t
}
```
but I don't know if it works.

ProgVal: It doesn't, I tried

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 6eb23fa has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 6eb23fa73302f04ba371ee049c597cac84fa1f66 with merge ce2fbb7dd2a45d27426db3f842a2f5fa68f7b700...

dlrobertson:
@ProgVal thanks for working on this!

PR #212, opened 2018-05-14
iface: Begin parsing Hop-By-Hop extension headers

dlrobertson:
Begin add support for parsing extension headers.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -1,5 +1,6 @@
 // Heads up! Before working on this file you should read the parts
-// of RFC 1122 that discuss Ethernet, ARP and IP.
+// of RFC 1122 that discuss Ethernet, ARP and IP for any IPv4 work
+// and RFCs 8200 and 4861 for any IPv6 and NDISC work.
 
 use core::cmp;
 use managed::ManagedSlice;
@@ -22,6 +23,10 @@ use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
 use wire::{Icmpv6Packet, Icmpv6Repr, Icmpv6ParamProblem};
 #[cfg(all(feature = "socket-icmp", any(feature = "proto-ipv4", feature = "proto-ipv6")))]
 use wire::IcmpRepr;
+#[cfg(feature = "proto-ipv6")]
+use wire::{Ipv6HopByHopHeader, Ipv6HopByHopRepr};
+#[cfg(feature = "proto-ipv6")]
+use wire::{Ipv6OptionRepr, Ipv6OptionFailureType};
 #[cfg(feature = "proto-ipv6")]
 use wire::{NdiscNeighborFlags, NdiscRepr};
 #[cfg(all(feature = "proto-ipv6", feature = "socket-udp"))]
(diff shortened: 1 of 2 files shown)


src/iface/ethernet.rs
```diff
@@ -853,6 +875,44 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+    fn process_hopbyhop<'frame>(&mut self, sockets: &mut SocketSet, timestamp: Instant,
+                                ipv6_repr: Ipv6Repr, handled_by_raw_socket: bool,
+                                ip_payload: &'frame [u8]) -> Result<Packet<'frame>>
+    {
+        let hbh_pkt = Ipv6HopByHopHeader::new_checked(ip_payload)?;
+        let hbh_repr = Ipv6HopByHopRepr::parse(&hbh_pkt)?;
+        let buffer_len = hbh_repr.buffer_len();
+
+        let mut i = 2;
+        while i < buffer_len {
```
dlrobertson: I don't like this.

src/iface/ethernet.rs
```diff
@@ -853,6 +875,44 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
...
+                Ipv6OptionRepr::Pad1 | Ipv6OptionRepr::PadN(_) => (),
+                _ => match Ipv6OptionFailureType::from(opt_pkt.option_type()) {
+                    Ipv6OptionFailureType::Skip => (),
+                    Ipv6OptionFailureType::Discard => {
+                        return Ok(Packet::None);
+                    },
+                    _ => {
+                        // FIXME(dlrobertson): Send an ICMPv6 parameter problem message
+                        // here.
+                        return Err(Error::Unrecognized);
```
dlrobertson: Generating a correct Parameter Problem message we also need to keep track of the current offset.

src/iface/ethernet.rs
```diff
@@ -853,6 +875,44 @@ impl<'b, 'c> InterfaceInner<'b, 'c> {
         packet
     }
 
+    #[cfg(feature = "proto-ipv6")]
+    fn process_hopbyhop<'frame>(&mut self, sockets: &mut SocketSet, timestamp: Instant,
```
dlrobertson: Basically all we do here is skip the hop-by-hop header.

src/iface/ethernet.rs
whitequark: How about making an `Ipv6OptionIterator`? Note that I'm not hung on this, and would be OK merging as is, but if you want, this would be how I improve it.

dlrobertson: :+1: yeah, I think that would be better. Should be pretty easy to implement.

dlrobertson:
Added the iterator and a fixed up a few other things

src/iface/ethernet.rs
```diff
@@ -1,5 +1,6 @@
 // Heads up! Before working on this file you should read the parts
-// of RFC 1122 that discuss Ethernet, ARP and IP.
+// of RFC 1122 that discuss Ethernet, ARP and IP for any IPv4 work
+// and RFCs 8200 and 4861 for and IPv6 and NDISC work.
```
whitequark: `for and`

dlrobertson: Fixed

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 1e6157b has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 1e6157ba97a18266c54e9e4c89ad1201b264a452 with merge 2363f97834b3d529c6344f1840ab001645e853d2...

Issue #104, opened 2017-12-25
Implement TCP Fast Retransmit


whitequark:
Right now we'll be blasting the entirety of our buffer before even considering retransmitting, which is silly.

barskern:
I would like to take a stab at implementing this feature. I have been programming in Rust for some months now, and I've taken a subject in University about networks and protocols. I've started to read through the code, but I was wondering if I could get a helping indication on where to start.

whitequark:
There's a comment that says `// FIXME: implement fast retransmit`, that would be a good start :)

barskern:
I've been reading the documentation, and started to dig into the code. Today I've been trying to write a test to isolate a situation where a fast retransmit would occur, however it doesnt seem to be right. Currently the test looks like this:

```rust
   #[test]
    fn test_fast_retransmit_after_triple_duplicate_ack() {
        let mut s = socket_established();
        s.remote_win_len = 6;
        
        s.send_slice(b"abcdef123456ABCDEFZYXWVU").unwrap();

        // This packet is supposed to not be acknowledged (recived)
        recv!(s, time 1000, Ok(TcpRepr {
            seq_number: LOCAL_SEQ + 1,
            ack_number: Some(REMOTE_SEQ + 1),
            payload:    &b"abcdef"[..],
            ..RECV_TEMPL
        }));
        recv!(s, time 1005, Ok(TcpRepr {
            seq_number: LOCAL_SEQ + 1 + 6,
            ack_number: Some(REMOTE_SEQ + 1),
            payload:    &b"123456"[..],
            ..RECV_TEMPL
        }));
        recv!(s, time 1010, Ok(TcpRepr {
            seq_number: LOCAL_SEQ + 1 + 6 + 6,
            ack_number: Some(REMOTE_SEQ + 1),
            payload:    &b"ABCDEF"[..],
            ..RECV_TEMPL
        }));
        recv!(s, time 1015, Ok(TcpRepr {
            seq_number: LOCAL_SEQ + 1 + 6 + 6 + 6,
            ack_number: Some(REMOTE_SEQ + 1),
            payload:    &b"ZYXWVU"[..],
            ..RECV_TEMPL
        }));
        // First ACK
        send!(s, time 1020, TcpRepr {
          seq_number: REMOTE_SEQ + 1,
          ack_number: Some(LOCAL_SEQ + 1),
          window_len: 6,
          ..SEND_TEMPL
        });
        // Second (duplicate) ACK
[...]

PR #214, opened 2018-05-15
Implement TCP Fast Retransmit

barskern:
Implement TCP Fast Retransmit by simply keeping track of the **last recived ACK** (`local_rx_last_ack`) and the **number of duplicate ACKs** (`local_rx_dup_acks`) recived. When a third duplicate ACK is encountered, the `set_for_fast_retransmit` is triggered. This will make the socket send the unacknowledged packet without a delay.

A building block of this PR is an implementation of `Timer::FastRetransmit`, which provides delay-less retransmit. This was needed because `Timer::Retransmit` included a delay which is uneccessary when we recived three duplicate ACKs.

## Drawbacks

This solution is rather simple and adds some "clutter" to the `TcpSocket`. I think that preferably _TCP flow control_ would be separated into a separate state which would be a single field on the `TcpSocket`.

Closes: #104

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -59,6 +59,7 @@ enum Timer {
         expires_at: Instant,
         delay:      Duration
     },
+    FastRetransmit,
     Close {
         expires_at: Instant
     }
@@ -89,7 +90,8 @@ impl Timer {
             Timer::Retransmit { expires_at, delay }
                     if timestamp >= expires_at => {
                 Some(timestamp - expires_at + delay)
-            }
+            },
+            Timer::FastRetransmit => Some(Duration::from_millis(0)),
             _ => None
         }
     }
@@ -109,6 +111,7 @@ impl Timer {
             Timer::Idle { keep_alive_at: Some(keep_alive_at) } => PollAt::Time(keep_alive_at),
             Timer::Idle { keep_alive_at: None } => PollAt::Ingress,
             Timer::Retransmit { expires_at, .. } => PollAt::Time(expires_at),
+            Timer::FastRetransmit => PollAt::Now,
             Timer::Close { expires_at } => PollAt::Time(expires_at),
         }
     }
@@ -154,8 +157,13 @@ impl Timer {
                 }
             }
             Timer::Retransmit { .. } => (),
+            Timer::FastRetransmit { .. } => (),
             Timer::Close { .. } => ()
         }
+    }
+
+    fn set_for_fast_retransmit(&mut self) {
+        *self = Timer::FastRetransmit
     }
 
     fn set_for_close(&mut self, timestamp: Instant) {
@@ -166,7 +174,7 @@ impl Timer {
 
     fn is_retransmit(&self) -> bool {
         match *self {
-            Timer::Retransmit {..} => true,
+            Timer::Retransmit {..} | Timer::FastRetransmit => true,
             _ => false,
         }
     }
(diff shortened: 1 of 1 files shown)


src/socket/tcp.rs
```diff
@@ -58,6 +58,9 @@ enum Timer {
     Retransmit {
         expires_at: Instant,
         delay:      Duration
+    }, 
+    FastRetransmit {
+        expires_at: Instant
```
whitequark: I think FastRetransmit doesn't need `expires_at`, since it always fires immediately.

barskern: Would returning `None` in `poll_at` imply that we want to be polled immidiately?
```rust
fn poll_at(&self) -> Option<Instant> {
    match *self {
        Timer::Idle { keep_alive_at } => keep_alive_at,
        Timer::Retransmit { expires_at, .. } => Some(expires_at),
        Timer::FastRetransmit => None,
        Timer::Close { expires_at } => Some(expires_at),
    }
}
```
https://github.com/barskern/smoltcp/blob/3d4a881b5d4129661a00d32747d6825c80531dc5/src/socket/tcp.rs#L111-L118

whitequark: No, that's an accident. The return value for "polled immediately" is `Some(Instant::from_millis(0))`. You could actually put that in `Retransmit { expires_at }`, but `FastRetransmit` is definitely cleaner.

barskern: But isn't the return value from `poll_at` supposed to be an absolute value? So we're kinda returning that it should have been polled in the future?

pothos: I'm not a reviewer, but I think I can answer your question. The return value of `poll_at` lands in `poll_delay` which is used for `wait`. `None` means to block until a packet arrives but a timeout of 0 will return immediately. `poll_delay` will do this for all return values of `poll_at` that are older than the current time given to `poll_delay`. To be sure I think you should just return 0 here as well.

dlrobertson: See the implementation for UDP socket [here]. `Some(Instant::from_millis(0))` would be some time in the past, which means poll now AFAIK.

[here]: https://github.com/m-labs/smoltcp/blob/master/src/socket/udp.rs#L215-L221

pothos: (Oh, wrote my answer before all the other comments but did not see them.)

whitequark: Do you think you can make a PR doing that?

barskern: It's on the way. I'll wait with implementing the rest of this PR until the new `poll_at` is defined.

barskern: I was thinking it would be appropriate to discuss the signature hence started #215.

m-labs-homu:
:umbrella: The latest upstream changes (presumably 3fb0c22fd4edcecd75f4c6767a86b0333cac9366) made this pull request unmergeable. Please resolve the merge conflicts.

barskern:
Yes will do. I normally do for my own projects I just wanted to keep history mostly intact when there are more people involved. Isn't it easier to review when there are multiple commits or should I just always squash so there is just one?

barskern:
Updated the docstrings for `PollAt` aswell (as mentionned in the review), hence the two commits.

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 17de43d has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 17de43d3eb373c3c5bcdd14e0a98aff83fc3c488 with merge 375b0266fe4a3d92ce685d4695636be63cc61d5f...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/380301960?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 375b0266fe4a3d92ce685d4695636be63cc61d5f to master...

PR #218, opened 2018-05-22
Add IPv6 Extension Routing Header

### LINKED ISSUE #160: Design: Implement IPv6 Extension Routing Header
The Routing Header spec is at https://tools.ietf.org/html/rfc8200#section-4.4

Two questions about this:

1. What routing types should we implement? IANA routing types are at https://www.iana.org/assignments/ipv6-parameters/ipv6-parameters.xhtml#ipv6-parameters-3.

* Type 0 (aka Source Route) was defined in the original IPv6 RFC but has since been deprecated. Should we implement this route?
* Type 1 (aka `Nimrod`) looks to be abandoned, so that is clear.
* Type 2 is related to IPv6 Mobility https://tools.ietf.org/html/rfc6275#section-6.4. I think this type only makes sense to implement if we implement the rest of [RFC 6275](https://tools.ietf.org/html/rfc6275).
* Type 3 (aka RPL Source Route) is defined in [RFC 6554](https://tools.ietf.org/html/rfc6554) and is the successor to Type 0.
* There are two experiment types defined in [RFC 4727 - Section 3.6](https://tools.ietf.org/html/rfc4727#section-3.6). There is no definitive format specified for these, as they are meant to be used for experiments.

2. How to represent the information for each routing type?

The [format](https://tools.ietf.org/html/rfc8200#section-4.4) of the routing type contains 4 common fields and an unspecified section for _type-specific data_ of variable length.

* Should `Header#check_len` only ensure that there is enough data for the first 4 fields?
* How should we deal with the different methods on each type? For example, there should be a `Header#home_address` method for a Type 3 header, but not one for a Type 2 header.
* How should the `Repr` look? I can imagine something like:

```rust
pub struct Repr {
    pub next_header: Protocol,
    pub header_len: u8,
    pub routing_type: u8,
    pub segments_left: u8,
    pub type_spec_data: TypeSpecData,
}

pub enum TypeSpecData {
    Type2 { home_address: Ipv6Address },
    Type3 { cmpr_i: u8, cmpr_e: u8, pad: u8 } // would need to add a vector of addresses too
}
```

hjr3:
Fixes #160

### DIFF
--- a/src/wire/ipv6routing.rs
+++ b/src/wire/ipv6routing.rs
@@ -1,8 +1,8 @@
 use core::fmt;
 use {Error, Result};
 
-pub use super::IpProtocol as Protocol;
-pub use super::Ipv6Address as Address;
+use super::IpProtocol as Protocol;
+use super::Ipv6Address as Address;
 
 enum_with_unknown! {
     /// IPv6 Extension Routing Header Routing Type
@@ -315,7 +315,7 @@ impl<T: AsRef<[u8]> + AsMut<[u8]>> Header<T> {
                 data[7] = 0;
             }
 
-            _ => {}
+            _ => panic!("Unrecognized routing type when clearing reserved fields.")
         }
     }
 }


hjr3:
@dlrobertson PR is up for IPv6 routing.

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,664 @@
...
+
+pub use super::IpProtocol as Protocol;
+pub use super::Ipv6Address as Address;
+
+enum_with_unknown! {
+    /// IPv6 Extension Routing Header Routing Type
+    pub doc enum Type(u8) {
+        /// Source Route (DEPRECATED)
+        ///
+        /// See http://www.iana.org/go/rfc5095 for details.
```
dlrobertson: Use the ietf. The iana links just 302 to the ietf. https://tools.ietf.org/html/rfc5095

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,664 @@
...
+        Type0 = 0,
+        /// Nimrod (DEPRECATED 2009-05-06)
+        Nimrod = 1,
+        /// Type 2 Routing Header for Mobile IPv6
+        ///
+        /// See https://tools.ietf.org/html/rfc6275#section-6.4 for details.
+        Type2 = 2,
+        /// RPL Source Routing Header
+        ///
+        /// See http://www.iana.org/go/rfc6554 for details.
```
dlrobertson: https://tools.ietf.org/html/rfc6554

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,664 @@
...
+        Rpl = 3,
+        /// RFC3692-style Experiment 1
+        ///
+        /// See http://www.iana.org/go/rfc4727 for details.
```
dlrobertson: https://tools.ietf.org/html/rfc4727

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,664 @@
...
+        Experiment1 = 253,
+        /// RFC3692-style Experiment 2
+        ///
+        /// See http://www.iana.org/go/rfc4727 for details.
```
dlrobertson: https://tools.ietf.org/html/rfc4727

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,664 @@
...
+    }
+
+    /// Return the segments left field.
+    #[inline]
+    pub fn segments_left(&self) -> u8 {
+        let data = self.buffer.as_ref();
+        data[field::SEG_LEFT]
+    }
+
+    /// Return the IPv6 Home Address
```
dlrobertson: Could type specific packet methods be in separate documented `impl` blocks. See `ndisc.rs`.

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,664 @@
...
+    pub fn addresses(&self) -> &[u8] {
+        let data = self.buffer.as_ref();
+        &data[field::ADDRESSES(data[field::LENGTH])]
+    }
+}
+
+impl<'a, T: AsRef<[u8]> + ?Sized> Header<&'a T> {
+    /// Return type specific data
+    #[inline]
+    pub fn data(&self) -> &'a[u8] {
```
dlrobertson: Is this really used? Type 2 uses `home_address` and RPL uses `address`.

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,664 @@
...
+                data[7] = 0;
+            }
+            Type::Rpl => {
+                // Retain the higher order 4 bits of the padding field
+                data[field::PAD] = data[field::PAD] & 0xF0;
+                data[6] = 0;
+                data[7] = 0;
+            }
+
+            _ => {}
```
dlrobertson: explicit panic if `routing_type` is not set?

src/wire/ipv6routing.rs
hjr3: fixed

src/wire/ipv6routing.rs
hjr3: fixed

src/wire/ipv6routing.rs
hjr3: fixed

src/wire/ipv6routing.rs
hjr3: fixed

src/wire/ipv6routing.rs
hjr3: I had this but thought we were moving away from it with only panic comments. Done.

src/wire/ipv6routing.rs
hjr3: It was used in a test. I removed `data` and `data_mut` and modified the test.

src/wire/ipv6routing.rs
hjr3: Should I panic on a value of `0` then? That is a valid routing type, though not one this code explicitly supports.

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,658 @@
+use core::fmt;
+use {Error, Result};
+
+pub use super::IpProtocol as Protocol;
```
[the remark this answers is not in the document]
hjr3: I forgot why we were doing this. It is in other extension headers as well: https://github.com/m-labs/smoltcp/blob/master/src/wire/ipv6fragment.rs#L6 and https://github.com/m-labs/smoltcp/blob/master/src/wire/ipv6hopbyhop.rs#L4

I have removed `pub` from the imports in this PR.

src/wire/ipv6routing.rs
hjr3: Makes sense. Done.

src/wire/ipv6routing.rs
```diff
@@ -0,0 +1,658 @@
...
+/// A high-level representation of an IPv6 Routing Header.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Repr<'a> {
+    Type2 {
+        /// The type of header immediately following the Routing header.
+        next_header:    Protocol,
+        /// Length of the Routing header in 8-octet units, not including the first 8 octets.
+        length:         u8,
+        /// Number of route segments remaining.
+        segments_left:  u8,
```
dlrobertson: This repr doesn't need `length` or `segments_left` they will always be 2 and 1 respectively. It is an invalid packet if they are not.

hjr3: I went back and forth on this. I decided to leave them explicit as I imagined that downstream code would not want these values implicit. If I remove those fields, should I provide methods on the Repr that return the number of length and number of segments left?

I have thought about this validity problem a lot, but thought validation would be handled outside of the wire modules. I have not validated the values in the Repr (for fragmentation, hop by hop and now routing) are all valid with respect to each other. For example, I do not check that the `cmpr_i` value is valid based on the vector of addresses being set.

dlrobertson: > I decided to leave them explicit as I imagined that downstream code would not want these values implicit.

I suppose it could cut down on magic numbers in downstream code when parsing a Type 2 routing header. No harm in leaving it in for now. We can always create an issue to discuss it further.

> I have thought about this validity problem a lot, but thought validation would be handled outside of the wire modules.

I'm not suggesting validating the packet. I was just enforcing that we know a Type 2 headers effective length is 2.

src/wire/ipv6routing.rs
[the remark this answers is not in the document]
dlrobertson: The other instances of `pub use super::IpProtocol as Protocol` also do not need to be `pub`.

dlrobertson:
@m-labs-homu r=whitequark

m-labs-homu:
:pushpin: Commit 5298144 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 52981446b91011d2614ce433d8d7f6d864f52d59 with merge 9c1900ff3f66196f659cbfc79213dc944196db9d...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/382389426?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 9c1900ff3f66196f659cbfc79213dc944196db9d to master...

PR #219, opened 2018-05-26
 Add support for arbitrarily many routes instead of only gateways. 

### LINKED ISSUE #209: RFC: multiple routes
As asked in #207, I'm opening an RFC issue about multiple routes.

I created a naive implementation of routes [here](https://github.com/m-labs/smoltcp/compare/master...ProgVal:naive-routes), based on simply iterating over a managed slice.

Of course, there are probably more efficient ways to do it. Any thoughts?

progval:
Closes #209.

Note: As this PR uses the newly introduced `managed::ManagedMap::range` method, I changed Cargo.toml to use `managed`'s git repo instead of the released version.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -11,7 +11,7 @@ use std::collections::BTreeMap;
 use std::os::unix::io::AsRawFd;
 use smoltcp::phy::wait as phy_wait;
 use smoltcp::wire::{EthernetAddress, Ipv4Address, IpAddress, IpCidr};
-use smoltcp::iface::{NeighborCache, EthernetInterfaceBuilder};
+use smoltcp::iface::{NeighborCache, EthernetInterfaceBuilder, Routes};
 use smoltcp::socket::{SocketSet, TcpSocket, TcpSocketBuffer};
 use smoltcp::time::Instant;
 
@@ -40,11 +40,14 @@ fn main() {
     let ethernet_addr = EthernetAddress([0x02, 0x00, 0x00, 0x00, 0x00, 0x02]);
     let ip_addrs = [IpCidr::new(IpAddress::v4(192, 168, 69, 2), 24)];
     let default_v4_gw = Ipv4Address::new(192, 168, 69, 100);
+    let mut routes_storage = [None; 1];
+    let mut routes = Routes::new(&mut routes_storage[..]);
+    routes.add_default_ipv4_route(default_v4_gw).unwrap();
     let mut iface = EthernetInterfaceBuilder::new(device)
             .ethernet_addr(ethernet_addr)
             .neighbor_cache(neighbor_cache)
             .ip_addrs(ip_addrs)
-            .ipv4_gateway(default_v4_gw)
+            .routes(routes)
             .finalize();
 
     let mut sockets = SocketSet::new(vec![]);
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -2,7 +2,7 @@
 // of RFC 1122 that discuss Ethernet, ARP and IP.
 
 use core::cmp;
-use managed::ManagedSlice;
+use managed::{ManagedSlice, ManagedMap};
 
 use {Error, Result};
 use phy::{Device, DeviceCapabilities, RxToken, TxToken};
@@ -13,7 +13,7 @@ use wire::{IpAddress, IpProtocol, IpRepr, IpCidr};
 #[cfg(feature = "proto-ipv6")]
 use wire::{Ipv6Address, Ipv6Packet, Ipv6Repr, IPV6_MIN_MTU};
 #[cfg(feature = "proto-ipv4")]
-use wire::{Ipv4Address, Ipv4Packet, Ipv4Repr, IPV4_MIN_MTU};
+use wire::{Ipv4Packet, Ipv4Repr, IPV4_MIN_MTU};
 #[cfg(feature = "proto-ipv4")]
 use wire::{ArpPacket, ArpRepr, ArpOperation};
 #[cfg(feature = "proto-ipv4")]
@@ -41,15 +41,16 @@ use socket::UdpSocket;
 #[cfg(feature = "socket-tcp")]
 use socket::TcpSocket;
 use super::{NeighborCache, NeighborAnswer};
+use super::Routes;
 
 /// An Ethernet network interface.
 ///
 /// The network interface logically owns a number of other data structures; to avoid
 /// a dependency on heap allocation, it instead owns a `BorrowMut<[T]>`, which can be
 /// a `&mut [T]`, or `Vec<T>` if a heap is available.
-pub struct Interface<'b, 'c, DeviceT: for<'d> Device<'d>> {
+pub struct Interface<'b, 'c, 'e, DeviceT: for<'d> Device<'d>> {
     device: DeviceT,
-    inner:  InterfaceInner<'b, 'c>,
+    inner:  InterfaceInner<'b, 'c, 'e>,
 }
 
 /// The device independent part of an Ethernet network interface.
@@ -59,31 +60,25 @@ pub struct Interface<'b, 'c, DeviceT: for<'d> Device<'d>> {
 /// the `device` mutably until they're used, which makes it impossible to call other
 /// methods on the `Interface` in this time (since its `device` field is borrowed
 /// exclusively). However, it is still possible to call methods on its `inner` field.
-struct InterfaceInner<'b, 'c> {
+struct InterfaceInner<'b, 'c, 'e> {
     neighbor_cache:         NeighborCache<'b>,
     ethernet_addr:          EthernetAddress,
     ip_addrs:               ManagedSlice<'c, IpCidr>,
-    #[cfg(feature = "proto-ipv4")]
-    ipv4_gateway:           Option<Ipv4Address>,
-    #[cfg(feature = "proto-ipv6")]
-    ipv6_gateway:           Option<Ipv6Address>,
+    routes:                 Routes<'e>,
     device_capabilities:    DeviceCapabilities,
 }
 
 /// A builder structure used for creating a Ethernet network
 /// interface.
-pub struct InterfaceBuilder <'b, 'c, DeviceT: for<'d> Device<'d>> {
+pub struct InterfaceBuilder <'b, 'c, 'e, DeviceT: for<'d> Device<'d>> {
     device:              DeviceT,
     ethernet_addr:       Option<EthernetAddress>,
     neighbor_cache:      Option<NeighborCache<'b>>,
     ip_addrs:            ManagedSlice<'c, IpCidr>,
-    #[cfg(feature = "proto-ipv4")]
-    ipv4_gateway:        Option<Ipv4Address>,
-    #[cfg(feature = "proto-ipv6")]
-    ipv6_gateway:        Option<Ipv6Address>,
+    routes:              Routes<'e>,
 }
 
-impl<'b, 'c, DeviceT> InterfaceBuilder<'b, 'c, DeviceT>
+impl<'b, 'c, 'e, DeviceT> InterfaceBuilder<'b, 'c, 'e, DeviceT>
         where DeviceT: for<'d> Device<'d> {
     /// Create a builder used for creating a network interface using the
     /// given device and address.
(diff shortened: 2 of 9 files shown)


src/wire/ipv4.rs
```diff
@@ -110,8 +110,8 @@ impl fmt::Display for Address {
 /// subnet masking prefix length.
 #[derive(Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
 pub struct Cidr {
-    address:    Address,
-    prefix_len: u8,
+    pub(crate) address:    Address,
+    pub(crate) prefix_len: u8,
```
whitequark: Why is this necessary? There are accessors and constructors for these...

whitequark:
Seems very good at cursory check, and very useful functionality too!
@dlrobertson Can you do a detailed review please?
cc @batonius

src/wire/ipv4.rs
ProgVal: I need it in order to write this:

```
const CIDR_1: Ipv4Cidr = Ipv4Cidr {      
    address: Ipv4Address([192, 0, 2, 0]),
    prefix_len: 24,                      
};                                       
```

Using a constructor is forbidden in const expressions.

whitequark: Can you make it a function instead? We can make `new` `const fn` when they land. I feel like making a change this big just for tests isn't worth it.

ProgVal: Sure. Done.

src/iface/ethernet.rs
```diff
@@ -152,43 +144,23 @@ impl<'b, 'c, DeviceT> InterfaceBuilder<'b, 'c, DeviceT>
...
-        self.ipv4_gateway = Some(addr);
-        self
-    }
-
-    /// Set the IPv6 gateway the interface will use. See also
-    /// [ipv6_gateway].
-    ///
-    /// # Panics
-    /// This function panics if the given address is not unicast.
+    /// This function panics if any of the addresses are not unicast.
```
dlrobertson: I don't think this is true any more.

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// This function panics if any of the addresses are not unicast.
+    pub fn update_routes<F: FnOnce(&mut ManagedMap<'a, IpCidr, Route>)>(&mut self, f: F) {
+        f(&mut self.storage);
+    }
+
+    /// Add a default ipv4 gateway (ie. "ip route add 0.0.0.0/0 via `gateway`").
+    ///
+    /// On success, returns the previous default gateway, if any.
+    ///
+    /// If not enough storage, returns the gateway address.
```
dlrobertson: Typically `Error::Exhausted` is returned if not enough storage. Is three a reason to return the previous gateway addr on error?

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+            Ok(route) => Ok(route),
+            Err((_cidr, route)) => Err(route.via_router)
+        }
+    }
+
+    /// Add a default ipv6 gateway (ie. "ip -6 route add ::/0 via `gateway`").
+    ///
+    /// On success, returns the previous default gateway, if any.
+    ///
+    /// If not enough storage, returns the gateway address.
```
dlrobertson: Same.

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub fn new<T>(storage: T) -> Routes<'a>
+            where T: Into<ManagedMap<'a, IpCidr, Route>> {
+        let storage = storage.into();
+        Routes { storage }
+    }
+
+    /// Update the routes of this node.
+    ///
+    /// # Panics
+    /// This function panics if any of the addresses are not unicast.
```
dlrobertson: I don't believe this is true now either.

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+        let route = Route::new_ipv6_gateway(gateway);
+        match self.storage.insert(cidr, route) {
+            Ok(route) => Ok(route),
+            Err((_cidr, route)) => Err(route.via_router)
+        }
+    }
+
+    pub(crate) fn lookup(&self, addr: &IpAddress, timestamp: Instant) ->
+            Option<IpAddress> {
+        if addr.is_broadcast() {
```
dlrobertson: Probably shouldn't run `lookup` on multicast addresses either. I think this can be `is_unicast`.

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+            if route.via_router != *addr && prefix.contains_addr(addr) {
+                return Some(route.via_router);
+            }
+        }
+
+        None
+    }
+}
+
+#[cfg(all(test, feature = "proto-ipv4"))]
```
dlrobertson: These tests should run if either `proto-ipv4` or `proto-ipv6` are enabled. See `MOCK_IP_ADDR_*` in [wire/ip.rs:852-876](https://github.com/m-labs/smoltcp/blob/master/src/wire/ip.rs#L853-L876). You will need to add `MOCK_IP_CIDR_*`.

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+        let cidr = match addr {
+            #[cfg(feature = "proto-ipv4")]
+            IpAddress::Ipv4(addr) => IpCidr::Ipv4(Ipv4Cidr::new(*addr, 32)),
+            #[cfg(feature = "proto-ipv6")]
+            IpAddress::Ipv6(addr) => IpCidr::Ipv6(Ipv6Cidr::new(*addr, 128)),
+            _ => unimplemented!()
+        };
+
+        for (prefix, route) in self.storage.range((Bound::Unbounded, Bound::Included(cidr))).rev() {
+            // TODO: do something with route.preferred_until
```
dlrobertson: You might want to create a FIXME or issue to research a clever way to clear out expired routes.

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+            if let Some(expires_at) = route.expires_at {
+                if timestamp > expires_at {
+                    continue;
+                }
+            }
+
+            if route.via_router != *addr && prefix.contains_addr(addr) {
```
dlrobertson: I don't think this is necessary. If the address we're querying is the address of the router we either didn't run address resolution or it didn't respond. If `lookup` returns the address of the router we'll either used the cached link layer address or run address resolution.

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+    const ADDR_1A: Ipv4Address = Ipv4Address([192, 0, 2, 1]);
+    const ADDR_1B: Ipv4Address = Ipv4Address([192, 0, 2, 13]);
+    const ADDR_1C: Ipv4Address = Ipv4Address([192, 0, 2, 42]);
+    fn cidr_1() -> Ipv4Cidr {
+        Ipv4Cidr::new(Ipv4Address([192, 0, 2, 0]), 24)
+    }
+
+    const ADDR_2A: Ipv4Address = Ipv4Address([198, 51, 100, 1]);
+    const ADDR_2B: Ipv4Address = Ipv4Address([198, 51, 100, 21]);
+    fn cidr_2() -> Ipv4Cidr {
```
dlrobertson: What is the purpose of this test?

src/iface/route.rs
```diff
@@ -0,0 +1,207 @@
...
+/// On systems without heap, use:
+///
+/// ```rust
+/// use smoltcp::iface::Routes;
+/// let mut routes_storage = [];
+/// let mut routes = Routes::new(&mut routes_storage[..]);
+/// ```
+#[derive(Debug)]
+pub struct Routes<'a> {
+    storage:      ManagedMap<'a, IpCidr, Route>,
```
dlrobertson: whitespace

src/iface/ethernet.rs
ProgVal: Indeed.

src/iface/route.rs
ProgVal: I mimicked ManagedMap::range's return result. I'll change that.

src/iface/route.rs
ProgVal: What do you mean by "create a FIXME"?

src/iface/route.rs
ProgVal: will do

src/iface/route.rs
ProgVal: Which one?

src/iface/route.rs
dlrobertson: Add a comment with something like the following.

```
FIXME(ProgVal): Find a clever way to clear out expired routes.
```

I think an issue would be best though.

src/iface/ethernet.rs
```diff
@@ -59,31 +60,25 @@ pub struct Interface<'b, 'c, DeviceT: for<'d> Device<'d>> {
...
-struct InterfaceInner<'b, 'c> {
+struct InterfaceInner<'b, 'c, 'e> {
     neighbor_cache:         NeighborCache<'b>,
     ethernet_addr:          EthernetAddress,
     ip_addrs:               ManagedSlice<'c, IpCidr>,
-    #[cfg(feature = "proto-ipv4")]
-    ipv4_gateway:           Option<Ipv4Address>,
-    #[cfg(feature = "proto-ipv6")]
-    ipv6_gateway:           Option<Ipv6Address>,
+    routes:                 Routes<'e>,
```
dlrobertson: Is there a reason to not use an `Option` here? Then you don't have to run `lookup` if the user hasn't configured a buffer for routes.

dlrobertson:
One more small question, but it looks really good!

src/iface/ethernet.rs
ProgVal: Not really. Is it really worth it? `lookup` should exit very quickly if there are no routes.

src/iface/route.rs
ProgVal: Done: https://github.com/m-labs/smoltcp/issues/221

examples/client.rs
```diff
@@ -40,11 +40,14 @@ fn main() {
     let ethernet_addr = EthernetAddress([0x02, 0x00, 0x00, 0x00, 0x00, 0x02]);
     let ip_addrs = [IpCidr::new(IpAddress::v4(192, 168, 69, 2), 24)];
     let default_v4_gw = Ipv4Address::new(192, 168, 69, 100);
+    let mut routes_storage = [None];
```
whitequark: Nit: `[None; 1]` here and `[None; 2]` in the other example as a demonstration that you don't have to list every item explicitly.

src/iface/route.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// Creates a routing tables. The backing storage is **not** cleared
+    /// upon creation.
+    pub fn new<T>(storage: T) -> Routes<'a>
+            where T: Into<ManagedMap<'a, IpCidr, Route>> {
+        let storage = storage.into();
+        Routes { storage }
+    }
+
+    /// Update the routes of this node.
+    pub fn update_routes<F: FnOnce(&mut ManagedMap<'a, IpCidr, Route>)>(&mut self, f: F) {
```
whitequark: This should just be `update`.

src/iface/route.rs
```diff
@@ -0,0 +1,231 @@
...
+        f(&mut self.storage);
+    }
+
+    /// Add a default ipv4 gateway (ie. "ip route add 0.0.0.0/0 via `gateway`").
+    ///
+    /// On success, returns the previous default gateway, if any.
+    #[cfg(feature = "proto-ipv4")]
+    pub fn add_default_ipv4_gateway(&mut self, gateway: Ipv4Address) -> Result<Option<Route>> {
```
whitequark: This adds a default IPv4 *route* (i.e. a single route is added), not *gateway* (which would imply that other routes might use the same gateway if none is specified). Also docstring and IPv6 needs to be changed.

ProgVal:
Any idea why Travis suddenly started failing at building `managed`?

whitequark:
You can bump version of `managed` to `0.7.0`.

whitequark:
Sorry, I didn't notice you updated the PR. It looks good to me now. Can you squash the commits?

ProgVal:
Done!

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit f81d902 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit f81d9028ac9befe4cc880a658b51f5a36e50371f with merge 06d18130cd8b25edbcad207bbdbb157f1c1203df...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/390228449?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 06d18130cd8b25edbcad207bbdbb157f1c1203df to master...

Issue #224, opened 2018-06-01
Stuck at retransmitting


pothos:
This code here https://github.com/m-labs/smoltcp/blob/master/src/socket/tcp.rs#L913 should either reset `local_rx_dup_acks` or let it count from 255 to 0 with a overflowing add or stay at 255 with a saturation. What is the desired behavior?
```
                        self.local_rx_dup_acks += 1;
                        if self.local_rx_dup_acks == 3 {
                        }

```
`thread 'main' panicked at 'attempt to add with overflow', smoltcp/src/socket/tcp.rs:913:25`

pothos:
Actually master stopped working for me, I try to find out if related to the fast retrainsmit code.

whitequark:
cc @barskern

barskern:
@pothos Do you think you would be able to write a test which mimics this behavior? Or explain me on how this managed to happen so that I could write a test?

I would think that the desired behavior is to do a saturating addition, and change the condition from `if self.local_rx_dup_acks == 3` to `if self.local_rx_dup_acks >= 3`. This would trigger a fast retransmit for every segment which is a duplicate ACK when we have recived more than 3 duplicate ACKs. However I did not find documentation that indicates that this is appropriate behavior for a TCP-implementation.

barskern:
> Per my understanding, with the modified code, if e.g. first packet in a burst of 8 gets lost, the other 7 will arrive, and 4 of them will trigger retransmit of the entire burst. Is that correct?

This is the reason I previously used `== 3`. However if we recive a burst of pacakges and the socket is not polled until all 8 packages have recived, I think there will only be 1 actual retransmit. This is because the way the socket retransmits is by setting an immidiate timeout, not directly sending a packet. Or thats what I think.. Perhaps we could find a way to test this theory.

This is such a edge case though. What is the probablility that we lose a packet and then lose the exact same package again in the retransmit, when all the other packages arrived. I think this is why I can't find a document about it online.

barskern:
After some more research, I found that what should happen after a duplicate ACK is recived is that the socket should enter "Fast recovery" and every following duplicate ACK only increases the congestion window. 

> After the fast retransmit algorithm sends what appears to be the
   missing segment, the "fast recovery" algorithm governs the
   transmission of new data until a non-duplicate ACK arrives. [RFC 5681, page 9](https://tools.ietf.org/html/rfc5681#page-9).

[...]

PR #225, opened 2018-06-02
Fixes bug regarding dup_acks overflow (TcpSocket)

barskern:
Removes the possibility of a panic caused by an 'attempt to add with
overflow' in TcpSocket.

This is a proposal is not based on an official RFC related to TCP, hence
it may not be the correct way of dealing with this problem. However I
thinks it should solve the problem.

Closes: #224

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -234,9 +234,9 @@ pub struct TcpSocket<'a> {
     remote_mss:      usize,
     /// The timestamp of the last packet received.
     remote_last_ts:  Option<Instant>,
-    /// The ACK number of the last packet recived. 
+    /// The ACK number of the last packet recived.
     local_rx_last_ack: Option<TcpSeqNumber>,
-    /// The number of packets recived directly after 
+    /// The number of packets recived directly after
     /// each other which have the same ACK number.
     local_rx_dup_acks: u8,
 }
@@ -900,29 +900,6 @@ impl<'a> TcpSocket<'a> {
                                ack_number, self.local_seq_no, self.local_seq_no + unacknowledged);
                     return Ok(Some(self.ack_reply(ip_repr, &repr)))
                 }
-
-                // 1. Check if duplicate ACK, and change self.local_rx_dup_acks accordingly
-                // 2. Update the last received ACK (self.local_rx_last_ack)
-                match self.local_rx_last_ack {
-                    // We have a duplicate ACK -> Increment dup ACKs count and 
-                    // set for retransmit if we just recived the third dup ACK
-                    Some(ref mut last_rx_ack) if *last_rx_ack == ack_number => {
-                        net_debug!("{}:{}:{}: duplicate ACK number {} for seq {}",
-                                 self.meta.handle, self.local_endpoint, self.remote_endpoint, 
-                                 self.local_rx_dup_acks, ack_number);
-                        self.local_rx_dup_acks += 1;
-                        if self.local_rx_dup_acks == 3 {
-                            self.timer.set_for_fast_retransmit();
-                            net_debug!("{}:{}:{}: set a fast retransmit for seq {}",
-                                     self.meta.handle, self.local_endpoint, self.remote_endpoint, ack_number);
-                        }
-                    },
-                    // We do not have a duplicate ACK -> Reset state
-                    _ => {
-                        self.local_rx_last_ack = Some(ack_number);
-                        self.local_rx_dup_acks = 1;
-                    }
-                };
             }
         }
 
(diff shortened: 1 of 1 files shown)


whitequark:
This seems OK but please add a test. You can set `local_rx_dup_acks` to 254 manually.

barskern:
@whitequark @pothos I have now implemented a suggestion which I think will resolve #224 and also improve the detection of duplicate ACKs. Now it should reset the counter when it encounters an ACK which contains data or that changes the send window. Further I implemented more tests to vertify this.

src/socket/tcp.rs
```diff
@@ -1170,6 +1147,51 @@ impl<'a> TcpSocket<'a> {
...
+                    // Increment duplicate ACK count
+                    match self.local_rx_dup_acks.checked_add(1) {
+                        Some(n) => {
+                            self.local_rx_dup_acks = n;
+                            net_debug!("{}:{}:{}: received duplicate ACK for seq {} (duplicate nr {})",
+                                    self.meta.handle, self.local_endpoint, self.remote_endpoint,
+                                    ack_number, self.local_rx_dup_acks);
+                        },
+                        None => {
+                            self.local_rx_dup_acks = u8::max_value();
```
whitequark: By definition it's already at `u8::max_value`.

src/socket/tcp.rs
```diff
@@ -1170,6 +1147,51 @@ impl<'a> TcpSocket<'a> {
...
+                            net_debug!("{}:{}:{}: received duplicate ACK for seq {} (duplicate nr overflowed)",
```
whitequark: I think it's more useful to say "255+" instead of just "overflowed".

src/socket/tcp.rs
```diff
@@ -1170,6 +1147,51 @@ impl<'a> TcpSocket<'a> {
...
+                                    self.meta.handle, self.local_endpoint, self.remote_endpoint, ack_number);
+                        }
+                    };
+
+                    if self.local_rx_dup_acks == 3 {
+                        self.timer.set_for_fast_retransmit();
+                        net_debug!("{}:{}:{}: set socket for fast retransmit",
```
whitequark: Nit: you're setting *timer* for fast retransmit; a better message is something like "starting fast retransmit".

src/socket/tcp.rs
```diff
@@ -3206,77 +3228,231 @@ mod test {
...
+        send!(s, time 0, TcpRepr {
+            seq_number: REMOTE_SEQ + 1,
+            ack_number: Some(LOCAL_SEQ + 1),
+            ..SEND_TEMPL
+        });
+        send!(s, time 0, TcpRepr {
+            seq_number: REMOTE_SEQ + 1,
+            ack_number: Some(LOCAL_SEQ + 1),
+            ..SEND_TEMPL
+        });
```
whitequark: An assertion that local_rx_dup_acks doesn't overflow to 0 would be good here.

src/socket/tcp.rs
barskern: I agree with this, however the reason for the awkward debug log is that this action doesn't technically send a retransmit, but it sets the socket up to send one. When the timeout expires for a "normal" retransmit it says something similar to this, hence I used the same phrase.

src/socket/tcp.rs
barskern: I added an assertion that checks this now, but I added it over lines you specified because if it would overflow to 0 it would happen after the `send` on line [3441](https://github.com/m-labs/smoltcp/pull/225/files#diff-d23c2633f4bdb8a9adc8ce0e490bcd12R3441).

src/socket/tcp.rs
```diff
@@ -1170,6 +1147,46 @@ impl<'a> TcpSocket<'a> {
...
+                // Increment duplicate ACK count and set for retransmit if we just recived
+                // the third duplicate ACK
+                Some(ref mut last_rx_ack) if
+                    repr.payload.len() == 0 &&
+                    *last_rx_ack == ack_number => {
+                    // Increment duplicate ACK count
+                    let overflowed_indication_symbol = match self.local_rx_dup_acks.checked_add(1) {
+                        Some(n) => { self.local_rx_dup_acks = n; "" },
+                        None => "+"
+                    };
```
whitequark: I hate to nitpick, but this match statement could be replaced with a `if self.local_rx_dup_acks == u8::max_value() { "+" } else { "" }` in the debug statement below, and then we wouldn't have logic and presentation of debug logging mixed.

barskern: Don't worry about nitpicking, I think that is a great way of maintaining a good code standard! And I like getting your feedback because it makes me think twice about how I would write something. Keep on nitpicking 😊

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 0dbfa65 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 0dbfa65ac460ed61a1a0eb4ad9d3b5f6e4a87474 with merge 216c393607927260ca2928648917ed27e75feeff...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/390192451?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 216c393607927260ca2928648917ed27e75feeff to master...

Issue #106, opened 2017-12-25
Implement TCP window scaling


whitequark:
Should be a matter of looking at the options and adding a few calls to functions that shift the window size back and forth.

whitequark:
Sure, go ahead.

pothos:
This can be closed, or? :)

jhwgh1968:
I think so, but since @whitequark opened it, I don't have the power to close it.

PR #232, opened 2018-06-09
TCP Window Scaling Phase 1

jhwgh1968:
Partially addresses #106.

As the title implies, this is the smallest change possible to _technically_ add support for TCP window scaling as defined in [section 2 of RFC 1323](https://tools.ietf.org/html/rfc1323#section-2). It avoids any changes to buffering behavior at all, and as a result, is not likely to provide any benefit to end users. (That is what Phase 2 will accomplish.)

The effects of this change are:
1. Window scaling is offered in the options of the initial SYN when initiating connections.
1. A scaling value of zero (i.e. multiply by 1) is always advertised for the local receiving window.
1. If window scaling is advertised by the remote side, that value is used for sending window calculations.
1. After window scaling is negotiated, several debug statements on the ``Repr`` of TCP are now misleading, because they print a "win" value without knowing the scaling factor in effect for the connection. 

Everything seems to work, except three unit tests are now failing because I somehow broke the MSS calculation.

Advice on how to fix that and comments about my approach are welcome.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -230,6 +230,8 @@ pub struct TcpSocket<'a> {
     /// The speculative remote window size.
     /// I.e. the actual remote window size minus the count of in-flight octets.
     remote_win_len:  usize,
+    /// The receive window scaling factor for remotes which support RFC 1323, None if unsupported.
+    remote_win_scale: Option<u8>,
     /// The maximum number of data octets that the remote side may receive.
     remote_mss:      usize,
     /// The timestamp of the last packet received.
@@ -272,6 +274,7 @@ impl<'a> TcpSocket<'a> {
             remote_last_ack: None,
             remote_last_win: 0,
             remote_win_len:  0,
+            remote_win_scale: None,
             remote_mss:      DEFAULT_MSS,
             remote_last_ts:  None,
             local_rx_last_ack: None,
@@ -398,6 +401,7 @@ impl<'a> TcpSocket<'a> {
         self.remote_last_ack = None;
         self.remote_last_win = 0;
         self.remote_win_len  = 0;
+        self.remote_win_scale = None;
         self.remote_mss      = DEFAULT_MSS;
         self.remote_last_ts  = None;
     }
@@ -769,6 +773,7 @@ impl<'a> TcpSocket<'a> {
             seq_number:   TcpSeqNumber(0),
             ack_number:   None,
             window_len:   0,
+            window_scale: None,
             max_seg_size: None,
             payload:      &[]
         };
@@ -1025,6 +1030,7 @@ impl<'a> TcpSocket<'a> {
                 if let Some(max_seg_size) = repr.max_seg_size {
                     self.remote_mss = max_seg_size as usize
                 }
+                self.remote_win_scale = repr.window_scale;
                 self.set_state(State::SynReceived);
                 self.timer.set_for_idle(timestamp, self.keep_alive);
             }
@@ -1135,7 +1141,10 @@ impl<'a> TcpSocket<'a> {
 
         // Update remote state.
         self.remote_last_ts = Some(timestamp);
-        self.remote_win_len = repr.window_len as usize;
+
+        // RFC 1323: The window field (SEG.WND) in the header of every incoming segment, with the
+        // exception of SYN segments, is left-shifted by Snd.Wind.Scale bits before updating SND.WND.
+        self.remote_win_len = (repr.window_len as usize) << (self.remote_win_scale.unwrap_or(0) as usize);
 
         if ack_len > 0 {
             // Dequeue acknowledged octets.
@@ -1376,6 +1385,7 @@ impl<'a> TcpSocket<'a> {
             seq_number:   self.remote_last_seq,
             ack_number:   Some(self.remote_seq_no + self.rx_buffer.len()),
             window_len:   self.rx_buffer.window() as u16,
+            window_scale: None,
             max_seg_size: None,
             payload:      &[]
         };
@@ -1396,6 +1406,9 @@ impl<'a> TcpSocket<'a> {
                 repr.control = TcpControl::Syn;
                 if self.state == State::SynSent {
                     repr.ack_number = None;
+                    repr.window_scale = Some(0);
+                } else {
+                    repr.window_scale = self.remote_win_scale.map(|_| 0);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/socket/tcp.rs
```diff
@@ -1396,6 +1406,9 @@ impl<'a> TcpSocket<'a> {
                 repr.control = TcpControl::Syn;
                 if self.state == State::SynSent {
                     repr.ack_number = None;
+                    repr.window_scale = Some(0);
+                } else {
+                    repr.window_scale = if self.remote_win_scale.is_some() { Some(0) } else { None }
```
ProgVal: nit: not sure about this repo's coding style, but I personally prefer this: `repr.window_scale = self.remote_win_scale.map(|_| 0)`

whitequark: `.unwrap_or(0)`

jhwgh1968: Fixed, @ProgVal.

Since this is an ``Option<u8>``, @whitequark, I'm not sure how ``unwrap_or()`` can create the same effect. Could you be more specific?

jhwgh1968:
I feel like I'm close. Now I just want to write one new regression unit test for non-windowed negotiation, and the test macros are not working the way I expect.

I have this code:

```rust
    #[test]
    fn test_connect_no_scaling() {
        let mut s = socket();
        s.local_seq_no = LOCAL_SEQ;
        s.connect(REMOTE_END, LOCAL_END.port).unwrap();
        assert_eq!(s.local_endpoint, IpEndpoint::new(MOCK_UNSPECIFIED, LOCAL_END.port));
        recv!(s, [TcpRepr {
            control:    TcpControl::Syn,
            seq_number: LOCAL_SEQ,
            ack_number: None,
            max_seg_size: Some(BASE_MSS),
            window_scale: None,
            ..RECV_TEMPL
        }]);
        send!(s, TcpRepr {
            control:    TcpControl::Syn,
            seq_number: REMOTE_SEQ,
            ack_number: Some(LOCAL_SEQ + 1),
            max_seg_size: Some(BASE_MSS - 80),
            window_scale: None,
            ..SEND_TEMPL
        });
    }
```
I expect it to mean: "pretend this packet was received on a new socket. Then this one should be sent". But the assertion is failing on the first packet, because ``window_scale`` is ``Some(0)`` instead of ``None``.

Any pointers?

jhwgh1968:
Nevermind. I was copying the wrong test as a basis.

I believe it's done now.

src/socket/tcp.rs
```diff
@@ -230,6 +230,8 @@ pub struct TcpSocket<'a> {
     /// The speculative remote window size.
     /// I.e. the actual remote window size minus the count of in-flight octets.
     remote_win_len:  usize,
+    /// The receive window scaling factor for remotes which support RFC-1323, None if unsupported.
+    remote_win_scale: Option<u8>,
```
whitequark: Do we really need an `Option` here? The `None` case seems identical to `Some(0)` in every respect.

src/wire/tcp.rs
```diff
@@ -690,7 +692,19 @@ impl<'a> Repr<'a> {
...
                 TcpOption::MaxSegmentSize(value) =>
                     max_seg_size = Some(value),
-                _ => ()
+                TcpOption::WindowScale(value) => {
+                    // RFC-1323: Thus, the shift count must be limited to 14 (which allows windows
+                    // of 2**30 = 1 Gbyte). If a Window Scale option is received with a shift.cnt
+                    // value exceeding 14, the TCP should log the error but use 14 instead of the
+                    // specified value.
+                    window_scale = if value > 14 {
+                        net_debug!("parsed window scaling factor >14, setting to 14");
```
whitequark: Take a look at how all other log statements are written--you're missing the socket endpoints in this one.

src/socket/tcp.rs
whitequark: Nit (here and elsewhere): we write RFC numbers as `RFC NNNN`, not `RFC-NNNN`.

src/wire/tcp.rs
```diff
@@ -702,20 +716,34 @@ impl<'a> Repr<'a> {
...
+            length += 3
+        }
         length
     }
 
+    /// Return the length of the header for the TCP protocol.
+    ///
+    /// Per RFC-6691, this should be used for MSS calculations. It may be smaller than the buffer
+    /// space required to accomodate this packet's data.
+    pub fn base_header_len(&self) -> usize {
```
whitequark: Maybe `mss_header_len`?

src/socket/tcp.rs
jhwgh1968: I will gladly fix the RFC numbers.

The reason for an option type is twofold. First, it is how phase 2 will handle the scaling window in my current prototype. Second, it is important in the case of listening sockets.

Our TCP should only respond with the window scaling option in the SYN-ACK if the client's SYN contained it. On line 1405 where that logic is applied, the only way to know what the original SYN contained is to check the field set on line 1032 when the packet was parsed. That value must represent the difference between "option not set" and "option set to multiplier of 1". Hence, values of ``None`` versus ``Some(0)``.

src/wire/tcp.rs
jhwgh1968: Fixed.

src/wire/tcp.rs
jhwgh1968: Since this is packet parsing code that does not contain an obvious reference to a socket, should I just print the source and destination addresses and ports? That means the fields will be reversed between packets we send and packets we receive, but it would be better than nothing.

jhwgh1968: Hmm, the obvious way looks a bit uglier than I thought:

```rust
net_debug!("{}->{}: parsed window scaling factor >14, setting to 14",
           IpEndpoint::new(src_addr.clone(), packet.src_port()), // borrowed src_addr, must be owned
           IpEndpoint::new(dst_addr.clone(), packet.dst_port()));  // borrowed dst_addr, must be owned
```

jhwgh1968: Nevermind, I was being too literal-minded. I think I've got it.

jhwgh1968:
I think I've addressed everything, @whitequark.

whitequark:
Thanks! Please squash the commits. r=me after that.

jhwgh1968:
@m-labs-homu r=whitequark

m-labs-homu:
@jhwgh1968: :key: Insufficient privileges: Not in reviewers

dlrobertson:
@m-labs-homu r=whitequark

m-labs-homu:
:pushpin: Commit d08173d has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit d08173d568fd3e5f1498fbe7f2cc32dba455701c with merge 42d25bfe2645fe442ad6e642582af83f79b7bb6f...

Issue #83, opened 2017-11-21
Eviction of entries from owned neighbor cache


whitequark:
Right now the neighbor cache will simply grow for as long as it can when unseen before IPs are added. This can potentially take a lot of memory if we're talking about a /8 network that someone runs an ARP scan against. We should probably do something about it.

squidarth:
hi @whitequark! I'm new here and v. interested in contributing to this library! I want to confirm what needs to happen here before diving in:

It looks like in the case where the neighbor cache is _not_ using heap-allocated storage that there is already logic in place to evict old entries:

```rust
     // src/iface/neighbor.rs:94

    Err((protocol_addr, neighbor)) => {
     // If we're going down this branch, it means that a fixed-size cache storage
     // is full, and we need to evict an entry.
     let old_protocol_addr = match self.storage {
         ManagedMap::Borrowed(ref mut pairs) => {
             pairs
                 .iter()
                 .min_by_key(|pair_opt| {
                     let (_protocol_addr, neighbor) = pair_opt.unwrap();
                     neighbor.expires_at
                 })
                 .expect("empty neighbor cache storage") // unwraps min_by_key
                 .unwrap() // unwraps pair
                 .0
         }
         // Owned maps can extend themselves.
         #[cfg(any(feature = "std", feature = "alloc"))]
         ManagedMap::Owned(_) => unreachable!()
     };

     let _old_neighbor =
         self.storage.remove(&old_protocol_addr).unwrap();
     match self.storage.insert(protocol_addr, neighbor) {
         Ok(None) => {
             net_trace!("filled {} => {} (evicted {} => {})",
                        protocol_addr, hardware_addr,
                        old_protocol_addr, _old_neighbor.hardware_addr);
         }
         // We've covered everything else above.
         _ => unreachable!()
     }
 }
```

is the task here to repeat this logic for the heap-allocated case? 

2. Do we want to make the the maximum size of the cache configurable, or do we want to be opinionated and set it to some reasonable number? Seems like the type of thing that ought to be configurable.

Thanks a lot for any guidance & for building this awesome library--I'm excited to dive in!

whitequark:
Hi @squidarth! The challenge here, first and foremost, in figuring out what logic exactly should be followed here. I didn't really have much time to put into this issue myself. Do you think you could look at other TCP/IP stacks and see what sort of logic they use for the neighbor table?

squidarth:
Yep--sounds good

[...]

PR #234, opened 2018-06-12
Add GC & threshold to the ARP cache.

squidarth:
Implementation of solution discussed in: https://github.com/m-labs/smoltcp/issues/83

Main things to look at:

1. I figured this only really makes sense for the `ManagedMap::Owned` case, so I conditionally compile for the `alloc` case. Let me know if this is wrong

2. There weren't any tests in the `neighbor.rs` file that use the `Owned` case, so I had to start using `BTreeMap`. I'm not totally sure what the downsides of this are, so let me know if there's anything I should be aware of.

3. The reason I needed to upgrade the version of `managed` is so that we could have `len` operator on ManagedMap.

### Manual Testing

I did manually test this with the `httpclient` example (i changed around where `run_gc` gets run in order to make it do something) and checked that the sizes were correct.

thanks for taking a look at this!

### DIFF
--- a/src/iface/neighbor.rs
+++ b/src/iface/neighbor.rs
@@ -6,6 +6,9 @@ use managed::ManagedMap;
 use wire::{EthernetAddress, IpAddress};
 use time::{Duration, Instant};
 
+#[cfg(any(feature = "std", feature = "alloc"))]
+use core::mem;
+
 /// A cached neighbor.
 ///
 /// A neighbor mapping translates from a protocol address to a hardware address,
@@ -51,6 +54,8 @@ pub(crate) enum Answer {
 pub struct Cache<'a> {
     storage:      ManagedMap<'a, IpAddress, Neighbor>,
     silent_until: Instant,
+    gc_threshold: usize
+
 }
 
 impl<'a> Cache<'a> {
@@ -60,16 +65,25 @@ impl<'a> Cache<'a> {
     /// Neighbor entry lifetime, in milliseconds.
     pub(crate) const ENTRY_LIFETIME: Duration = Duration { millis: 60_000 };
 
+    /// Default number of entries in the cache before GC kicks in
+    pub(crate) const GC_THRESHOLD: usize = 1024;
+
     /// Create a cache. The backing storage is cleared upon creation.
     ///
     /// # Panics
     /// This function panics if `storage.len() == 0`.
     pub fn new<T>(storage: T) -> Cache<'a>
             where T: Into<ManagedMap<'a, IpAddress, Neighbor>> {
+
+        Cache::new_with_limit(storage, Cache::GC_THRESHOLD)
+    }
+
+    pub fn new_with_limit<T>(storage: T, gc_threshold: usize) -> Cache<'a>
+            where T: Into<ManagedMap<'a, IpAddress, Neighbor>> {
         let mut storage = storage.into();
         storage.clear();
 
-        Cache { storage, silent_until: Instant::from_millis(0) }
+        Cache { storage, gc_threshold, silent_until: Instant::from_millis(0) }
     }
 
     pub(crate) fn fill(&mut self, protocol_addr: IpAddress, hardware_addr: EthernetAddress,
@@ -77,6 +91,23 @@ impl<'a> Cache<'a> {
         debug_assert!(protocol_addr.is_unicast());
         debug_assert!(hardware_addr.is_unicast());
 
+        #[cfg(any(feature = "std", feature = "alloc"))]
+        let current_storage_size = self.storage.len();
+
+        match self.storage {
+            ManagedMap::Borrowed(_) =>  (),
+            #[cfg(any(feature = "std", feature = "alloc"))]
+            ManagedMap::Owned(ref mut map) => {
+                if current_storage_size >= self.gc_threshold {
+                    let new_btree_map = map.into_iter()
+                        .map(|(key, value)| (*key, *value))
+                        .filter(|(_, v)| timestamp < v.expires_at)
+                        .collect();
+
+                    mem::replace(map, new_btree_map);
+                }
+            }
+        };
         let neighbor = Neighbor {
             expires_at: timestamp + Self::ENTRY_LIFETIME, hardware_addr
         };
@@ -161,7 +192,9 @@ impl<'a> Cache<'a> {
 #[cfg(test)]
 mod test {
     use super::*;
+    use std::collections::BTreeMap;
     use wire::ip::test::{MOCK_IP_ADDR_1, MOCK_IP_ADDR_2, MOCK_IP_ADDR_3, MOCK_IP_ADDR_4};
+
 
     const HADDR_A: EthernetAddress = EthernetAddress([0, 0, 0, 0, 0, 1]);
     const HADDR_B: EthernetAddress = EthernetAddress([0, 0, 0, 0, 0, 2]);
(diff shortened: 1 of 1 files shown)


squidarth:
@dlrobertson, awesome, thanks for the review. updated!

squidarth:
cool! @dlrobertson do you have any ides for other issues that might be good to get started on next? thanks

src/iface/neighbor.rs
```diff
@@ -51,6 +51,7 @@ pub(crate) enum Answer {
 pub struct Cache<'a> {
     storage:      ManagedMap<'a, IpAddress, Neighbor>,
     silent_until: Instant,
+    gc_thresh: usize
```
whitequark: `gc_threshold`, please

src/iface/neighbor.rs
```diff
@@ -60,23 +61,59 @@ impl<'a> Cache<'a> {
     /// Neighbor entry lifetime, in milliseconds.
     pub(crate) const ENTRY_LIFETIME: Duration = Duration { millis: 60_000 };
 
+    /// Default number of entries in the cache before GC kicks in
+    pub(crate) const GC_THRESH: usize = 1024;
```
whitequark: also here

src/iface/neighbor.rs
```diff
@@ -60,23 +61,59 @@ impl<'a> Cache<'a> {
...
+        Cache::new_with_limit(storage, Cache::GC_THRESH)
+    }
+
+    pub fn new_with_limit<T>(storage: T, gc_thresh: usize) -> Cache<'a>
+            where T: Into<ManagedMap<'a, IpAddress, Neighbor>> {
         let mut storage = storage.into();
         storage.clear();
 
-        Cache { storage, silent_until: Instant::from_millis(0) }
+        Cache { storage, gc_thresh: gc_thresh, silent_until: Instant::from_millis(0) }
```
whitequark: You can use `gc_thresh` instead of `gc_thresh: gc_thresh`.

src/iface/neighbor.rs
```diff
@@ -60,23 +61,59 @@ impl<'a> Cache<'a> {
...
+    }
+
+    #[cfg(any(feature = "std", feature = "alloc"))]
+    pub fn run_owned_gc(&mut self, timestamp: Instant) {
```
whitequark: Should this really be `pub`?

src/iface/neighbor.rs
```diff
@@ -60,23 +61,59 @@ impl<'a> Cache<'a> {
...
+        self.storage = ManagedMap::Owned(new_map);
     }
 
     pub(crate) fn fill(&mut self, protocol_addr: IpAddress, hardware_addr: EthernetAddress,
                        timestamp: Instant) {
         debug_assert!(protocol_addr.is_unicast());
         debug_assert!(hardware_addr.is_unicast());
 
+        match self.storage {
+            ManagedMap::Borrowed(_) =>  {
```
dlrobertson: nit

```rust
ManagedMap::Borrowed(_) => (),
```

src/iface/neighbor.rs
```diff
@@ -60,23 +61,59 @@ impl<'a> Cache<'a> {
...
+    }
+
+    #[cfg(any(feature = "std", feature = "alloc"))]
+    pub fn run_owned_gc(&mut self, timestamp: Instant) {
+        let new_map = match self.storage {
+            ManagedMap::Borrowed(_) => {
+                unreachable!()
+            }
+            ManagedMap::Owned(ref mut map) => {
+                map.iter_mut()
```
whitequark: Maybe `.into_iter()` so that we don't need to clone values? If the naive form doesn't pass borrow checker, use `swap`.

src/iface/neighbor.rs
```diff
@@ -60,23 +61,59 @@ impl<'a> Cache<'a> {
...
         debug_assert!(hardware_addr.is_unicast());
 
+        match self.storage {
+            ManagedMap::Borrowed(_) =>  {
+                ();
+            }
+            #[cfg(any(feature = "std", feature = "alloc"))]
+            ManagedMap::Owned(_) => {
+                if self.storage.len() >= self.gc_thresh {
+                    self.run_owned_gc(timestamp);
```
dlrobertson: Would it be more readable to have a static method that runs GC on a `BTreeMap` instead of `self`? That way we don't have the `match` with the `unreachable` statement. Just a thought.

whitequark: :+1:

squidarth:
@whitequark @dlrobertson Updated. The problem with the static btree approach is that you can't assign `self.storage` inside the `match` block--I addressed this by using `swap`, but am not totally sure if this is more readable. Let me know what you think, and thanks again for the review

dlrobertson:
> am not totally sure if this is more readable.

I think a doc comment could solve this.

squidarth:
Ah, realized something with the test failure. This isn't going to work with the case where we are using `alloc` without `std`, since the static gc method depends on `std` (for importing the btreemap).

I also don't like that i have to add  ` let _current_storage_size = self.storage.len();` to the top of the match statement.

Thinking it might make sense to just switch back to the non-static `run_owned_gc`method, what do you think?

squidarth:
Here's what I had originally for reference (with updates for the other PR comments here): https://github.com/m-labs/smoltcp/compare/master...squidarth:ss-83-arp-cache-eviction-2?expand=1

dlrobertson:
> I also don't like that i have to add let _current_storage_size = self.storage.len(); to the top of the match statement.

IMO adding a `cfg` to the variable would be better.

> Ah, realized something with the test failure. This isn't going to work with the case where we are using alloc without std, since the static gc method depends on std (for importing the btreemap).

I think you have two other alternatives here that would be a bit cleaner.

1) Now that `run_gc` is static, I can more readily see you could inline the body of `run_gc into the match statement without much of a readability penalty.
2) `BTreeMap` and `mem::replace` are not defined strictly in `std`. You could use `core::mem::replace` instead of `std::mem::replace` and something like the following for `BTreeMap`

```rust
#[cfg(feature = "std")]
use std::collections::BTreeMap;
#[cfg(all(not(feature = "std"), feature = "alloc"))]
use alloc::btree_map::BTreeMap;
```

src/iface/neighbor.rs
```diff
@@ -60,23 +65,49 @@ impl<'a> Cache<'a> {
...
+            ManagedMap::Borrowed(_) =>  (),
+            #[cfg(any(feature = "std", feature = "alloc"))]
+            ManagedMap::Owned(ref mut map) => {
+                if current_storage_size >= self.gc_threshold {
+                    let new_btree_map = map.into_iter()
+                        .map(|(key, value)| (*key, *value))
+                        .filter(|(_, v)| timestamp < v.expires_at)
+                        .collect();
+
+                    replace(map, new_btree_map);
```
dlrobertson: 1 last nit: `use core::mem` so that this becomes `mem::replace`, then it becomes more clear that `core::mem::replace` is being used here.

squidarth:
👍 sounds good, updated

dlrobertson:
@whitequark any other feedback for this patch?

whitequark:
Sorry for the delay on this--looks good! Thanks @squidarth!

@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 4f783ce has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 4f783ceaebe3bd4d46db8e5bf4b401c7f0600cdd with merge 80c8be3cc6f7ce1208e7cfeba59d7ee30e4a2c83...

m-labs-homu:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/396001312?utm_source=github_status&utm_medium=notification)

whitequark:
@m-labs-homu retry

m-labs-homu:
:hourglass: Testing commit 4f783ceaebe3bd4d46db8e5bf4b401c7f0600cdd with merge d7eb96963790751dc390c91b8ca3da568373c4f6...

whitequark:
Oh--can you please squash the commits?

dlrobertson:
@m-labs-homu r=whitequark

m-labs-homu:
:pushpin: Commit fba1afd has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit fba1afdde9abaf22c293048c569f5bef826c8ac6 with merge d4b3318dcdfd30b92e25ee5f6bb853e9ccd7671d...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/396138586?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing d4b3318dcdfd30b92e25ee5f6bb853e9ccd7671d to master...

dlrobertson:
@squidarth Thanks for working on this!

PR #236, opened 2018-06-13
Ipv4 fragmentation refactored

podhrmic:
Based on #190 except rebased on the recent master.

Notes based on the previous reviews:
1) the reassembly algorithm was tested and works
2) to actually reassemble packets, it requires adding ethernet header lenght to MTU. This will be handled in a separate PR
3) FragmentSet being backed by ManagedSlice is tested and working. ManagedMap will not work in "no-std" environments


podhrmic:
Added tests and the Ethernet header size to MTU, instead of #237

podhrmic:
@pothos @whitequark any feedback here?

src/iface/ethernet.rs
```diff
@@ -675,14 +695,97 @@ impl<'b, 'c, 'e> InterfaceInner<'b, 'c, 'e> {
         }
     }
 
+/// Process an IPv4 fragment
```
dlrobertson: whitespace

Cargo.toml
```diff
@@ -34,6 +34,7 @@ verbose = []
 "phy-raw_socket" = ["std", "libc"]
 "phy-tap_interface" = ["std", "libc"]
 "proto-ipv4" = []
+"fragmentation-ipv4" = ["proto-ipv4"]
```
dlrobertson: :bike: :house: : `ipv4-fragmentation`? Will it ever make sense to have a single generic `ip-fragmentation` feature? Is there a reason to keep two separate features for ipv4 and ipv6?

src/iface/fragments.rs
```diff
@@ -0,0 +1,291 @@
...
+                           return true;
+                       }
+                },
+            }
+        }
+        false
+    }
+
+    /// Get a packet with given key
+    pub fn get_packet(&mut self, id: u16, src_addr: Ipv4Address, dst_addr: Ipv4Address, time: Instant) -> Option<&mut Packet<'a>> {
```
dlrobertson: The is `O(3n)` right? Is there a way we could make this faster? What if we used a `ManagedMap` instead of a `ManagedSlice`?

dlrobertson (commented):
Thanks for working on this! I'm pumped about getting IP fragmentation support into `smoltcp`.

Just an initial read through. I'll need to read through a few more times.

m-labs-homu:
:umbrella: The latest upstream changes (presumably 94d7338aed6d4b5157d328915b5ac94236fc6427) made this pull request unmergeable. Please resolve the merge conflicts.

Cargo.toml
podhrmic: I am not familiar with Ipv6 so I cannot comment on how similar to IPv4 it is. But it can be refactored in the future, if someone with more knowledge about Ipv6 can pitch in. I would keep it separate for now, better some fragmentation support than none:-)

src/iface/fragments.rs
podhrmic: There is a thread in the old PR where I tried to explain why `ManagedMap` doesn't work. In short, to insert a new packet to `ManagedMap` you need to allocate memory for it somewhere. And either you create a new `Vec<u8>` which is easy but requires `alloc`. Or you have a static buffers somewhere, and that effectively makes it a `ManagedSlice`.

I am not sure how much fragmented packets you normally encounter (TCP doesn't use fragmented packets typically), in our tests the performance penalty was negligible (with large UDP packets).

But - to use fragmentation in `no-std` builds, you still need to solve problem with `From` conversion for `ManagedSlice` here: https://github.com/m-labs/rust-managed/blob/master/src/slice.rs#L74

If you don't have alloc/std then rust doesn't know how to create a `FragmentSet` from static buffers. It would be cool to be able to do fragmentation without `alloc`, but I am not sure what is the best way around this.

src/iface/ethernet.rs
podhrmic: Fixed

src/iface/fragments.rs
podhrmic: Edit: I realized even the `loopback.rs` example you have to run with at least `alloc` because of the `From` trait implementation. This definitely limits smoltcp's usage on bare metal systems:-(

dlrobertson: > And either you create a new Vec<u8> which is easy but requires alloc. Or you have a static buffers somewhere, and that effectively makes it a ManagedSlice.

Yeah it is basically the same, but you get massivly improved lookup times due to the use of a btree. Couldn't you just use something like the following.

```rust
const FRAGMENT_BUFFER_SIZE: usize = // ...
static FRAGMENT_BUFFER: [Option<Ident, FragmentPacket> ; FRAGMENT_BUFFER_SIZE] = [None; FRAGMENT_BUFFER_SIZE];
let fragment_set = FragmentSet::new(FRAGMENT_BUFFER);
```
> I am not sure how much fragmented packets you normally encounter (TCP doesn't use fragmented packets typically), in our tests the performance penalty was negligible (with large UDP packets).

Lots of embedded applicaitons make heavy use of fragmented packets. See 6LoWPAN ([RFC 4944] and [RFC 6282]). As `smoltcp` matures this is going to become more and more important. IMO the next two big hurdles are this and breaking up the `EthernetInterface`.

[RFC 4944]: https://tools.ietf.org/html/rfc4944
[RFC 6282]: https://tools.ietf.org/html/rfc6282

whitequark: Thanks @dlrobertson, these are my thoughts exactly.

podhrmic:
@whitequark @dlrobertson I switched FragmentedSet to ManagedMap.

The unresolved things are:
1) when and how to purge stale fragments? Should I iterate over all fragments every time I call `get_packet`? Or should I somehow check the size of the map, and remove old fragments only when the size is too large? Stale means the fragmented packet is not completed, and it wasn't modified for more than a timeout (e.g. 500ms)?
2) how to remove (K,V) pairs from the map? I can do a copy of the buffer and create a new packet here: https://github.com/podhrmic/smoltcp/blob/867edfefeaa7f8e5dc1f75676a7a27e35f52a3d9/src/iface/ethernet.rs#L760 but fragments are still mutually borrowed at that point so I cannot borrow it again.
3) `ManagedMap` means you need `alloc` at least, so it won't work anymore on bare-metal systems. 

what do you think is the best way to tackle these problems?

dlrobertson:
> ManagedMap means you need alloc at least, so it won't work anymore on bare-metal systems.

I don't think this is true. See [the implementation in rust-managed](https://github.com/m-labs/rust-managed/blob/master/src/map.rs#L39-L45).

> when and how to purge stale fragments? Should I iterate over all fragments every time I call get_packet? Or should I somehow check the size of the map, and remove old fragments only when the size is too large? Stale means the fragmented packet is not completed, and it wasn't modified for more than a timeout (e.g. 500ms)?

I think the cleanup implementation for the neighbor cache turned out to be a clean implementation. Maybe something like that would work here? I think some sort of count threshold makes more sense. Especially since IIRC the `Borrowed` variant of the `ManagedMap` does this.

podhrmic:
I am having a problem using `ManagedMap` in `no-std` mode. If I do something like this in `loopback.rs`:

```Rust
const FRAGMENT_BUFFER_SIZE: usize = 6;
    static FRAGMENT_BUFFER: [Option<(FragmentedKey, FragmentedPacket)> ; FRAGMENT_BUFFER_SIZE] = [None; FRAGMENT_BUFFER_SIZE];
    let fragment_set = FragmentSet::new(FRAGMENT_BUFFER);
```

I get this error:
```
error[E0277]: the trait bound `smoltcp::iface::Packet<'static>: core::marker::Copy` is not satisfied in `(smoltcp::iface::Key, smoltcp::iface::Packet<'static>)`
   --> examples/loopback.rs:123:96
    |
123 |     static FRAGMENT_BUFFER: [Option<(FragmentKey, FragmentedPacket)> ; FRAGMENT_BUFFER_SIZE] = [None; FRAGMENT_BUFFER_SIZE];
    |                                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ within `(smoltcp::iface::Key, smoltcp::iface::Packet<'static>)`, the trait `core::marker::Copy` is not implemented for `smoltcp::iface::Packet<'static>`
    |
    = note: required because it appears within the type `(smoltcp::iface::Key, smoltcp::iface::Packet<'static>)`
    = note: required because of the requirements on the impl of `core::marker::Copy` for `core::option::Option<(smoltcp::iface::Key, smoltcp::iface::Packet<'static>)>`
    = note: the `Copy` trait is required because the repeated element will be copied


```

I am compiling without `std` or `alloc`. Packet indeed doesn't implement `Copy`, but how to go around this limit?

src/phy/mod.rs
```diff
@@ -265,3 +265,8 @@ pub trait TxToken {
     fn consume<R, F>(self, timestamp: Instant, len: usize, f: F) -> Result<R>
         where F: FnOnce(&mut [u8]) -> Result<R>;
 }
+
+/// Ethernet header length to be added to the MTU, since the header is not counted into MTU
+/// 6 bytes src addr, 6 bytes dst addr, 2 bytes EthType,
+/// the optionbal 4 bytes of IEEE 802.1Q Header are not accounted for
+pub const ETHERNET_HAEDER_MAX_LEN: usize = 14;
```
pothos: I think it would be better to use 18 here as correct maximal length because otherwise you could just use `EthernetFrame::<&[u8]>::header_len()` which is 14.

When using 18, this has to be changed as well:
```
diff --git a/src/iface/ethernet.rs b/src/iface/ethernet.rs
index e44abb4..e3fd029 100644
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -388,7 +388,7 @@ impl<'b, 'c, 'e, DeviceT> Interface<'b, 'c, 'e, DeviceT>
 
     fn socket_egress(&mut self, sockets: &mut SocketSet, timestamp: Instant) -> Result<bool> {
         let mut caps = self.device.capabilities();
-        caps.max_transmission_unit -= EthernetFrame::<&[u8]>::header_len();
+        caps.max_transmission_unit -= ETHERNET_HAEDER_MAX_LEN;
 
         let mut emitted_any = false;
         for mut socket in sockets.iter_mut() {
@@ -1122,7 +1122,7 @@ impl<'b, 'c, 'e> InterfaceInner<'b, 'c, 'e> {
                     // I'm really not happy about this "solution" but I don't know what else to do.
                     if let Some(max_burst_size) = caps.max_burst_size {
                         let mut max_segment_size = caps.max_transmission_unit;
-                        max_segment_size -= EthernetFrame::<&[u8]>::header_len();
+                        max_segment_size -= ETHERNET_HAEDER_MAX_LEN;
                         max_segment_size -= ip_repr.buffer_len();
                         max_segment_size -= tcp_repr.header_len();
```

m-labs-homu:
:umbrella: The latest upstream changes (presumably 58a54730dba398aaa5ea88e6c6f89e0234f89f42) made this pull request unmergeable. Please resolve the merge conflicts.

jean-airoldie:
@podhrmic This issue is caused because rust requires elements to implement copy when using the `[type; size]` syntax for some reason. Its kind of a pain since you usually have to manually 
initialize the array ([this example](https://github.com/tokio-rs/tokio/blob/74168ae82f220fd5e7644b72426e1d851d2f622e/tokio-timer/src/wheel/level.rs#L53) comes to mind).

However in this case, since `Option<T>` implements `Default` as `None` you should be able to write:
```rust
    static FRAGMENT_BUFFER: [Option<(FragmentedKey, FragmentedPacket)> FRAGMENT_BUFFER_SIZE] = Default::default();
```

MabezDev:
Is there any movement on this? It looks so close to complete it would be a shame to not see it merged.

podhrmic:
Unfortunately I don't have time to finish this, so somebody else would have to step in. I think a little bit more work is needed, but I had the algorithm working in a fork of smoltcp last year.

Dirbaio:
Closing in favor of #624

PR #237, opened 2018-06-13
Fix MTU settings so fragmented packets can be received

podhrmic:
This is needed to get the fragmentation to work. In Linux MTU refers to the max length of the ethernet payload. But smoltcp counts MTU as max length of the ethernet frame. Hence without this added ethernet length, all fragmented packets are truncated. I am happy to fix is more meaningfully later, but without this fix the fragmentation reassembly doesn't work.

### DIFF
--- a/src/phy/mod.rs
+++ b/src/phy/mod.rs
@@ -113,6 +113,10 @@ pub use self::raw_socket::RawSocket;
 #[cfg(all(feature = "phy-tap_interface", target_os = "linux"))]
 pub use self::tap_interface::TapInterface;
 
+/// Ethernet header length to be added to the MTU, since the header is not counted into MTU
+/// 6 bytes src addr, 6 bytes dst addr, 2 bytes EthType, (optional) 4 bytes IEEE 802.1Q Header
+pub const ETHERNET_HAEDER_MAX_LEN: usize = 18;
+
 /// A tracer device for Ethernet frames.
 pub type EthernetTracer<T> = Tracer<T, super::wire::EthernetFrame<&'static [u8]>>;
 
--- a/src/phy/raw_socket.rs
+++ b/src/phy/raw_socket.rs
@@ -5,7 +5,7 @@ use std::io;
 use std::os::unix::io::{RawFd, AsRawFd};
 
 use Result;
-use phy::{self, sys, DeviceCapabilities, Device};
+use phy::{self, sys, DeviceCapabilities, Device, ETHERNET_HAEDER_MAX_LEN};
 use time::Instant;
 
 /// A socket that captures or transmits the complete frame.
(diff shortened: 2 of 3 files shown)


pothos:
Hi, I tried this but in the end the packets were 4 bytes too big. Maybe its the calculation in the TCP code or the ethernet frame code has a different length? Or Linux reports something different…
Did sending TCP with MSS work for you?

`tx buffer: sending 1464 octets af offset […]` → this works with 1460, i.e. changing your code to `+ 14`

podhrmic:
For Ethernet type II 14 bytes would work. In Linux, the default MTU is 1500, so a fragmented Ethernet frame with an IPV4 packet is 1514 bytes long. Maybe we can settle this as a temporary measure and properly detect the ethernet type in a different PR once the fragmentation code is merged?

pothos:
Or maybe just fix this line here to subtract the MAX length you defined instead of `EthernetFrame::<&[u8]>::header_len();` (didn't try it myself yet…) https://github.com/m-labs/smoltcp/blob/06d18130cd8b25edbcad207bbdbb157f1c1203df/src/iface/ethernet.rs#L391

podhrmic:
Nope, that doesn't work. Mostly because it is on the egress part, while I am parsing incoming (ingress) frames.

pothos:
I'm talking about this which solves the problem of too big TCP MSS:
```
index e44abb4..2be3991 100644
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -388,7 +388,7 @@ impl<'b, 'c, 'e, DeviceT> Interface<'b, 'c, 'e, DeviceT>
 
     fn socket_egress(&mut self, sockets: &mut SocketSet, timestamp: Instant) -> Result<bool> {
         let mut caps = self.device.capabilities();
-        caps.max_transmission_unit -= EthernetFrame::<&[u8]>::header_len();
+        caps.max_transmission_unit -= ETHERNET_HAEDER_MAX_LEN;
 
         let mut emitted_any = false;
         for mut socket in sockets.iter_mut() {
diff --git a/src/phy/raw_socket.rs b/src/phy/raw_socket.rs
index 5303d7d..95a1aaa 100644
--- a/src/phy/raw_socket.rs
+++ b/src/phy/raw_socket.rs
@@ -32,7 +32,7 @@ impl RawSocket {
         let mtu = lower.interface_mtu()?;
         Ok(RawSocket {
             lower: Rc::new(RefCell::new(lower)),
-            mtu:   mtu
+            mtu:   mtu + ETHERNET_HAEDER_MAX_LEN
         })
     }
 }
diff --git a/src/phy/tap_interface.rs b/src/phy/tap_interface.rs
index c34790d..53010ea 100644
--- a/src/phy/tap_interface.rs
+++ b/src/phy/tap_interface.rs
@@ -52,7 +52,7 @@ impl TapInterface {
         let mtu = lower.interface_mtu()?;
         Ok(TapInterface {
             lower: Rc::new(RefCell::new(lower)),
-            mtu:   mtu
+            mtu:   mtu + ETHERNET_HAEDER_MAX_LEN
         })
     }
 }
```

podhrmic:
Please see #236 I added the header size there. I am closing this and moving the discussing there.

PR #244, opened 2018-06-19
Clean up routes module slightly

jD91mZM2:
- Replace update with get_ref/get_mut
- Rename ipv4 add_default_ipv4_route to set since it can only have one gateway
- Add get_default_ipv4_route and get_default_ipv6_routes
- Add remove_default_ipv4_route and clear_default_ipv6_routes

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -267,16 +267,13 @@ impl<'b, 'c, 'e, DeviceT> Interface<'b, 'c, 'e, DeviceT>
         self.inner.ip_addrs.as_ref()
     }
 
-    /// Replace the set of IP addresses.
+    /// Update the IP addresses of the interface.
     ///
     /// # Panics
     /// This function panics if any of the addresses are not unicast.
-    pub fn set_ip_addrs<S>(&mut self, ips: S)
-        where S: Into<ManagedSlice<'c, IpCidr>>
-    {
-        let ips = ips.into();
-        InterfaceInner::check_ip_addrs(&ips);
-        self.inner.ip_addrs = ips;
+    pub fn update_ip_addrs<F: FnOnce(&mut ManagedSlice<'c, IpCidr>)>(&mut self, f: F) {
+        f(&mut self.inner.ip_addrs);
+        InterfaceInner::check_ip_addrs(&self.inner.ip_addrs)
     }
 
     /// Check whether the interface has the given IP address assigned.
@@ -1901,9 +1898,10 @@ mod test {
         let (mut iface, _) = create_loopback();
         let mut new_addrs = vec![IpCidr::new(IpAddress::v6(0xfe80, 0, 0, 0, 1, 2, 0, 2), 64),
                                  IpCidr::new(IpAddress::v6(0xfe80, 0, 0, 0, 3, 4, 0, 0xffff), 64)];
-        new_addrs.extend(iface.ip_addrs());
-        iface.set_ip_addrs(new_addrs);
-
+        iface.update_ip_addrs(|addrs| {
+            new_addrs.extend(addrs.to_vec());
+            *addrs = From::from(new_addrs);
+        });
         assert!(iface.inner.has_solicited_node(Ipv6Address::new(0xff02, 0, 0, 0, 0, 1, 0xff00, 0x0002)));
         assert!(iface.inner.has_solicited_node(Ipv6Address::new(0xff02, 0, 0, 0, 0, 1, 0xff00, 0xffff)));
         assert!(!iface.inner.has_solicited_node(Ipv6Address::new(0xff02, 0, 0, 0, 0, 1, 0xff00, 0x0003)));


jD91mZM2:
I think `update` is fine, but there's no reason to have it take a closure :)

whitequark:
There is--if you just give away a `&mut` there is no way to maintain any internal invariants. There aren't any right now but that is likely wrong.

jD91mZM2:
What exactly do you mean by internal invariants?

whitequark:
Not every route is legal. For example, no multicast addresses should be added to the route table. This isn't checked currently, but should be, and it can't be checked with `get_mut`.

jD91mZM2:
Aaaaaah okay, that explains it. I'll implement your add_route idea

ProgVal:
The `update` method is inspired from the one in `ethernet.rs`, so you might want to change it as well.

jD91mZM2:
There wasn't much to do in `ethernet.rs` unfortunately since a ManagedSlice cannot be pushed/removed to.

src/iface/ethernet.rs
```diff
@@ -267,13 +267,16 @@ impl<'b, 'c, 'e, DeviceT> Interface<'b, 'c, 'e, DeviceT>
...
-    /// Update the IP addresses of the interface.
+    /// Replace the set of IP addresses.
     ///
     /// # Panics
     /// This function panics if any of the addresses are not unicast.
-    pub fn update_ip_addrs<F: FnOnce(&mut ManagedSlice<'c, IpCidr>)>(&mut self, f: F) {
-        f(&mut self.inner.ip_addrs);
-        InterfaceInner::check_ip_addrs(&self.inner.ip_addrs)
+    pub fn set_ip_addrs<S>(&mut self, ips: S)
+        where S: Into<ManagedSlice<'c, IpCidr>>
```
whitequark: The purpose of this method is to be able to update IP addresses without assigning a different ManagedSlice. It's for memory-constrained devices without an allocator.

jD91mZM2: Well you still can't append to it. Should I just make a way to set an ip based on index?

src/iface/route.rs
```diff
@@ -195,9 +248,7 @@ mod test {
             via_router: ADDR_1A.into(),
             preferred_until: None, expires_at: None,
         };
-        routes.update(|storage| {
-            storage.insert(cidr_1().into(), route).unwrap();
```
whitequark: So what does this achieve? One line less in route setup code? I'm not convinced it's worth adding so many functions in the public API (and you don't even validate input in some of them).

jD91mZM2: Well, imagine you want to fetch the default gateway with the current code:
```Rust
let mut gateway = None;
routes.update(|storage| {
     let cidr = IpCidr::new(IpAddress::v4(0, 0, 0, 0), 0);
    gateway = Some(storage.get(&cidr).clone());
});
let gateway = gateway.unwrap();
```

jD91mZM2: See the [redox netstack code after this PR](https://gitlab.redox-os.org/jD91mZM2/netstack/commit/f9d6f5d87f1f5903d301c6205b9b5637c8ad1512).

(Your existing code doesn't validate input either btw)

src/iface/route.rs
```diff
@@ -74,16 +74,36 @@ impl<'a> Routes<'a> {
...
+    ///
+    /// On success, return the previous route, if any.
+    pub fn add_route(&mut self, cidr: IpCidr, route: Route) -> Result<Option<Route>> {
+        // TODO: Validate
+        match self.storage.insert(cidr, route) {
+            Ok(route) => Ok(route),
+            Err((_cidr, _route)) => Err(Error::Exhausted)
+        }
+    }
+    /// Returns a route registered on cidr
```
whitequark: This docstring isn't informative at all. What does "registered on cidr" mean? It's not clear from the documentation whether this function would only return a route that is an exact match or any route matching the argument.

Also, "CIDR" should be capitalized, it's an acronym.

Also, there's a missing newline.

src/iface/route.rs
```diff
@@ -74,16 +74,36 @@ impl<'a> Routes<'a> {
...
+    pub fn get_route(&mut self, cidr: &IpCidr) -> Option<&Route> {
+        self.storage.get(cidr)
+    }
+    /// Returns a mutable reference to the route registered by cidr
```
whitequark: Same as above.

src/iface/route.rs
```diff
@@ -74,16 +74,36 @@ impl<'a> Routes<'a> {
...
+    pub fn get_route_mut(&mut self, cidr: &IpCidr) -> Option<&mut Route> {
+        self.storage.get_mut(cidr)
+    }
+    /// Delete a route.
+    ///
+    /// On success, returns the value it deleted, if any.
```
whitequark: `returns the route it deleted, if any`

src/iface/route.rs
```diff
@@ -74,16 +74,36 @@ impl<'a> Routes<'a> {
...
+    pub fn del_route(&mut self, cidr: &IpCidr) -> Option<Route> {
```
whitequark: This function should be called `delete_route` (or `remove_route`; not `del_route`).

src/iface/route.rs
```diff
@@ -74,16 +74,36 @@ impl<'a> Routes<'a> {
         Routes { storage }
     }
 
-    /// Update the routes of this node.
-    pub fn update<F: FnOnce(&mut ManagedMap<'a, IpCidr, Route>)>(&mut self, f: F) {
-        f(&mut self.storage);
+    /// Add a new route.
+    ///
+    /// On success, return the previous route, if any.
+    pub fn add_route(&mut self, cidr: IpCidr, route: Route) -> Result<Option<Route>> {
```
whitequark: The naming of this function is inconsistent with `set_*_default_route`. Either this should be called `set_route`, or those functions should be called `add_*_default_route`. They fundamentally do the same thing.

src/iface/route.rs
```diff
@@ -74,16 +74,36 @@ impl<'a> Routes<'a> {
...
     }
 
-    /// Add a default ipv4 gateway (ie. "ip route add 0.0.0.0/0 via `gateway`").
+    /// Set the default ipv4 gateway (ie. "ip route add 0.0.0.0/0 via `gateway`").
     ///
     /// On success, returns the previous default route, if any.
     #[cfg(feature = "proto-ipv4")]
-    pub fn add_default_ipv4_route(&mut self, gateway: Ipv4Address) -> Result<Option<Route>> {
+    pub fn set_default_ipv4_route(&mut self, gateway: Ipv4Address) -> Result<Option<Route>> {
         let cidr = IpCidr::new(IpAddress::v4(0, 0, 0, 0), 0);
```
whitequark: This function should use `self.add_route`.

src/iface/route.rs
```diff
@@ -74,16 +74,36 @@ impl<'a> Routes<'a> {
...
+    }
+    /// Delete a route.
+    ///
+    /// On success, returns the value it deleted, if any.
+    pub fn del_route(&mut self, cidr: &IpCidr) -> Option<Route> {
+        self.storage.remove(&cidr)
     }
 
-    /// Add a default ipv4 gateway (ie. "ip route add 0.0.0.0/0 via `gateway`").
+    /// Set the default ipv4 gateway (ie. "ip route add 0.0.0.0/0 via `gateway`").
```
whitequark: `IPv4 gateway`

src/iface/route.rs
```diff
@@ -92,6 +112,22 @@ impl<'a> Routes<'a> {
         }
     }
 
+    /// Get the current default ipv4 gateway, if any
```
whitequark: `IPv4 gateway`

src/iface/route.rs
```diff
@@ -92,6 +112,22 @@ impl<'a> Routes<'a> {
...
+    #[cfg(feature = "proto-ipv4")]
+    pub fn get_default_ipv4_route(&mut self) -> Option<&Route> {
+        let cidr = IpCidr::new(IpAddress::v4(0, 0, 0, 0), 0);
```
whitequark: This function should use `self.get_route`.

src/iface/route.rs
```diff
@@ -92,6 +112,22 @@ impl<'a> Routes<'a> {
...
+        self.storage.get(&cidr)
+    }
+
+    /// Remove default ipv4 gateway.
+    ///
+    /// Returns the removed route, if any.
+    #[cfg(feature = "proto-ipv4")]
+    pub fn remove_default_ipv4_route(&mut self) -> Option<Route> {
```
whitequark: Once again, inconsistent naming with `del_route`.

src/iface/route.rs
```diff
@@ -92,6 +112,22 @@ impl<'a> Routes<'a> {
...
     }
 
+    /// Get the current default ipv4 gateway, if any
+    #[cfg(feature = "proto-ipv4")]
+    pub fn get_default_ipv4_route(&mut self) -> Option<&Route> {
+        let cidr = IpCidr::new(IpAddress::v4(0, 0, 0, 0), 0);
+        self.storage.get(&cidr)
+    }
+
+    /// Remove default ipv4 gateway.
```
whitequark: `IPv4 gateway`.

src/iface/route.rs
```diff
@@ -105,6 +141,23 @@ impl<'a> Routes<'a> {
         }
     }
 
+    /// Get the current default ipv6 gateways, if any
```
whitequark: `IPv6 gateways`

src/iface/route.rs
```diff
@@ -105,6 +141,23 @@ impl<'a> Routes<'a> {
...
+    #[cfg(feature = "proto-ipv6")]
+    pub fn get_default_ipv6_routes(&mut self) -> impl Iterator<Item = &Route> {
+        let cidr = IpCidr::new(IpAddress::v6(0, 0, 0, 0, 0, 0, 0, 0), 0);
+        // TODO: Support multiple default ipv6 gateways
```
whitequark: Why is this a TODO if you already return an iterator?

src/iface/route.rs
```diff
@@ -105,6 +141,23 @@ impl<'a> Routes<'a> {
...
+        self.storage.get(&cidr).into_iter()
+    }
+
+    /// Clear default ipv6 gateways.
```
whitequark: `IPv6 gateways`

src/iface/route.rs
```diff
@@ -105,6 +141,23 @@ impl<'a> Routes<'a> {
...
+    ///
+    /// On success, returns the cleared routes, if any.
```
whitequark: This function returns exactly one route.

src/iface/route.rs
```diff
@@ -105,6 +141,23 @@ impl<'a> Routes<'a> {
...
+    #[cfg(feature = "proto-ipv6")]
+    pub fn clear_default_ipv6_routes(&mut self) -> Option<Route> {
+        let cidr = IpCidr::new(IpAddress::v6(0, 0, 0, 0, 0, 0, 0, 0), 0);
+        self.storage.remove(&cidr)
```
whitequark: This function should use `self.remove_route`.

jD91mZM2:
Since you previously said you don't think this is needed, I'll just close this because I'm lazy :stuck_out_tongue:. Better now than later when I put in more time into something that isn't even desired.

PR #248, opened 2018-06-22
Log and print error for all examples

### LINKED ISSUE #231: Examples should not panic when poll returns an error
The examples currently panic when an error condition is hit (e.g. a unregognized packet is received). Per the [`poll` documentation] the example should not panic when an error is returned, rather the error should be logged.

CC: @podhrmic 
Related To: #228 #227 

[`poll` documentation]: https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs#L359-L368

podhrmic:
Closes #231


podhrmic:
FYI Travis fails because it cannot download rustc, not because of my commits.

whitequark:
Thanks!

PR #253, opened 2018-07-09
Implement TCP Window Scaling

### LINKED ISSUE #106: Implement TCP window scaling

jhwgh1968:
This should close #106 ~~once I put in a couple unit tests~~ (EDIT: they're done).

I would like an early review and a set of independent testing because I suspect this PR may cause future issues.

I did testing with the server example using buffers that were 10, 100, and 1000 times bigger, and did repeated 200 MB file transfers. The expanded windows did seem to behave correctly, and the file was correctly assembled at the other end.

However, my testing also suggested that this increased "speed limit" has made smoltcp quite timing sensitive, and is exposing unrelated edge cases and/or timing bugs.

The most annoying was when, on occasion, it would send a duplicate ACK every 20 *microseconds* between two received segments for reasons unknown. Another was more rare, but ominous. If segments arrived at a high enough speed with the CPU was starved, they would seem to get lost; the Assembler would read those that made it as "out of order". If the CPU remained starved for long enough, it was possible to exceed its tracking limit and hit a panic.

Both problems were intermittent, and I could not troubleshoot them because adding debug code made them go away. :unamused:  

Also, changes were needed to allow rx_buffer > TCP window, even before any of the actual window sizing changes. It makes me wonder if there are more rx_buffer == TCP window baked-in assumptions that I missed somewhere. I hope not. :crossed_fingers: 

And finally, it seems no one tried to run ``cargo bench`` lately, or they would have discovered my last patch broke it. :blush: That's fixed in this patch as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
     - rust: nightly
       env: FEATURES='std phy-tap_interface proto-ipv6 socket-udp' MODE='test'
     - rust: nightly
-      env: FEATURES='std proto-ipv4 socket-raw' MODE='test'
+      env: FEATURES='std proto-ipv4 proto-igmp socket-raw' MODE='test'
     - rust: nightly
       env: FEATURES='std proto-ipv6 socket-udp' MODE='test'
     - rust: nightly
@@ -33,7 +33,7 @@ matrix:
       env: FEATURES='proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp alloc'
         MODE='test'
     - rust: nightly
-      env: FEATURES='proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp'
+      env: FEATURES='proto-ipv4 proto-ipv6 proto-igmp socket-raw socket-udp socket-tcp socket-icmp'
         MODE='build'
     - rust: nightly
       env: MODE='fuzz run' ARGS='packet_parser -- -max_len=1536 -max_total_time=30'
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,17 +34,22 @@ verbose = []
 "phy-raw_socket" = ["std", "libc"]
 "phy-tap_interface" = ["std", "libc"]
 "proto-ipv4" = []
+"proto-igmp" = ["proto-ipv4"]
 "proto-ipv6" = []
 "socket-raw" = []
 "socket-udp" = []
 "socket-tcp" = []
 "socket-icmp" = []
+"proto-dhcpv4" = ["proto-ipv4"]
 default = [
   "std", "log", # needed for `cargo test --no-default-features --features default` :/
   "phy-raw_socket", "phy-tap_interface",
-  "proto-ipv4", "proto-ipv6",
+  "proto-ipv4", "proto-igmp", "proto-ipv6",
   "socket-raw", "socket-icmp", "socket-udp", "socket-tcp"
 ]
+
+# experimental; do not use; no guarantees provided that this feature will be kept
+"rust-1.28" = []
 
 [[example]]
 name = "packet2pcap"
@@ -76,8 +81,12 @@ name = "loopback"
 required-features = ["log", "proto-ipv4", "socket-tcp"]
 
 [[example]]
+name = "multicast"
+required-features = ["std", "phy-tap_interface", "proto-ipv4", "proto-igmp", "socket-udp"]
+
+[[example]]
 name = "benchmark"
-required-features = ["std", "phy-tap_interface", "proto-ipv4", "socket-tcp"]
+required-features = ["std", "phy-tap_interface", "proto-ipv4", "socket-raw", "socket-udp"]
 
 [profile.release]
 debug = 2
--- a/README.md
+++ b/README.md
@@ -47,6 +47,16 @@ The only supported medium is Ethernet.
   * IPv6 default gateway is **not** supported.
   * IPv6 extension headers are **not** supported.
 
+### IP multicast
+
+#### IGMP
+
+The IGMPv1 and IGMPv2 protocols are supported, and IPv4 multicast is available.
+
+  * Membership reports are sent in response to membership queries at
+    equal intervals equal to the maximum response time divided by the
+    number of groups to be reported.
+
 ### ICMP layer
 
 #### ICMPv4
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -61,7 +61,12 @@ fn main() {
     let mut tcp_active = false;
     loop {
         let timestamp = Instant::now();
-        iface.poll(&mut sockets, timestamp).expect("poll error");
+        match iface.poll(&mut sockets, timestamp) {
+            Ok(_) => {},
+            Err(e) => {
+                debug!("poll error: {}", e);
+            }
+        }
 
         {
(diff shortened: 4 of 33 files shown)


jhwgh1968:
Since @whitequark elsewhere said that he actually looks at the full diff, I thought I might as well squash these and skip a step if he approves.

jhwgh1968:
Now that I've finished the unit tests, I'm trying to characterize these timing issues more.

They are all reproduced by transferring a 200 MB file into the server example's sink, with a buffer size of 6553500 bytes. I used an old gparted ISO, which was exactly 255,852,544 bytes.

Today, I managed to reproduce a too-fast TCP timeout:

```
[1532887156.197s] (server): tcp:6971 recv 8 octets
[1532887156.744s] <- EthernetII src=8a-0f-8b-8c-c4-97 dst=02-00-00-00-00-01 type=IPv4
                \ IPv4 src=192.168.69.100 dst=192.168.69.1 proto=TCP
                 \ TCP src=50472 dst=6971 seq=3181168926 ack=1290742821 win=229 len=0
[1532887156.744s] (socket::tcp): #3:192.168.69.1:6971:192.168.69.100:50472: received a keep-alive or window probe packet, will send an ACK
[1532887156.744s] -> EthernetII src=02-00-00-00-00-01 dst=8a-0f-8b-8c-c4-97 type=IPv4
                \ IPv4 src=192.168.69.1 dst=192.168.69.100 proto=TCP
                 \ TCP src=6971 dst=50472 seq=1290742821 ack=3181168927 win=0 len=0
[1532887156.744s] (socket::tcp): #3:192.168.69.1:6971:192.168.69.100:50472: timeout exceeded
[1532887156.744s] (socket::tcp): #3:192.168.69.1:6971:192.168.69.100:50472: state=ESTABLISHED=>CLOSED
[1532887156.744s] (socket::tcp): #3:192.168.69.1:6971:192.168.69.100:50472: outgoing segment will abort connection
[1532887156.744s] (socket::tcp): #3:192.168.69.1:6971:192.168.69.100:50472: sending RST|ACK
```

Unfortunately, I cannot capture the complete log, because /tmp can't hold both it and the wireshark capture.

According to wireshark, that last ACK and the RST are separated by 15 microseconds. Prior to that was the following events:

44.179957 sender sends the first segment of a burst, seq=176882179, len=1446
44.180065 sender sends the last segment of a burst, seq=176942911, len=1446
44.180066 sender transfers 94 bytes, seq=176944357, which fills the receiver's window
44.201210 receiver sends ACK 176944451, indicates window length 0
44.747640 sender sends keep-alive (546 ms after zero window ack)
44.748426 receiver sends identical ACK, window length still zero
44.748841 receiver sends RST/ACK (4 ms after last ack)

It seems like if the buffer is slow to be drained by the application, there is a way for the connection to time out too fast. Any idea why this is? Shouldn't that keep-alive packet have kept it alive?

pothos:
@jhwgh1968 In my tests it works well except for buffer sizes which lead to a situation where a socket can't connect. Your test value 262143 works, 524288 works not, 1000000 works, 1048576 does not etc… I haven't looked into detailed stocket state yet, but maybe you already have an idea?

```
correct behavior for e.g. buffer size 2007152
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::tcp: #0:0.0.0.0:52126:1.0.0.1:80: state=CLOSED=>SYN-SENT
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::tcp: #0:0.0.0.0:52126:1.0.0.1:80: outgoing segment will update window
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::tcp: #0:0.0.0.0:52126:1.0.0.1:80: sending SYN
DEBUG 2018-08-06T07:46:44Z: smoltcp::iface::ethernet: address 192.168.1.1 not in neighbor cache, sending ARP request
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::meta: #0: neighbor 1.0.0.1 missing, silencing until t+3.0s
TRACE 2018-08-06T07:46:44Z: smoltcp::iface::neighbor: filled 192.168.1.1 => xx-…-xx (was empty)
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::meta: #0: neighbor 1.0.0.1 discovered, unsilencing
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::tcp: #0:0.0.0.0:52126:1.0.0.1:80: outgoing segment will update window
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::tcp: #0:0.0.0.0:52126:1.0.0.1:80: sending SYN
TRACE 2018-08-06T07:46:44Z: smoltcp::socket::tcp: #0:0.0.0.0:52126:1.0.0.1:80: received SYN|ACK

misbehavior for e.g. buffer size 2097152:
TRACE 2018-08-06T07:49:27Z: smoltcp::socket::tcp: #0:0.0.0.0:50008:1.0.0.1:80: state=CLOSED=>SYN-SENT
DEBUG 2018-08-06T07:49:30Z: smoltcp::iface::ethernet: address 192.168.1.1 not in neighbor cache, sending ARP request
DEBUG 2018-08-06T07:49:30Z: smoltcp::iface::ethernet: cannot dispatch response packet: unaddressable destination
poll result: Unaddressable
```

jhwgh1968:
I am able to reproduce the problem, @pothos. It seems the socket does not send data with 524288, but does with 524287.

Unfortunately, I cannot tell why. In fact, I cannot get the trace output you got by patching httpclient. What did you change to accomplish that?

pothos:
@jhwgh1968 I think e.g. for the tap interface setup as mentioned in the README it should be

```
cargo build --examples
RUST_LOG=trace target/debug/examples/httpclient tap0 1.1.1.1 http://1.1.1.1
```

jhwgh1968:
I think I have added all the unit tests, @whitequark. And sure enough, I think they found your bug, @pothos.

whitequark:
> And sure enough, I think they found your bug, @pothos.

What was the bug?

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit d23033a has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit d23033a90fe71cd8c8462db4f0462c7ae3d0cd42 with merge 92e970b3798737cfaa2a829d3f1bb7a7353696ee...

whitequark:
@jhwgh1968 How many holes do you think we should track?

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/418016631?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 92e970b3798737cfaa2a829d3f1bb7a7353696ee to master...

jhwgh1968:
I got consistently good results with 16, but I didn't spend too much time tuning it.

I plan to seeing how much sACK improves it first because I presume 4 was picked for a reason, e.g. memory consumption. If that is not the case, then increasing the number of holes would indeed solve the issue as far as I can tell.

whitequark:
> I plan to seeing how much sACK improves it first because I presume 4 was picked for a reason, e.g. memory consumption. If that is not the case, then increasing the number of holes would indeed solve the issue as far as I can tell.

I chose 4 completely arbitrarily. :)

PR #264, opened 2018-09-22
Increase number of assembler gaps to 32

pothos:
This is just a proposal, maybe a build feature is prefered? If yes, what values should be included?
Or should we rather introduce a decoupled scheme as established with the socket buffers?

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ The TCP protocol is supported over IPv4 and IPv6, and server and client TCP sock
   * Maximum segment size is negotiated.
   * Window scaling is negotiated.
   * Multiple packets are transmitted without waiting for an acknowledgement.
-  * Reassembly of out-of-order segments is supported, with no more than 4 gaps in sequence space.
+  * Reassembly of out-of-order segments is supported, with no more than 4 or 32 gaps in sequence space.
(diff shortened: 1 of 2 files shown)


pothos:
Rather an API change for Assembler and ManagedSlice or really just a Box for std/alloc?

pothos:
Seems that `cargo test` has a bug in evaluating the cfg macros: There is a compile error because some definitions are compiled with cfg std and some without!?

pothos:
Yes, but I'm surprised about the combination of code with differently evaluated cfg parts.

pothos:
Got my mistake, I didn't see that the same lines also were in `mod test`… and got confused by the warning.

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 2f9a5ba has been approved by `whitequark`

whitequark:
@pothos Sorry, I did not notice that you updated the branch.

m-labs-homu:
:hourglass: Testing commit 2f9a5babd45458b254eff0d31054d401d886f2e0 with merge ef014b118b19d2c4e9218be6ab52ffe4c2aecff3...

PR #266, opened 2018-11-12
Minimal Implementation of TCP Selective Acknowledgement

jhwgh1968:
This PR correctly handles all "must" requirements of RFC 2018:
- The sACK Permitted and sACK Range options are transmitted in accordance with the RFC.
- If the remote does not send sACK permitted, sACK range options are never transmitted.
- When a sACK range is transmitted, it contains the range of data corresponding to the segment that triggered the ACK unless the segment caused the ACK field to update (i.e. filled in an assembler hole that permitted progress).
- Once a sACK range is announced, that data is never thrown away. (This is not a "must" requirement, but if the TCP does this, then additional "must" requirements are added.)

However, since this is a "phase 1" PR, compliance is all this implementation has to offer. In practical terms, it only improves throughput in one case, and even then, only for data flowing in one direction.

This is because of the shortcuts taken to save significant complexity:
- All sACK information received from the remote side is ignored
- Only one sACK range will ever be transmitted, even though the data structures support up to 3
- If the segment causes an update in the ACK field, smoltcp will choose a "helpful hint" range if there are still holes in the assembler. The algorithm it uses to select this is contrary to a "should" in the RFC.

Each one of these is avoiding a significant effort. The first is a whole new data structure, the second two are because the "should" algorithm is based on keeping a history of transmitted sACK ranges on a queue, and pulling them off as holes are filled.

In its current state, this PR helps the case of a "burst" loss significantly, and Linux TCP seems to play nicely with the "helpful hint" range. But I would like more testing, if possible, before I consider this deviation from "should" safe to merge.

Extra testing and one unit test are all that is left before this is no longer WIP.

### DIFF
--- a/src/wire/tcp.rs
+++ b/src/wire/tcp.rs
@@ -103,6 +103,8 @@ mod field {
     pub const OPT_NOP: u8 = 0x01;
     pub const OPT_MSS: u8 = 0x02;
     pub const OPT_WS:  u8 = 0x03;
+    pub const OPT_SACKPERM: u8 = 0x04;
+    pub const OPT_SACKRNG:  u8 = 0x05;
 }
 
 impl<T: AsRef<[u8]>> Packet<T> {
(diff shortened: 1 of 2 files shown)


whitequark:
Thanks for the PR! Is there anything necessary for you to finish it?

jhwgh1968:
I reviewed my code, and I think the case I was thinking of is already covered.

Ready for review.

src/wire/tcp.rs
```diff
@@ -553,6 +595,47 @@ impl<'a> TcpOption<'a> {
...
+                    (field::OPT_SACKPERM, _) =>
+                        return Err(Error::Malformed),
+                    (field::OPT_SACKRNG, n) => {
+                        if n < 10 || (n-2) % 8 != 0 {
+                            return Err(Error::Malformed)
+                        }
+                        if n > 26 {
+                            // It's possible for a remote to send 4 SACK blocks, but extremely rare.
+                            // Better to "lose" that 4th block and save the extra RAM and CPU
+                            // cycles in the vastly more common case.
```
whitequark: This seems like a weird optimization, but since the RFC spells it out explicitly, I guess it's fine.

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 2552d0b has been approved by `whitequark`

whitequark:
@m-labs-homu r+

m-labs-homu:
:bulb: This pull request was already approved, no need to approve it again.

m-labs-homu:
:pushpin: Commit 2552d0b has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 2552d0b414dc88d23db32150f02d84b807fdbe85 with merge ed8dce015c45ee6e37c8d671f290a6d9b7fb49ac...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/474022451?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing ed8dce015c45ee6e37c8d671f290a6d9b7fb49ac to master...

PR #268, opened 2018-12-24
implement BPF sockets to be used as RawSocketDesc on BSD-like OSes

### LINKED ISSUE #71: not working on macOS
Line: https://github.com/m-labs/smoltcp/blob/master/src/phy/sys/raw_socket.rs#L21

hi, the `libc::AF_PACKET` constant only work on linux kernel.

1uka:
I added a file under `phy/sys` named `bpf.rs` and put all the code there so you can review the code easily. I also made other small changes so as to be able to build and run on macOS and FreeBSD with the feature `phy-raw_sockets`. I also named the methods of the `BpfDevice` struct same as in the `RawSocketDesc` struct, and reimport `BpfDevice` as `RawSocketDesc`.

Another point that can be discussed about my implementation is about the `interface_mtu` method (see the comment in the code).

I tested the code with the `tcpdump` example and it works seamlessly, I didn't test it with the other examples yet though.

I did everything on macOS, but from my experience with BPF sockets this should work the same on FreeBSD, and probably other BSD-like OSes.

PS: Probably resolves #71.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
+/.idea
 Cargo.lock
 *.pcap
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /target
-/.idea
 Cargo.lock
 *.pcap
--- a/src/phy/mod.rs
+++ b/src/phy/mod.rs
@@ -95,7 +95,7 @@ mod fault_injector;
 mod pcap_writer;
 #[cfg(any(feature = "std", feature = "alloc"))]
 mod loopback;
-#[cfg(all(feature = "phy-raw_socket", target_os = "linux"))]
+#[cfg(all(feature = "phy-raw_socket", unix))]
 mod raw_socket;
 #[cfg(all(feature = "phy-tap_interface", target_os = "linux"))]
 mod tap_interface;
@@ -108,7 +108,7 @@ pub use self::fault_injector::FaultInjector;
 pub use self::pcap_writer::{PcapLinkType, PcapMode, PcapSink, PcapWriter};
 #[cfg(any(feature = "std", feature = "alloc"))]
 pub use self::loopback::Loopback;
-#[cfg(all(feature = "phy-raw_socket", target_os = "linux"))]
+#[cfg(all(feature = "phy-raw_socket", unix))]
 pub use self::raw_socket::RawSocket;
 #[cfg(all(feature = "phy-tap_interface", target_os = "linux"))]
 pub use self::tap_interface::TapInterface;
--- a/src/phy/sys/bpf.rs
+++ b/src/phy/sys/bpf.rs
@@ -6,13 +6,16 @@ use libc;
 use super::{ifreq, ifreq_for};
 
 /// set interface
+#[cfg(target_os = "macos")]
 const BIOCSETIF: libc::c_ulong = 0x8020426c;
 /// get buffer length
+#[cfg(target_os = "macos")]
 const BIOCGBLEN: libc::c_ulong = 0x40044266;
 /// set immediate/nonblocking read
+#[cfg(target_os = "macos")]
 const BIOCIMMEDIATE: libc::c_ulong = 0x80044270;
-
 // TODO: check if this is same for OSes other than macos
+#[cfg(target_os = "macos")]
 const BPF_HDRLEN: usize = 18;
 
 macro_rules! try_ioctl {
@@ -69,8 +72,23 @@ impl BpfDevice {
     /// but it returns the size of the buffer that the app needs to allocate
     /// for the BPF device
     ///
-    /// The SIOGIFMTU cant be called on a BPF descriptor. There is a workaround
+    /// The `SIOGIFMTU` cant be called on a BPF descriptor. There is a workaround
     /// to get the actual interface mtu, but this should work better
+    ///
+    /// To get the interface MTU, you would need to create a raw socket first,
+    /// and then call `SIOGIFMTU` for the same interface your BPF device is "bound" to.
+    /// This MTU that you would get would not include the length of `struct bpf_hdr`
+    /// which gets prepended to every packet by BPF,
+    /// and your packet will be truncated if it has the length of the MTU.
+    ///
+    /// The buffer size for BPF is usually 4096 bytes, MTU is typically 1500 bytes.
+    /// You could do something like `mtu += BPF_HDRLEN`,
+    /// but you must change the buffer size the BPF device expects using `BIOCSBLEN` accordingly,
+    /// and you must set it before setting the interface with the `BIOCSETIF` ioctl.
+    ///
+    /// The reason I said this should work better is because you might see some unexpected behavior,
+    /// truncated/unaligned packets, I/O errors on read()
+    /// if you change the buffer size to the actual MTU of the interface.
     pub fn interface_mtu(&mut self) -> io::Result<usize> {
         let mut bufsize: libc::c_int = 1;
         try_ioctl!(self.fd, BIOCIMMEDIATE, &mut bufsize as *mut libc::c_int);
--- a/src/phy/sys/bpf.rs
+++ b/src/phy/sys/bpf.rs
@@ -1,0 +1,122 @@
+use std::io;
+use std::os::unix::io::{AsRawFd, RawFd};
+
+use libc;
+
+use super::{ifreq, ifreq_for};
+
+/// set interface
+const BIOCSETIF: libc::c_ulong = 0x8020426c;
+/// get buffer length
+const BIOCGBLEN: libc::c_ulong = 0x40044266;
+/// set immediate/nonblocking read
+const BIOCIMMEDIATE: libc::c_ulong = 0x80044270;
+
+// TODO: check if this is same for OSes other than macos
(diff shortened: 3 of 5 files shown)


src/phy/sys/bpf.rs
```diff
@@ -0,0 +1,122 @@
...
+impl AsRawFd for BpfDevice {
+    fn as_raw_fd(&self) -> RawFd {
+        self.fd
+    }
+}
+
+impl BpfDevice {
+    pub fn new(name: &str) -> io::Result<BpfDevice> {
+        let fd = unsafe {
+            let mut fd = -1;
```
dlrobertson: Would be more "rusty" to use an option here instead of using `-1` as the "unpopulated" value.

src/phy/sys/bpf.rs
```diff
@@ -0,0 +1,122 @@
...
+            for i in 0..100 {
```
dlrobertson: Why 100? I don't know bpf, but I have seen other scripts that go up to 255.

src/phy/mod.rs
```diff
@@ -95,7 +95,7 @@ mod fault_injector;
 mod pcap_writer;
 #[cfg(any(feature = "std", feature = "alloc"))]
 mod loopback;
-#[cfg(all(feature = "phy-raw_socket", target_os = "linux"))]
+#[cfg(all(feature = "phy-raw_socket", unix))]
```
dlrobertson: Currently it is possible that this `cfg` is met, but the `cfg` for the `bpf` is not met. E.g. `target_os = "netbsd"`.

Either this needs to be something like the following:

```rust
#[cfg(all(feature = "phy-raw_socket", any(target_os = "linux", target_os = "freebsd", target_os = "macos")))]
```

Or the `bpf` module needs something like the following `cfg`:

```rust
#[cfg(all(feature = "phy-raw_socket", not(target_os = "linux"), unix))]
```

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
+/.idea
```
whitequark: Please add this to your `$HOME/.gitignore`; smoltcp should not collect files of every IDE that exists.

dlrobertson (approved):
:+1: thanks

src/phy/sys/bpf.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        Ok(())
+    }
+
+    /// This in fact does not return the interface's mtu,
+    /// but it returns the size of the buffer that the app needs to allocate
+    /// for the BPF device
+    ///
+    /// The SIOGIFMTU cant be called on a BPF descriptor. There is a workaround
+    /// to get the actual interface mtu, but this should work better
```
whitequark: Why should this work better? This isn't clear to me.

src/phy/sys/bpf.rs
```diff
@@ -0,0 +1,129 @@
+use std::io;
+use std::os::unix::io::{AsRawFd, RawFd};
+
+use libc;
+
+use super::{ifreq, ifreq_for};
+
+/// set interface
+const BIOCSETIF: libc::c_ulong = 0x8020426c;
```
whitequark: Surely these hardcoded ioctl values are not valid for every Unix OS, are they?

1uka: I could not find out their values for other OSes, the most I can do is change this to only compile on macOS if that is what you are implying.

src/phy/sys/bpf.rs
1uka: To get the interface MTU, you would need to create a raw socket first, and then call `SIOGIFMTU` for the same interface your BPF device is "bound" to. Now this MTU that you would get would not include the length of `struct bpf_hdr` which gets prepended to every packet by BPF, and your packet will be truncated if it has the length of the MTU. 

The buffer size for BPF is usually 4096 bytes, MTU is typically 1500 bytes. You could do something like `mtu += BPF_HDRLEN`, but you must change the buffer size the BPF device expects using `BIOCSBLEN` accordingly.

The reason I said this should work better is because you might see some unexpected behavior, truncated/unaligned packets, I/O errors on `read()` if you change the buffer size to the actual MTU of the interface.

1uka: Yes, sure

whitequark:
Can you please add a Travis builder that uses macOS? After that I'll merge.

1uka:
Should I add one only for builds with the 'phy-raw_socket' feature or for all builds? I am not familiar with Travis CI, just now I am reading through the docs (specifically [this page](https://docs.travis-ci.com/user/multi-os))

whitequark:
I think one build with `phy-raw_socket` feature enabled, as everything else is not platform-dependent. Thanks!

1uka:
Sorry if I am annoying, there was an error when building only with 'phy-raw_socket', namely:

```rust
error[E0392]: parameter `'a` is never used
   --> src/iface/ethernet.rs:244:13
    |
244 | enum Packet<'a> {
    |             ^^ unused type parameter
    |
    = help: consider removing `'a` or using a marker such as `core::marker::PhantomData`
```

So instead I added a builder with the default features, since the above error was caused by the fact that every member of `Packet` is feature-dependent. I hope this is ok.

whitequark:
That's fine.

whitequark (approved):

whitequark:
Thanks!

1uka:
My pleasure.

JakkuSakura:
This may not always work on MacOS. At least on my machine, I have error code 22: invalid argument. I have to implement a raw socket with `pnet`.

```rust
use alloc::rc::Rc;
use std::cell::RefCell;
use std::time::Duration;

use eyre::{bail, ContextCompat};
use pnet::datalink::{Channel, Config, DataLinkReceiver, DataLinkSender};

use crate::{phy::{ChecksumCapabilities, Device, DeviceCapabilities, Medium, RxToken, TxToken},
            time::Instant};

pub struct PnetRawSocket {
	tx: Rc<RefCell<Box<dyn DataLinkSender>>>,
	rx: Rc<RefCell<Box<dyn DataLinkReceiver>>>,
	cap: DeviceCapabilities,
}
pub struct PnetRxToken {
	buffer: Vec<u8>,
}

impl RxToken for PnetRxToken {
	fn consume<R, F>(mut self, _timestamp: Instant, f: F) -> crate::Result<R>
	where
		F: FnOnce(&mut [u8]) -> crate::Result<R>,
	{
		f(&mut self.buffer[..])
	}
}
pub struct PnetTxToken {
	tx: Rc<RefCell<Box<dyn DataLinkSender>>>,
}
impl TxToken for PnetTxToken {
	fn consume<R, F>(self, _timestamp: Instant, len: usize, f: F) -> crate::Result<R>
	where
		F: FnOnce(&mut [u8]) -> crate::Result<R>,
	{
		let mut f = Some(f);
		let mut ret = Err(crate::Error::Illegal);
		self.tx
			.borrow_mut()
			.build_and_send(1, len, &mut |x| {
				ret = f.take().unwrap()(x);
			})
			.ok_or(crate::Error::Exhausted)?
			.map_err(|err| {
				net_warn!("error when sending to eth: {:?}", err);
				crate::Error::Illegal
			})?;
		ret
	}
}
impl PnetRawSocket {
	pub fn new(name: &str) -> eyre::Result<Self> {
		let ifc = pnet::datalink::interfaces()
			.into_iter()
			.filter(|x| x.name == name)
			.next()
			.with_context(|| format!("No interface named {}", name))?;
		let channel = pnet::datalink::channel(
			&ifc,
			Config {
				// to have nonblocking bahavior
				read_timeout: Some(Duration::from_micros(0)),
				..Default::default()
			},
		)?;
		match channel {
			Channel::Ethernet(tx, rx) => Ok(Self {
				tx: Rc::new(RefCell::new(tx)),
				rx: Rc::new(RefCell::new(rx)),
				cap: DeviceCapabilities {
					medium: Medium::Ethernet,
					max_transmission_unit: 1400,
					max_burst_size: None,
					checksum: ChecksumCapabilities::default(),
				},
			}),
			_ => bail!("unsupported channel type"),
		}
	}
}
#[allow(deprecated_where_clause_location)]
impl Device for PnetRawSocket {
	type RxToken<'a>
	where
		Self: 'a,
	= PnetRxToken;
	type TxToken<'a>
	where
		Self: 'a,
	= PnetTxToken;

	fn capabilities(&self) -> DeviceCapabilities {
		self.cap.clone()
	}

	fn receive(&mut self) -> Option<(Self::RxToken<'_>, Self::TxToken<'_>)> {
		let mut lower = self.rx.borrow_mut();
		let mut buffer = Vec::with_capacity(self.cap.max_transmission_unit);
		match lower.next() {
			Ok(size) => {
				buffer.extend_from_slice(size);
				let rx = PnetRxToken { buffer };
				let tx = PnetTxToken {
					tx: self.tx.clone(),
				};
				Some((rx, tx))
			}
			Err(ref err) if err.kind() == std::io::ErrorKind::TimedOut => None,
			Err(err) => {
				net_warn!("Error when receiving data: {:?}", err);
				None
			}
		}
	}

	fn transmit(&mut self) -> Option<Self::TxToken<'_>> {
		Some(PnetTxToken {
			tx: self.tx.clone(),
		})
	}
}

```

PR #279, opened 2019-03-12
tcp: mitigate illegal state transitions on simultaneous close

cbranch:
If a socket is closed and the remote side sends a FIN before the local
side sends its own FIN, it is possible to jump directly from FIN-WAIT-1
to CLOSING without ever sending a FIN to the remote side.

The main side effect is that if untransmitted data is present, the
socket is never exhausted, causing `EthernetInterface::poll` to loop
forever.

The ideal fix is to follow RFC more closely and not transition to
FIN-WAIT-1 until the FIN has been sent, but there are too many
assumptions based on the current state that would be broken by remaining
in ESTABLISHED with a closed transmit buffer to be worth fixing
a relatively rare edge case. Instead, add another illegal transition to
fix the failures of a previous violation, in the spirit of all good TCP
stacks.

---

The main reason why this is a problem at all is because `poll` performs ingress before egress, which is how this possibility can even arise. I haven't thought too hard about the implications of swapping them around. Maybe because, as already stated, this is a pretty rare occurrence to begin with.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -247,6 +247,8 @@ pub struct TcpSocket<'a> {
     /// The number of packets recived directly after
     /// each other which have the same ACK number.
     local_rx_dup_acks: u8,
+    /// Whether or not a FIN packet was sent after closing the transmit half of the socket.
+    sent_fin: bool,
 }
 
 const DEFAULT_MSS: usize = 536;
@@ -296,6 +298,7 @@ impl<'a> TcpSocket<'a> {
             local_rx_last_ack: None,
             local_rx_last_seq: None,
             local_rx_dup_acks: 0,
+            sent_fin:        false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


whitequark:
I acknowledge the issue, but I'll need to think about the fix.

whitequark:
Can you tell me a bit more about this fix? In particular you are talking about RST in the PR description, but I can't find how your code or tests use RST at all. Do you mean FIN?

cbranch:
Oof, that’s a dumb error. Please s/RST/FIN

Dirbaio:
Thank you for identifying this issue! This is definitely a bug in the state machine that needs fixing.

I believe the fix is incorrect though. Tracking whether we've sent a FIN is not enough: it may get lost, so we may have to retransmit it. Therefore we need to send FINs in the CLOSING state itself. Enabling the existing code to transmit FINs for CLOSING (in addition to FIN-WAIT-1 and LAST-ACK) fixes the issue both for the initial FIN and the retransmissions.

To speed things up I've opened #398 with that fix, so I'm closing this.

PR #284, opened 2019-04-09
Implement Fuzz Injector And TCP Header Fuzzer

jhwgh1968:
Just for fun, I recently tried running the fuzzer for packet parsing in smoltcp. After 2.5 billion iterations, it only hit 270 code paths and gave me no crashes.

That made me decide to fuzz the TCP implementation by having the guided fuzzer modify TCP headers in packets send to a live socket. But there was no support for that. This PR is the first step in that direction.

Most of the code is copied from the fault injector. It seemed simpler and cleaner to make a separate file than to try and make that support the hooks libfuzzer uses.

Since I am still working on the actual fuzzing harness, this code is currently unused and is ``#[allow]``ed to be.

### DIFF
--- a/fuzz/fuzz_targets/tcp_headers.rs
+++ b/fuzz/fuzz_targets/tcp_headers.rs
@@ -1,0 +1,209 @@
+#![no_main]
+#[macro_use] extern crate libfuzzer_sys;
+extern crate smoltcp;
+
+use std as core;
+extern crate getopts;
+
+use core::cmp;
+use smoltcp::phy::Loopback;
+use smoltcp::wire::{EthernetAddress, EthernetFrame, EthernetProtocol};
+use smoltcp::wire::{IpAddress, IpCidr, Ipv4Packet, Ipv6Packet, TcpPacket};
+use smoltcp::iface::{NeighborCache, EthernetInterfaceBuilder};
+use smoltcp::socket::{SocketSet, TcpSocket, TcpSocketBuffer};
+use smoltcp::time::{Duration, Instant};
+
+mod utils {
+    include!("../utils.rs");
+}
+
+mod mock {
+    use std::sync::Arc;
+    use std::sync::atomic::{Ordering, AtomicUsize};
+	use smoltcp::time::{Duration, Instant};
+
+    // should be AtomicU64 but that's unstable
+    #[derive(Debug, Clone)]
+    pub struct Clock(Arc<AtomicUsize>);
+
+    impl Clock {
+        pub fn new() -> Clock {
+            Clock(Arc::new(AtomicUsize::new(0)))
+        }
+
+        pub fn advance(&self, duration: Duration) {
+            self.0.fetch_add(duration.total_millis() as usize, Ordering::SeqCst);
+        }
+
+        pub fn elapsed(&self) -> Instant {
+            Instant::from_millis(self.0.load(Ordering::SeqCst) as i64)
+        }
+    }
+}
+
+struct TcpHeaderFuzzer([u8; 56], usize);
+
+impl TcpHeaderFuzzer {
+    // The fuzzer won't fuzz any packets with the SYN flag set in order to make sure the connection
+    // is established before the fuzzed headers arrive.
+    //
+    // It will also not fuzz the source and dest port so it reaches the open socket.
+    //
+    // Otherwise, it replaces the entire rest of the TCP header with the fuzzer's output.
+    pub fn new(data: &[u8]) -> TcpHeaderFuzzer {
+        let copy_len = cmp::min(data.len(), 56 /* max TCP header length without port numbers*/);
+
+        let mut fuzzer = TcpHeaderFuzzer([0; 56], copy_len);
+        fuzzer.0[..copy_len].copy_from_slice(&data[..copy_len]);
+        fuzzer
+    }
+}
+
+impl smoltcp::phy::Fuzzer for TcpHeaderFuzzer {
+    fn fuzz_packet(&self, frame_data: &mut [u8]) {
+        if self.1 == 0 {
+            return;
+        }
+
+        let tcp_packet_offset = {
+            let eth_frame = EthernetFrame::new_unchecked(&frame_data);
+            EthernetFrame::<&mut [u8]>::header_len() + match eth_frame.ethertype() {
+                EthernetProtocol::Ipv4 =>
+                    Ipv4Packet::new_unchecked(eth_frame.payload()).header_len() as usize,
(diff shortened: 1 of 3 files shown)


whitequark:
Looks really nice! I'm going to merge this with the actual fuzz harness.

jhwgh1968:
I now have the fuzzer working!

But it seems very slow to find more than one code path. I will run it for a couple days to make sure I haven't done something wrong.

jhwgh1968:
Sure enough, it wasn't actually fuzzing. I don't know why, but the actual code in the FuzzInjector is not being triggered.

@whitequark, I copied the loopback example to write this code. Can you see something obvious I did wrong?

jhwgh1968:
Okay, I got it now fuzzing one packet... then stuck in an infinite loop on ``iface.poll()``, because the two sockets on the loopback interface constantly trade out-of-range ACKs without even returning to the caller. Since that is how the clock advances, it never times out.

If I can't think of anything in the next couple days, I may have to give up on this... :disappointed:

jhwgh1968:
I finally figured it out.

There is still some ugliness and hackiness, but I will let @whitequark or others be the judge.

fuzz/fuzz_targets/tcp_headers.rs
```diff
@@ -0,0 +1,209 @@
...
+            .ethernet_addr(EthernetAddress::default())
+            .neighbor_cache(neighbor_cache)
+            .ip_addrs(ip_addrs)
+            .finalize();
+
+    let server_socket = {
+        // It is not strictly necessary to use a `static mut` and unsafe code here, but
+        // on embedded systems that smoltcp targets it is far better to allocate the data
+        // statically to verify that it fits into RAM rather than get undefined behavior
+        // when stack overflows.
```
whitequark: I'm confused. The fuzzer is not an embedded target, is it?

src/phy/fuzz_injector.rs
```diff
@@ -0,0 +1,120 @@
+use Result;
+use phy::{self, DeviceCapabilities, Device};
+use time::Instant;
+
+// This could be fixed once associated consts are stable.
+const MTU: usize = 1536;
+
+/// Represents a fuzzer. It is expected to either generate or contain data, which it replaces bytes
```
whitequark: Contain data?

src/phy/fuzz_injector.rs
```diff
@@ -0,0 +1,120 @@
...
+            token:   token,
+        })
+    }
+}
+
+#[doc(hidden)]
+pub struct RxToken<'a, Rx: phy::RxToken, F: Fuzzer>{
+    fuzzer: &'a F,
+    token:   Rx,
+    fuzzed: [u8; MTU]
```
whitequark: Is this buffer actually necessary? Why?

src/phy/fuzz_injector.rs
```diff
@@ -0,0 +1,120 @@
...
+#[derive(Debug)]
+pub struct FuzzInjector<D: for<'a> Device<'a>, FTx: Fuzzer, FRx: Fuzzer> {
+    inner:   D,
+    fuzz_tx: FTx,
+    fuzz_rx: FRx,
+}
+
+#[allow(unused)]
+impl<D: for<'a> Device<'a>, FTx: Fuzzer, FRx: Fuzzer> FuzzInjector<D, FTx, FRx> {
+    /// Create a fuzz injector device, which defaults to passing the packets unaltered.
```
whitequark: This doesn't look right--it doesn't default to anything since it's not configurable...

whitequark:
Seems promising!

fuzz/fuzz_targets/tcp_headers.rs
jhwgh1968: This was copy-pasted from the loopback example.

src/phy/fuzz_injector.rs
jhwgh1968: That was a comment from a previous version. I will remove it.

src/phy/fuzz_injector.rs
jhwgh1968: The buffer is necessary because the buffer passed into the RxToken.consume is not mutable. I tried really hard to eliminate it, but the closure simply would not support it.

The fixed MTU was copied from the loopback example.

src/phy/fuzz_injector.rs
jhwgh1968: I was referring to the behavior of the Fuzz object. I will clarify this.

src/phy/fuzz_injector.rs
whitequark: Hmm, maybe we should make it mutable?

jhwgh1968: You mean change the trait? Because the ``F`` type on the trait is ``FnOnce(&[u8])``, which is why I can't mutate the data.

That would mean a bunch of little changes scattered everywhere, not to mention any multi-threading consequences...

whitequark: > That would mean a bunch of little changes scattered everywhere, not to mention any multi-threading consequences...

Yeah. Would you mind making this change? I think there's no especially good reason RxToken doesn't already present a mutable buffer. In this context it is better thought of as "exclusive" moreso than "mutable"; any network buffer would be consumed by only one network stack anyway, and making it mutable would potentially allow doing some interesting things with it, like the equivalent of a "mangle" firewall table.

jhwgh1968: Opened separate PR #294

jhwgh1968:
Since my other PR merged, I think I have addressed all your review comments, @whitequark. If you don't object to anything else -- e.g. copying a bunch of the utils file from the examples to the fuzzing code -- then I think it is ready to merge.

jhwgh1968:
Ping @whitequark, is there anything else I need to fix?

whitequark:
I'll take another look soon!

whitequark:
Sorry for the delay.
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 158891f has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 158891fe19934877561b8d26f9be97417c603f5e with merge a315dd1e98ecdbe97e3dfb4dff032c8c4ff501ce...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/534299442?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing a315dd1e98ecdbe97e3dfb4dff032c8c4ff501ce to master...

Issue #281, opened 2019-03-22
Ethernet denial of service


197g:
# Bug report

A malicious sender over an ethernet interface can cause a denial of service by making every call to `poll` return an error before `process_egress` could have been called. This will cause the interface to never send any packets that are pending on any socket.

The same happens inwittingly when the ingress call can constantly keep the buffer of a socket filled, for example by saturating line rate of a fast enough nic with a packet generator.

## Reproduction steps

1. Create a `phy::Device` that has an infinite stream of packets and somehow signals when a packet would have been sent (for example by panicking).
2. Wrap that device into an interface, setup routing to be valid for layers eth, ip, udp
2. Enqueue a packet to transmit on a UdpSocket within a SocketSet
3. `loop` infinitely with `iface.poll(&mut socketset)`
  3.1 Each loop, you can dequeue packets received, or not, doesn't matter
4. Observe that the send code is never triggered

Instead of having an infinite stream of packets read each time poll is called, you can alternatively implement the `phy` such that exactly one malformed packet is ready each time. This has the same effect on the poll-loop.

I have code ready for this, should I link it or not?

## Fixes

There are several possible ways to fix this:
* Collect the result of both `ingress` and `egress`, for example by removing the `?` from the [their initial call point](https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs#L468-L469) and instead 
  ```
  let ingress = self.socket_ingress(sockets, timestamp);
  let egress = self.socket_egress(sockets, timestamp);
  let processed_any = ingress? | egress?;
  ```
  This has the risk of losing the egress error information if ingress errored as well.
* Store a 'spin'-bool within `EthernetInterface` that determines whether to ingress or egress first. Flip it everytime `poll` is invoked.

Both of the above are still unsatisfying, as each `poll` call will at most egress a single packet before evaluating the `Err` from the next `socket_ingress` call and returning. Note that `process_ingress` will loop while there are `RxToken` returned from the `phy` while `proces_egress` will try to send a single packet from each socket within the set. This imbalance is quite confusing.

## Workarounds

[...]

PR #285, opened 2019-04-11
[WIP. Don't merge] Resilience to error in packet decoding

197g:
Addresses #281. Work in progress. Addresses DOS by errors in the decoding step by establishing an error ring-buffer associated to the ethernet interface and socket type. When some decoding step fails, the error is put into the most specific error buffer available but otherwise ignored. This ensures that faulty packets *can not* affect further packet processing. **This is a work in progress, intended for an early preview of ideas. As such, it may be incomplete, not compile at times and/or be buggy. I may possibly rebase or force-push the branch at will.**

A second step of the mechanism design rework will try to rebalance the recv/send inbalance in case of  a saturated `phy`. As layed out in the issue report, a full receive buffer can adversely affect send performance by blocking further sends. The error from the full buffer should break the receive part of the loop but it's not completely clear yet how to handle finally handle it. One possibility is putting it into the ring buffer or the ethernet iface, but this of course creates some friction with the return value a caller expects.

## Alternatives

Callbacks have been evaluated as inferior for major reasons:
* It's missing contextual information resulting from partial decoding. A failed udp decoding could for example still access and log the frames around itself. Re-decoding within the error callback is both inefficient, creates ambiguity and is awkward as it is effectively invoking a processing step that it expects to fail. This seems highly unintuitive.
* Invoking the callback *during* polling means that constraints such as terminiation, timing, resource usage, ... need to be fulfilled by arbitrary user code. Without an extremely extensive documentation, it remains unclear how one could expect the programmer of the callback to know the internals enough to make good engineering decisions about its design.
* The callback must fight the borrow checker if it wants to inspect any state such as the socket on which the error originated or the `phy` interface. Making error inspection a separate call makes this cleaner, less constrained and more efficient.
* Ignoring most errors, as many usages likely want to, would become a no-op for the caller.

## TODO

- [ ] Cleanup some of the changes that were made just to quickly iterate over alternative designs for this feature. This is still ongoing

- [ ] Make the code compile under a subset of features such as (`socket-udp`, `proto-ipv4`). This is actually an issue with the current library and I don't see any guide on which feature set should be supported other than `all` and `default`.

- [ ] Implement error buffers for more than `udp` and decide on a strategy to handle left-over errors that don't belong to any socket.

 - [ ] Evaluate and choose the strategy for recv/send imbalance.

- [ ] Implement tests to simulate and ensure resilience against the DOS potential presented in #281.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -1934,7 +1934,7 @@ mod test {
             0x6f, 0x2c, 0x20, 0x57,
             0x6f, 0x6c, 0x64, 0x21
         ];
-        let (iface, mut socket_set) = create_loopback();
+        let (mut iface, mut socket_set) = create_loopback();
 
         let mut udp_bytes_unicast = vec![0u8; 20];
         let mut udp_bytes_broadcast = vec![0u8; 20];
@@ -2018,7 +2018,7 @@ mod test {
 
         static UDP_PAYLOAD: [u8; 5] = [0x48, 0x65, 0x6c, 0x6c, 0x6f];
 
-        let (iface, mut socket_set) = create_loopback();
+        let (mut iface, mut socket_set) = create_loopback();
 
         let rx_buffer = UdpSocketBuffer::new(vec![UdpPacketMetadata::EMPTY], vec![0; 15]);
         let tx_buffer = UdpSocketBuffer::new(vec![UdpPacketMetadata::EMPTY], vec![0; 15]);
@@ -2097,7 +2097,7 @@ mod test {
         #[cfg(feature = "proto-ipv6")]
         const MAX_PAYLOAD_LEN: usize = 1192;
 
-        let (iface, mut socket_set) = create_loopback();
+        let (mut iface, mut socket_set) = create_loopback();
 
         #[cfg(all(feature = "proto-ipv4", not(feature = "proto-ipv6")))]
         let src_addr = Ipv4Address([192, 168, 1, 1]);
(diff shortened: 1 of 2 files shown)


whitequark:
It's an interesting approach. It is a bit awkward, and I don't like that. However, I agree that it seems like our best way going forward, so let's use it.

I think your lifetime changes would need to be a part of a separate PR. They change the public interface as well. I haven't looked at them closely yet as it's hard to follow here.

HeroicKatora:
Yes, it's all a bit raw and several changes are mixed together. I plan to split this up into several independent chunks once I have determined the overall usability.

HeroicKatora:
@whitequark Cleanly rebased only the error buffer related changes. Note that the lifetime within `iface` and `udp` is currently linked to other attributes as I wasn't sure what the original motivation was for having some of them overlap and some of them exposed in the generic parameters. Some feedback on this would be helpful. And, this still needs to be performed for the other socket types to ensure that really no error is generated. I finally may want to remove the `Result<_, Error>` part from the inner methods to make this explicit.

m-labs-homu:
:umbrella: The latest upstream changes (presumably 7c606acdd75853ebc8cf2ece79e7498b7b285a08) made this pull request unmergeable. Please resolve the merge conflicts.

HeroicKatora:
Just as a head up: The require changes are far too extensive to justify working on for me. I'm going to abandon this PR for good.

whitequark:
I see. That's a bit unfortunate.

PR #286, opened 2019-04-12
Add ip version specific addr/endpoint converters

197g:
Adds some ip-version specific `std` converter implementations to complement the existing ones. This is part of a larger effort to enable building with a mutually exclusive combination of the `proto-ipv4` and `proto-ipv6` features.

### DIFF
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -224,6 +224,20 @@ impl From<::std::net::IpAddr> for Address {
     }
 }
 
+#[cfg(all(feature = "std", feature = "proto-ipv4"))]
+impl From<::std::net::Ipv4Addr> for Address {
+    fn from(ipv4: ::std::net::Ipv4Addr) -> Address {
+        Address::Ipv4(ipv4.into())
+    }
+}
+
+#[cfg(all(feature = "std", feature = "proto-ipv6"))]
+impl From<::std::net::Ipv6Addr> for Address {
+    fn from(ipv6: ::std::net::Ipv6Addr) -> Address {
+        Address::Ipv6(ipv6.into())
+    }
+}
+
 impl Default for Address {
     fn default() -> Address {
         Address::Unspecified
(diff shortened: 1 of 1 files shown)


whitequark:
> This is part of a larger effort to enable building with a mutually exclusive combination of the `proto-ipv4` and `proto-ipv6` features.

I don't understand what this means.

HeroicKatora:
The other `*ip` and `*Endpoint` converters from standard library types are only enabled when both `proto-ipv4` and `proto-ipv6` are enabled. Thus one needs to manually unstructure the std-types and reconstruct them for `smoltcp`. A couple more of these instances are scattered around the library, making it needlessly hard to work with the lib when only one of the ip-versions is enabled. Also note that without the `std` feature, the library doesn't even build in some instances. There is no further guide other than the `.travis.yml` file on which feature combinations are supposed to work.

whitequark:
Ack. Can you write more descriptive commit messages? I don't like when the history consists of commits with confusing one-liners that explain nothing about the rationale.

HeroicKatora:
@whitequark Is this the style you are looking for? If so, I suppose I will perform the same for the other set of changes.

whitequark:
Yep, thanks!

@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 6928cb1 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 6928cb19e0975749c6cdc2a562cf5bbd4dae99a2 with merge e2fcd02414236e2c3f6f2ae2aed9dce73e4c4524...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/519561798?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing e2fcd02414236e2c3f6f2ae2aed9dce73e4c4524 to master...

PR #288, opened 2019-04-14
Fix IPv4 broadcast handling

### LINKED ISSUE #287: ICMPv4 ignores broadcast echo requests
I'm not sure if this is deliberate or not. The [current code](https://github.com/m-labs/smoltcp/blob/master/src/iface/ethernet.rs#L1245) is:

```rust
        if ipv4_repr.dst_addr.is_unicast() {
            let ipv4_reply_repr = Ipv4Repr {
                // snip...
            };
            Packet::Icmpv4((ipv4_reply_repr, icmp_repr))
        } else {
            // Do not send any ICMP replies to a broadcast destination address.
            Packet::None
        }
```

The comment suggests we're avoiding _transmitting_ broadcast replies, but the test is on `dst_addr`, which is about whether we're _receiving_ broadcast requests. We definitely don't want to send a broadcast reply, but for that we should check `src_addr` is unicast; as it stands I think we would reply to a spoofed broadcast `src_addr`.

We might or might not want to reply to a broadcast request. If we did we'd need to grab a suitable interface IP for the source of the response (we can't use the incoming `dst_addr` as it's broadcast).

In any event I guess this check should probably be changed; I'm happy to make a quick PR for that if you agree.

adamgreig:
A regression [here](https://github.com/m-labs/smoltcp/commit/58a54730dba398aaa5ea88e6c6f89e0234f89f42#diff-e4835a196a3b800b97af9b79bf187f33L708) meant broadcast IPv4 packets were ignored. Restore
handling those packets and also reply to broadcast ICMPv4 echo requests.

Closes #287.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -941,8 +941,10 @@ impl<'b, 'c, 'e> InterfaceInner<'b, 'c, 'e> {
         #[cfg(feature = "socket-raw")]
         let handled_by_raw_socket = self.raw_socket_filter(sockets, &ip_repr, ip_payload);
 
-        if !self.has_ip_addr(ipv4_repr.dst_addr) && !self.has_multicast_group(ipv4_repr.dst_addr) {
-            // Ignore IP packets not directed at us or any of the multicast groups
+        if !self.has_ip_addr(ipv4_repr.dst_addr) &&
+           !ipv4_repr.dst_addr.is_broadcast() &&
+           !self.has_multicast_group(ipv4_repr.dst_addr) {
+            // Ignore IP packets not directed at us, or broadcast, or any of the multicast groups.
             // If AnyIP is enabled, also check if the packet is routed locally.
             if !self.any_ip {
                 return Ok(Packet::None);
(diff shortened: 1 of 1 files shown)


whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit ad65e2a has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit ad65e2a463dae4fc5fdd8924dda74f8adfdf977b with merge 0a21cb609da34b52fafe24e7b78d357b7ce1f6b7...

m-labs-homu:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/519974876?utm_source=github_status&utm_medium=notification)

whitequark:
@m-labs-homu retry

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit ed93a77 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit ed93a77a460bb1f68f0aad0b4f28a7d0766384df with merge 5334b93158c954b61f5daf0d6d5d9d0c91588a26...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/519986118?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 5334b93158c954b61f5daf0d6d5d9d0c91588a26 to master...

PR #289, opened 2019-04-24
 Allow access to phy-device through the ethernet stack

197g:
Adds accessor functions to the device implementing the phy::Device trait
within an ethernet interface. The intent is to allow access to read-only
methods such as gathering of device statistics or other diagnostics
while the interface is running.

Additionally, `Deref` is implemented for `Tracer`. This makes the `Tracer` 
wrapper for a phy device a drop-in replacement for the encapsulated
device. All reference based methods of the traced device stay callable, 
only those that take self by value need an explicit unwrap with `into_inner`.

I'd appreciate some feedback on the latter change. There are no 
expected collisions with the methods of the inner device as `into_inner`
is the only method provided by tracer. However, the Rust api guidelines
would suggest [making `into_inner` a non-inherent method](https://rust-lang-nursery.github.io/api-guidelines/predictability.html#smart-pointers-do-not-add-inherent-methods-c-smart-ptr)
and it could be surprising to view `Tracer` as a smart pointer which would
suggest that explicit access functions instead of `Deref` are also suitable.

### DIFF
--- a/src/phy/tracer.rs
+++ b/src/phy/tracer.rs
@@ -19,6 +19,21 @@ impl<D: for<'a> Device<'a>, P: PrettyPrint> Tracer<D, P> {
         Tracer { inner, writer }
     }
 
+    /// Get a reference to the underlying device.
+    ///
+    /// Even if the device offers reading through a standard reference, it is inadvisable to
+    /// directly read from the device as doing so will circumvent the tracing.
+    pub fn get_ref(&self) -> &D {
+        &self.inner
+    }
+
+    /// Get a mutable reference to the underlying device.
+    ///
+    /// It is inadvisable to directly read from the device as doing so will circumvent the tracing.
+    pub fn get_mut(&mut self) -> &mut D {
+        &mut self.inner
+    }
+
     /// Return the underlying device, consuming the tracer.
     pub fn into_inner(self) -> D {
         self.inner


src/iface/ethernet.rs
```diff
@@ -353,6 +353,22 @@ impl<'b, 'c, 'e, DeviceT> Interface<'b, 'c, 'e, DeviceT>
         InterfaceInner::check_ethernet_addr(&self.inner.ethernet_addr);
     }
 
+    /// Get a reference to the inner device.
+    pub fn phy(&self) -> &DeviceT {
```
whitequark: Nit: I think this should be called `device`, since you aren't necessarily retrieving a phy.

whitequark:
I feel like it would be cleaner to not use `Deref` for `Tracer`. I originally thought that adding it is no big deal because `Tracer` is relatively rarely used, but then an explicit method is just as fine.

src/iface/ethernet.rs
HeroicKatora: I didn't know on which part of the trait `phy::Device` to base the naming here, but that is no problem to change.

whitequark: Yeah, understandable, that entire part of hierarchy is a bit weird.

HeroicKatora:
> I feel like it would be cleaner to not use Deref for Tracer. I originally thought that adding it is no big deal because Tracer is relatively rarely used, but then an explicit method is just as fine.

Hm, I think I see what you're saying. Some common trait (`AsRef` is an alternative) would be nice in any case, to allow writing code where enabling tracing is a matter of (un-)commenting one line or where it could be enabled with a feature flag.

whitequark:
As is you need to uncomment (or gate) two lines, which doesn't seem that bad to me?

HeroicKatora:
If the access to the actual underlying device is not possible with a trait, one change is necessary at every point where it is accessed that way. With a trait, the methods can be written polymorphically to take the type change into account. `Borrow` would is maybe smarter as `impl Borrow<T> for T` is available. Then this should work:

```
impl MyPhy {
    fn received(&self) -> usize { ... }
}

/// Would work with both:
/// * `EthernetInterface<MyPhy>`
/// * `EthernetInterface<EthernetTracer<MyPhy>>`
fn stats_for<T: Borrow<MyPhy>>(ether: &EthernetInterface<T>) -> usize {
    ether.device().borrow().received()
}
```

HeroicKatora:
But, it's only debugging and not critical functionality. I think we can also just add the base accessors and see if there are issues with the usage where this becomes more annoying.

whitequark:
Base accessors?

HeroicKatora:
I simply mean:
```
fn inner(&self) -> &Device {
    &self.inner
}
```
consistent with `Tracer` having `into_inner(self) -> Device`

HeroicKatora:
I think the name should be `get_ref` and `get_mut`, which is what the standard library uses primarily for types that have `into_inner`, such as `BufReader`, `io::Take`, ...

whitequark:
OK

whitequark:
@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 2e0efc8 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 2e0efc8c4d882973fdcdc7e3d64e3e7f78ae2e1c with merge acbe139e361f95a222a6f2030778492acf7dc0f3...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/524924898?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing acbe139e361f95a222a6f2030778492acf7dc0f3 to master...

PR #301, opened 2019-07-23
Restore Session::finish call in socket ref

197g:
Using move semantics allows an Option to keep track of the
initialization state while satisfying the borrow checker. This replaces
the functionality that the 'consumed' flag had before. It also retains
the smaller object size since the Option of a reference can use the
representation of the null pointer, which is an invalid reference, as a
niche for the None variant.

### DIFF
--- a/src/socket/ref_.rs
+++ b/src/socket/ref_.rs
@@ -32,7 +32,12 @@ impl<'a> Session for TcpSocket<'a> {}
 ///
 /// Allows the network stack to efficiently determine if the socket state was changed in any way.
 pub struct Ref<'a, T: Session + 'a> {
-    socket: &'a mut T
+    /// Reference to the socket.
+    ///
+    /// This is almost always `Some` except when dropped in `into_inner` which removes the socket
+    /// reference. This properly tracks the initialization state without any additional bytes as
+    /// the `None` variant occupies the `0` pattern which is invalid for the reference.
+    socket: Option<&'a mut T>,
 }
 
 impl<'a, T: Session + 'a> Ref<'a, T> {
@@ -42,7 +47,7 @@ impl<'a, T: Session + 'a> Ref<'a, T> {
     ///
     /// [into_inner]: #method.into_inner
     pub fn new(socket: &'a mut T) -> Self {
-        Ref { socket }
+        Ref { socket: Some(socket) }
     }
 
     /// Unwrap a smart pointer to a socket.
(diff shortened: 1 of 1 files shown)


whitequark:
Right, I knew this is a potential solution but it irritates me on a fundamental level, because there's no inherent safety issue here, just Rust being fussy about destructuring values that impl Drop. I even wrote [an RFC](https://internals.rust-lang.org/t/pre-rfc-destructuring-values-that-impl-drop/10450) about this.

whitequark:
Arguably, that branch would not matter much, and I'll merge this if someone makes a convincing argument about that. I just hate it because there's no good reason Rust has to force this overhead on me.

HeroicKatora:
If you really see this as a performance argument, then I urge you to define some benchmarks for it. It shouldn't make any noticable difference in practice, that's at most 2 micro-ops on a perfectly predicted branch. The `unwrap` failure is even marked as `unlikely` in the Rust standard library and due to uniqueness of `&mut ` the compiler is perfectly able to hoist the check outside of loops, etc. It's *as far* from paying for discarded instruction pipelining as it can be.

Should you intend to, nevertheless, not re-instantiate the `Session::finish` call, that's alright and you can close the PR but maybe think about removing the function altogether. As far as I can tell it's not even implemented anywhere? That would be much less confusing.

whitequark:
@HeroicKatora I see it more as a code size argument, since most targets smoltcp runs on don't even have micro-ops or branch prediction. But... there's probably lots of lower hanging fruit. So: I'm convinced.

@m-labs-homu r+

m-labs-homu:
:pushpin: Commit 108f72d has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 108f72d439bcc92baddcc5deb2aa3b41a9441d21 with merge efdde337b912dc69ae48a9885d0a762549ae1667...

whitequark:
> As far as I can tell it's not even implemented anywhere? That would be much less confusing.

The idea behind this function is to set up a poll only on sockets that were actually updated, as opposed to re-polling every socket on each call to `iface.poll`. It never quite materialized, but I think keeping the option open is good?.. Not sure.

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/562560428?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing efdde337b912dc69ae48a9885d0a762549ae1667 to master...

PR #303, opened 2019-08-02
Mac

kornelski:
edit: oops, I've meant to make PR for a different fork :)

### DIFF
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -61,6 +61,12 @@ fn client(kind: Client) {
 
 static CLIENT_DONE: AtomicBool = ATOMIC_BOOL_INIT;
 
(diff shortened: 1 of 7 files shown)


whitequark:
@kornelski Which fork?

kornelski:
https://github.com/cbranch/smoltcp

pothos:
Nice to see some work is done and even in the open. @cbranch and @kornelski, maybe you can give some context? Do you want to upstream the patches here someday or mind if others are doing it since you kept the license?

whitequark:
I am guessing the problem is that I don't have enough review bandwidth.

pothos:
That's probably one factor in general but the current situation is not terribly bad and could be much worse – you did quite well ;)

Issue #223, opened 2018-05-31
Unable to create UDP and TCP sockets


crawford:
I'm trying to create two listening sockets, one UDP and one TCP, with the following code:

```rust
    let mut udp_rx_payload = [0; 128];   
    let mut udp_rx_metadata = [PacketMetadata::EMPTY; 1];
    let mut udp_tx_payload = [0; 128];
    let mut udp_tx_metadata = [PacketMetadata::EMPTY; 1];
    let udp_socket = UdpSocket::new(
        UdpSocketBuffer::new(udp_rx_metadata.as_mut(), udp_rx_payload.as_mut()),
        UdpSocketBuffer::new(udp_tx_metadata.as_mut(), udp_tx_payload.as_mut()),
    );                        

    let mut tcp_rx_payload = [0; 128];                      
    let mut tcp_tx_payload = [0; 128];
    let tcp_socket = TcpSocket::new(
        TcpSocketBuffer::new(tcp_rx_payload.as_mut()),                         
        TcpSocketBuffer::new(tcp_tx_payload.as_mut()),
    );
            
    let mut sockets = [None, None];
    let mut socket_set = SocketSet::new(sockets.as_mut());
    let udp_handle = socket_set.add(udp_socket);
    let tcp_handle = socket_set.add(tcp_socket);
```

When I try to build this, I get the following error:

```
error[E0597]: `udp_rx_payload` does not live long enough
   --> src/main.rs:117:56
    |
117 |         UdpSocketBuffer::new(udp_rx_metadata.as_mut(), udp_rx_payload.as_mut()),
    |                                                        ^^^^^^^^^^^^^^ borrowed value does not live long enough
...
167 | }
    | - borrowed value only lives until here
    |
    = note: borrowed value must be valid for the static lifetime...

error[E0597]: `udp_tx_payload` does not live long enough
   --> src/main.rs:118:56
    |
118 |         UdpSocketBuffer::new(udp_tx_metadata.as_mut(), udp_tx_payload.as_mut()),
    |                                                        ^^^^^^^^^^^^^^ borrowed value does not live long enough
...
167 | }
    | - borrowed value only lives until here
    |
    = note: borrowed value must be valid for the static lifetime...
```

If I comment out either of the calls to `socket_set.add`, it builds and runs.

I'm using rustc 1.28.0-nightly (c2d46037f 2018-05-24).

whitequark:
[...]

PR #304, opened 2019-08-14
Allow TCPSockets in a SocketSet without 'static storage lifetime

richardeoin:
`TCPSocket` buffers are type `RingBuffer<'a, _>` whilst other socket types have buffers with type `PacketBuffer<'a, 'b, _>`.

There is an enum over socket types `pub enum Socket<'a, 'b: 'a>`, which is in turn used by `set::Item` and `SocketSet`. The lifetime `'b` is not required for `TcpSocket`, but the `Into<Socket>` implementation for `TCPSocket` specifies the `'static` lifetime for `'b`.

Therefore using `TCPSocket` in a `SocketSet` currently requires the lifetime `'b` for other sockets to be `'static` which is unnecessarily restrictive. This patch fixes this.

The lifetime signature of socket specifies `'b: 'a`, that is `'b` outlives `'a`. Instead of `'static`, this is also satisfied by `impl<'a> Into<Socket<'a, 'a>> for TcpSocket<'a>` as `'a` "outlives" `'a` by definition.

The example [here](https://gist.github.com/richardeoin/b562e79d0d499e00a8bf4944f00594d9) fails to compile without this patch (tested on `rustc 1.38.0-nightly (07e0c3651 2019-07-16)`)

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1646,8 +1646,8 @@ impl<'a> TcpSocket<'a> {
     }
 }
 
-impl<'a> Into<Socket<'a, 'static>> for TcpSocket<'a> {
-    fn into(self) -> Socket<'a, 'static> {
+impl<'a> Into<Socket<'a, 'a>> for TcpSocket<'a> {
+    fn into(self) -> Socket<'a, 'a> {
         Socket::Tcp(self)
     }
 }


crawford:
At a glance, it looks like this might fix https://github.com/m-labs/smoltcp/issues/223.

whitequark:
@m-labs-homu r+

Very nice, thank you.

m-labs-homu:
:pushpin: Commit 3ccb327 has been approved by `whitequark`

m-labs-homu:
:hourglass: Testing commit 3ccb327f4e15871ba96d5287fcedd0be33f8f540 with merge 1ada3da5aa33917daf891dda2242d7f5beadfb10...

m-labs-homu:
:sunny: Test successful - [status-travis](https://travis-ci.org/m-labs/smoltcp/builds/572090999?utm_source=github_status&utm_medium=notification)
Approved by: whitequark
Pushing 1ada3da5aa33917daf891dda2242d7f5beadfb10 to master...

PR #309, opened 2019-10-04
Revert to prevous ordering of TCP options

MabezDev:
This shouldn't change anything for spec compliant routers, infact, the options used to be in the order, [see](https://github.com/m-labs/smoltcp/pull/232/files#diff-02cfbd2822bc829309d766c8998854afR773) unfortunately I've found a router that blocks my packets unless mss is sent first (very annoying bug to find).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,32 +12,35 @@ matrix:
       env: FEATURES='default' MODE='test'
     ### Test select feature permutations, chosen to be as orthogonal as possible
     - rust: nightly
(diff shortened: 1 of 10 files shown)


whitequark:
cc @jhwgh1968

jhwgh1968:
Interesting. I don't _think_ that should cause a problem...

(If you were waiting on me, sorry for delaying your PR. I have updated my GitHub status to reflect my current plans.)

PR #310, opened 2019-10-07
dhcpv4: use offered ip in requested ip option

### LINKED ISSUE #302: DHCP: is setting address before ACKed ok?
Migrating my project to the now merged version of the DHCP client, I noticed that the IP address of the interface has to be set to the OFFERed address before the client will send out a proper REQUEST.  I feel like this is racy, and the "official" IP of the interface should only be set once the server has ACKed the request.

The client is already tracking state information in its `state` enum (which btw I think should be available publicly, or at least as a discriminant-only Discovering/Requesting/Renew), so the last offered IP could also go there?

/cc @astro

MabezDev:
Instead of configuring the ifaces ip address when a dhcp server offers
and ip, the client now use the requested_ip dhcp option, based on the offered ip
(yiaddr). The user now only has to configure the iface once the dhcp
transaction has completed (Ack'd).

Closes #302.

cc: @astro

### DIFF
--- a/src/dhcp/clientv4.rs
+++ b/src/dhcp/clientv4.rs
@@ -34,6 +34,7 @@ struct RequestState {
     retry: u16,
     endpoint_ip: Ipv4Address,
     server_identifier: Ipv4Address,
+    requested_ip: Ipv4Address,
 }
 
 #[derive(Debug)]
@@ -118,8 +119,7 @@ impl Client {
     /// DHCP requests when timeouts are ready.
     ///
     /// Applying the obtained network configuration is left to the
-    /// user. You must configure the new IPv4 address from the
-    /// returned `Config`. Otherwise, DHCP will not work.
+    /// user.
     ///
     /// A Config can be returned from any valid DHCP reply. The client
     /// performs no bookkeeping on configuration or their changes.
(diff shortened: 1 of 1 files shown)


MabezDev:
@whitequark would it be possible to squeeze this in be fore the next release? Not sure who's best to review this as @astro is busy atm I think.

whitequark:
Okay, I'll review it before that.

src/dhcp/clientv4.rs
```diff
@@ -198,9 +198,8 @@ impl Client {
         };
         net_debug!("DHCP recv {:?} from {} ({})", dhcp_repr.message_type, src_ip, server_identifier);
 
-        let config = if (dhcp_repr.message_type == DhcpMessageType::Offer ||
-                         dhcp_repr.message_type == DhcpMessageType::Ack) &&
-                      dhcp_repr.your_ip != Ipv4Address::UNSPECIFIED {
+        // once we recieve the ack, we can pass the config to the user
```
astro: s/recieve/receive/ :)

MabezDev:
@whitequark I see this didn't make it into 0.6 :(, Will this now have to wait till 0.7 or could it make it into a patch release? (I don't *think* this is breaking change)

whitequark:
We can make a patch release just fine. I was just really low on time lately--as soon as I review this, I can cut a release for you.

MabezDev:
Thanks. I appreciate the time you've taken to review my PR's thus far. Let me know if there is anything I can do to help.

astro (approved):

PR #311, opened 2019-10-07
raw socket truncated packets

MabezDev:
Gracefully handle packet truncation errors, instead of panicking. Ran into this when testing smaller MTU sizes.


whitequark:
1. Can you add a test?
2. Can you change the commit message so that the first line isn't just `raw_socket:`?

MabezDev:
Done. Let me know if there is anything else.

PR #313, opened 2019-10-29
Add capactity methods to packet_buffer & sockets

MabezDev:


src/iface/ethernet.rs
```diff
@@ -851,7 +851,7 @@ impl<'b, 'c, 'e> InterfaceInner<'b, 'c, 'e> {
                 // The packet is valid and handled by socket.
                 Ok(()) => handled_by_raw_socket = true,
                 // The socket buffer is full or the packet was truncated
-                Err(Error::Exhausted) | Err(Error::Truncated) => (),
+                Err(Error::Exhausted) | Err(Error::Truncated) => net_debug!("Raw packet dropped"),
```
whitequark: This change isn't related to adding capacity methods.

MabezDev: Apologies, I have removed it.

whitequark:
I feel like the documentation would be more clear if instead of "element" it said which specific kind of element it means (packets and bytes depending on the kind of the socket). Also, please add dots at the end of sentences.

MabezDev:
> packets and bytes depending on the kind of the socket

The current methods only describe the capacities of the payload buffers, which are all bytes. Would you like me to add capacity methods for meta buffers for the socket types that support it (udp, icmp etc) too?

whitequark:
> The current methods only describe the capacities of the payload buffers, which are all bytes. Would you like me to add capacity methods for meta buffers for the socket types that support it (udp, icmp etc) too?

Oh, good point. For those sockets, we should not have a single `capacity` method but two different ones. What about `payload_capacity` and `packet_capacity`?

PR #322, opened 2020-01-13
Change mem::uninitialized to mem::MaybeUninit

ureeves:
Since Rust 1.39.0 [mem::uninitialized()](https://doc.rust-lang.org/core/mem/fn.uninitialized.html) has been deprecated. Removed its usages
and used the recommended [mem::MaybeUninit](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html) instead.


whitequark:
Our MSRV is lower than 1.36.0, so I'm going to keep this PR open until it is at least that high.

ureeves:
Woops. Hadn't noticed you support 1.28.0 and later. In any case, happy to keep it open until appropriate.

whitequark:
Merged in #357.

PR #325, opened 2020-02-19
Transfer smoltcp to the smoltcp-rs organization.

astro:
As per <https://github.com/rust-embedded/wg/blob/master/ops/post-transfer.md>

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @smoltcp-rs/team
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,0 +1,4 @@
+block_labels = ["needs-decision"]
+delete_merged_branches = true
+required_approvals = 1
+status = ["continuous-integration/travis-ci/push"]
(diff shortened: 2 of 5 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+smoltcp-rs astro
```
therealprof: Shouldn't that be:
```suggestion
* @smoltcp-rs
```
instead?

astro: Doesn't it have to be in `@org/team` syntax?

therealprof: Not sure, but as is it's definitely wrong. ;)

rvolgers:
You may also want to update the `homepage` and `repository` in `Cargo.toml`.

therealprof:
@astro Would you mind rebasing this?

astro:
@therealprof Done

Are there any news regarding whether we are going to continue with the smoltcp-rs Github organization?

therealprof:
Idea is to bring up a community/libs team and have them handle important but non-core crates. No progress on that though.

therealprof (approved):

Dirbaio:
Closing, as this is stale by now. Smoltcp is staying in the smoltcp-rs org.

PR #329, opened 2020-03-17
use provided ip for TcpSocket::connect instead of 0.0.0.0

trinity-1686a:
the comment seems to say what was to do, but the code wasn't doing that, it was unconditionally returning `remote_endpoint.addr.to_unspecified()` as the first branch was unreachable, due to src/socket/tcp.rs:480


whitequark (approved):
Looks good, thanks!

PR #335, opened 2020-04-16
Fix support table rendering in toplevel doc

nabijaczleweli:
As [seen](https://docs.rs/smoltcp/0.6.0/smoltcp/#packet-and-representation-layer-support) on docs.rs:
before SS

With patch:
after SS


whitequark (approved):

whitequark:
Thanks!

Issue #334, opened 2020-04-16
Support for layer 3 mediums (IP only, non-Ethernet)


Dirbaio:
It would be useful to have support for non-ethernet interfaces, that send/receive raw IP packets without ethernet headers and don't do ARP/NDISC.

There are several use cases for this:

- PPP connections in GSM/GPRS/2G/3G/LTE/whatever mobile modules (simcom, quectel, ublox..). These are super popular for connecting embedded devices to a mobile network. To access the internet with these you typically use PPP over a serial link, and PPP carries IP packets. This is the usecase I'm personally interested in.
- VPNs such as Wireguard typically carry IP packets. Adding support would be a TCP socket in a IP-only interface whose IP packets get encrypted and wrapped in UDP packets that are then sent through the real Ethernet interface.
- We could use tun interfaces instead of tap interfaces (though this is not very useful on its own). #82 

I'm willing to take a look at implementing it. Is this something you would be interested in? 

If so, how do you think it'd be the best way to implement it?

From a quick look at the code IMO the cleanest way would be to separate all the IP stuff into an "IP interface" struct, and have the Ethernet interface embed it. Socket handling, IP handling would go in the IP interface, ARP, NDISC and Ethernet header stuff would go in the Ethernet interface. I'm not sure how to handle some details though, such as NDISC handling which is deep in the ipv6 handling code.

whitequark:
> Is this something you would be interested in?

This is something that is definitely useful, see #326 for example.

> If so, how do you think it'd be the best way to implement it?

I'm not entirely sure, either. I suspect the most practical way to do this is to have a completely separate IPInterface, because there are many concepts that won't apply at all to a pure-IP interface, like ARP, NDISC, broadcast/multicast, and so on.

Personally, I would be fine with a significant amount of code duplication between `IPInterface` and `EthernetInterface` because I strongly suspect that embedding the former into the latter wouldn't really work (or would be so complex as to outweigh the benefits of sharing code between them), but if you experiment with it and it turns out I'm wrong, we can do that too.

lattice0:
I was about to open an issue on this. I'm working on making a OpenVPN3 library. That is, I want to send/receive through OpenVPN as a library, not using the OS's tun/tap devices. Since OpenVPN3 client has only support for IP packet transmission (no ethernet packets transmission) my idea was to use this library to process the IP packets, without even touching ethernet. But it looks like it's not possible?

[...]

PR #336, opened 2020-04-17
Add support for IP mediums, v1

Dirbaio:
Opening this PR so I can get feedback earlier in case there are problems with the approach I'm taking.

This is a first draft of #334. The goal is to split EthernetInterface into Ethernet and IP parts, so the IP parts can be reused in other non-ethernet interface types (linux tun, VPNs, PPP).

The IP module defines the following structs:
- `Config`: contains IP configuration (addresses, routes) that doesn't change when processing IP packets.
- `State`: contains state that can change when processing IP packets.
- `Processor`: contains a `&Config` and a `&mut State`. Most of the functions moved from ethernet are here. This is just a convenience wrapper, as most functions need the config and the state, so they don't have to be passed around as parameters all the time.

The motivation for separating `Config` and `State` is that all the ethernet code needs read-only access to `Config`. This way `Config` can be immutably borrowed many times and shared everywhere it's needed, at the same time `State` is mutably borrowed to process packets. Without this I was getting borrow issues with both the ethernet and ip code needing to access the ip config.

For packet egress, the Ethernet layer passes a struct implementing a `Dispatcher` traint to the IP layer, which is used to pass IP packets back to the Ethernet layer.

Missing items:
- Tests. I'll update them as soon as we've confirmed the code structure is OK
- IPv6 NDISC. I don't know what to do, I have 2 ideas:
   - Have `ethernet::Interface` inject a Socket that handles NDISC. The socket would have to mut borrow ethernet::InnerInterface so it would have to be created on the fly every time, not sure if it's even possible.
   - Pass a `process_ndisc` callback to `process_ipv6`... ugly but should work?
- Some random FIXMEs in the code.

My questions:

1. any general feedback on the taken approach, or possible improvements?
2. Most of the functions in the Ethernet code take as param the `ip::Config`, passing it around is somewhat annoying. This could benefit from an abstraction similar to `ip::Processor`: `ethernet::InterfaceInner` would become `ethernet::State`, and there would be an `ethernet::Processor` with ip_config, ethernet_state. Pros: no passing around `ip` all the time, design is consistent with IP. Cons: it's an extra struct. What do you think?
3. Should the IP structs be `pub` or `pub(crate)`? I guess the latter, unless we want people to implement their interfaces out-of-tree?
4. Is the route table an ethernet concept or an IP concept? To me it looks like it is an Ethernet concept. The only use in the IP code is for AnyIP. To me it feels a bit strange that AnyIP reads the route table instead of just using the CIDRs in `ip_addrs`. (Also AFAICT these CIDRs aren't used for anything?). If we change it to use just `ip_addrs`, then `routes` could be moved to ethernet.

### DIFF
--- a/src/iface/ip.rs
+++ b/src/iface/ip.rs
@@ -1,0 +1,106 @@
+#[cfg(not(feature = "proto-igmp"))]
+use core::marker::PhantomData;
+
+use phy::{DeviceCapabilities};
+#[cfg(feature = "proto-igmp")]
+use wire::{IpAddress, IpRepr};
+#[cfg(feature = "proto-ipv6")]
+use wire::{Ipv6Repr};
+#[cfg(feature = "proto-ipv4")]
+use wire::{Ipv4Repr};
+#[cfg(feature = "proto-ipv4")]
+use wire::{Icmpv4Packet, Icmpv4Repr};
+#[cfg(feature = "proto-igmp")]
+use wire::{IgmpPacket, IgmpRepr};
+#[cfg(feature = "proto-ipv6")]
+use wire::{Icmpv6Packet, Icmpv6Repr};
+#[cfg(feature = "socket-udp")]
+use wire::{UdpPacket, UdpRepr};
+#[cfg(feature = "socket-tcp")]
+use wire::{TcpPacket, TcpRepr};
+
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum Packet<'a> {
+    #[cfg(feature = "proto-ipv4")]
+    Icmpv4((Ipv4Repr, Icmpv4Repr<'a>)),
+    #[cfg(feature = "proto-igmp")]
+    Igmp((Ipv4Repr, IgmpRepr)),
+    #[cfg(feature = "proto-ipv6")]
+    Icmpv6((Ipv6Repr, Icmpv6Repr<'a>)),
+    #[cfg(feature = "socket-raw")]
+    Raw((IpRepr, &'a [u8])),
+    #[cfg(feature = "socket-udp")]
+    Udp((IpRepr, UdpRepr<'a>)),
+    #[cfg(feature = "socket-tcp")]
+    Tcp((IpRepr, TcpRepr<'a>))
+}
+
+impl<'a> Packet<'a> {
+    pub(crate) fn neighbor_addr(&self) -> IpAddress {
+        return self.ip_repr().dst_addr()
+    }
+
+    pub(crate) fn ip_repr(&self) -> IpRepr {
+        match &self {
+            #[cfg(feature = "proto-ipv4")]
+            &Packet::Icmpv4((ipv4_repr, _)) => IpRepr::Ipv4(ipv4_repr.clone()),
+            #[cfg(feature = "proto-igmp")]
+            &Packet::Igmp((ipv4_repr, _)) => IpRepr::Ipv4(ipv4_repr.clone()),
+            #[cfg(feature = "proto-ipv6")]
+            &Packet::Icmpv6((ipv6_repr, _)) => IpRepr::Ipv6(ipv6_repr.clone()),
+            #[cfg(feature = "socket-raw")]
+            &Packet::Raw((ip_repr, _)) => ip_repr.clone(),
+            #[cfg(feature = "socket-udp")]
+            &Packet::Udp((ip_repr, _)) => ip_repr.clone(),
+            #[cfg(feature = "socket-tcp")]
+            &Packet::Tcp((ip_repr, _)) => ip_repr.clone(),
+        }
+    }
+
+    pub(crate) fn emit_payload(&self, _ip_repr: IpRepr, payload: &mut [u8], caps: &DeviceCapabilities) {
+        match self {
+            #[cfg(feature = "proto-ipv4")]
+            Packet::Icmpv4((_, icmpv4_repr)) => 
+                icmpv4_repr.emit(&mut Icmpv4Packet::new_unchecked(payload), &caps.checksum),
+            #[cfg(feature = "proto-igmp")]
+            Packet::Igmp((_, igmp_repr)) =>
+                igmp_repr.emit(&mut IgmpPacket::new_unchecked(payload)),
+            #[cfg(feature = "proto-ipv6")]
+            Packet::Icmpv6((_, icmpv6_repr)) =>
+                icmpv6_repr.emit(&_ip_repr.src_addr(), &_ip_repr.dst_addr(),
+                         &mut Icmpv6Packet::new_unchecked(payload), &caps.checksum),
+            #[cfg(feature = "socket-raw")]
+            Packet::Raw((_, raw_packet)) =>
+                payload.copy_from_slice(raw_packet),
+            #[cfg(feature = "socket-udp")]
+            Packet::Udp((_, udp_repr)) =>
+                udp_repr.emit(&mut UdpPacket::new_unchecked(payload),
+                              &_ip_repr.src_addr(), &_ip_repr.dst_addr(), &caps.checksum),
+            #[cfg(feature = "socket-tcp")]
+            Packet::Tcp((_, mut tcp_repr)) => {
+                // This is a terrible hack to make TCP performance more acceptable on systems
+                // where the TCP buffers are significantly larger than network buffers,
+                // e.g. a 64 kB TCP receive buffer (and so, when empty, a 64k window)
+                // together with four 1500 B Ethernet receive buffers. If left untreated,
+                // this would result in our peer pushing our window and sever packet loss.
+                //
+                // I'm really not happy about this "solution" but I don't know what else to do.
+                if let Some(max_burst_size) = caps.max_burst_size {
+                    let mut max_segment_size = caps.max_transmission_unit;
+                    max_segment_size -= _ip_repr.buffer_len();
+                    max_segment_size -= tcp_repr.header_len();
+
+                    let max_window_size = max_burst_size * max_segment_size;
+                    if tcp_repr.window_len as usize > max_window_size {
+                        tcp_repr.window_len = max_window_size as u16;
+                    }
+                }
+
+                tcp_repr.emit(&mut TcpPacket::new_unchecked(payload),
+                                &_ip_repr.src_addr(), &_ip_repr.dst_addr(),
+                                &caps.checksum);
+            }
+        }
+    }
+}
--- a/src/iface/ip.rs
+++ b/src/iface/ip.rs
@@ -1,24 +1,127 @@
+use core::cmp;
+use managed::ManagedSlice;
+#[cfg(feature = "proto-igmp")]
+use managed::ManagedMap;
 #[cfg(not(feature = "proto-igmp"))]
 use core::marker::PhantomData;
 
+use {Error, Result};
 use phy::{DeviceCapabilities};
 #[cfg(feature = "proto-igmp")]
-use wire::{IpAddress, IpRepr};
+use time::Duration;
+use time::Instant;
+use wire::{IpAddress, IpProtocol, IpRepr, IpCidr};
 #[cfg(feature = "proto-ipv6")]
-use wire::{Ipv6Repr};
+use wire::{Ipv6Packet, Ipv6Repr, IPV6_MIN_MTU};
 #[cfg(feature = "proto-ipv4")]
-use wire::{Ipv4Repr};
+use wire::{Ipv4Address, Ipv4Packet, Ipv4Repr, IPV4_MIN_MTU};
 #[cfg(feature = "proto-ipv4")]
-use wire::{Icmpv4Packet, Icmpv4Repr};
+use wire::{Icmpv4Packet, Icmpv4Repr, Icmpv4DstUnreachable};
 #[cfg(feature = "proto-igmp")]
-use wire::{IgmpPacket, IgmpRepr};
+use wire::{IgmpPacket, IgmpRepr, IgmpVersion};
 #[cfg(feature = "proto-ipv6")]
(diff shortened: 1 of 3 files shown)


therealprof:
I had a quick glance and I do like the idea.

> Is the route table an ethernet concept or an IP concept?

Routing is layer 3 together with IP, Ethernet is layer 2 in the ISO/OSI model.

whitequark:
The ISO/OSI model isn't terribly relevant to real-world TCP/IP stacks; TCP is not separate from IP, and IP is not easily separate from Ethernet. Rather than considering this in abstract terms, it is best to see what effect any particular kind of separation has on your network stack, and go from that.

@Dirbaio I think your idea about separating `Config` and `State` is quite interesting. Do you think we could apply it more incrementally, i.e. first splitting the Ethernet structs into two, and only then splitting `EthernetInterface`?

therealprof:
> IP is not easily separate from Ethernet.

I disagree. There's absolutely no relationship between Ethernet and IP whatsoever. You can run a lot of different protocols on top of Ethernet and there're also some commonly used IP implementations which do not run on top of Ethernet, e.g. PPP.

In this particular case the question was about routing and since routing can only work on top of a protocol like IP, there's no question for me that this does not belong in Ethernet at all which is why I brought up the ISO/OSI layers.

> Rather than considering this in abstract terms, it is best to see what effect any particular kind of separation has on your network stack, and go from that.

I agree in priciple but in this particular case it seems pretty clear that routing does not belong into Ethernet.

whitequark:
> You can run a lot of different protocols on top of Ethernet

Of course, I never implied that you can't.

> there're also some commonly used IP implementations which do not run on top of Ethernet, e.g. PPP

Of course also true.

> There's absolutely no relationship between Ethernet and IP whatsoever.

And this is just black-and-white thinking that obscures the real situation. If you said that IP is not strictly coupled to Ethernet in the way that TCP is coupled to IP, I would agree. However, a practical IPv4 stack, for example, cannot have a clean separation beteween the LLC sublayer and the IP layer because the concept of default gateways violates it. There is a similar issue with IPv6 NDISC. We also have DHCP, a layer 3 configuration protocol that runs on layer 4.

That is what I mean by "not easily separate": not in the sense that it is impossible to decouple IP from Ethernet (of course it is), but in the sense that real world IP implementations are a lot more messy than the model might lead you to think, and have interdepenencies between the Ethernet and IP parts, as this PR very clearly demonstrates. 

The reason I bring it up is because one of the core principles of smoltcp, one that I decided on before I even wrote any code for the project, is to implement the protocols as they exist in real world and serve real users rather than adhere to a rigid model that doesn't fully apply to our system. That is something I would like to be preserved.

<hr>

> I agree in priciple but in this particular case it seems pretty clear that routing does not belong into Ethernet.

We are in agreement on this issue. I mostly wanted to provide historical context and explain the principles I used when designing smoltcp.

Dirbaio:
I've pushed a new version of the patches following @whitequark's suggestion. First commit splits ethernet into config+state, last commit splits these into a config+state for IP and another for Ethernet. 

I find this way makes Ethernet and IP code more consistent, and hopefully makes review easier. Also, there's no more passing around ip::Config since the Ethernet processor now has access to it.

whitequark:
@Dirbaio I've reviewed them and I do very much like the first two commits! (Some tiny style nits, that's it.)

Regarding the third, do you think you can summarize the change in a few sentences or pargraphs? It's a massive diff and that would make it a lot easier for me to review.

Dirbaio:
I've pushed a new set of commits, splitting the last commit in two. First one splits Packet, second one splits Config, State, Processor. I'm afraid I can't think of more ways to split the last one.

Also, commit messages now have more info, summarizing the changes.

Dirbaio:
Pushed new commits.
- Implemented ndisc. (I had to add two traits, LowerDispatcher and LowerProcessor. I tried doing it with just one, but it runs into borrow issues)
- Updated tests to the new structure (they all pass without needing logic changes).

With these, this is ready for review now.

Dirbaio:
Tests are failing due to an unrelated breakage from latest nightly. I've sent a separate PR for that:  #337

whitequark:
Thanks for splitting the changes. I like your general approach.

You've added a bunch of new `pub` (not `pub(crate)`) functions; is that intended?

Could you go through the added documentation and fix up the formatting and capitalization to be consistent with existing doc strings? This isn't terribly important but makes it nicer to read.

lattice0:
I'm trying to make it work by comparing with the ethernet implementation. I'd like to create an HTTP client example. As of now, `ip.rs` has no polling function `poll`. Shouldn't it have? So an example client would be almost like the one we already have on examples/client.rs. Is there a reason on why you didn't implement it, @Dirbaio?

Dirbaio:
`IP` is supposed to be used only internally from the actual interface structs, it's not supposed to be an interface itself. For example if you want to use a `tun` linux device you'd create a `TunInterface` that uses `IP` internally. `poll` & co methods would be duplicated, maybe there's a way around that

I'm planning to send a follow-up PR adding it.

lattice0:
@Dirbaio since I want to use this stack on OpenVPN, instead of creating a `TunInterface` impementing the `Device` trait that taks with a linux file descriptor, I'm thinking about creating a `VirtualTunInterface` that reads and writes packets from memory, so I can integrate into OpenVPN. The usage of a `VirtualTunInterface` would be the same as the usage of `TunInterface`, so I created a `tun.rs` file that mimics `ethernet.rs`. And then I created `virtual_tun.rs` that implements a virtual TUN device. Maybe you can implement the `TunInterface` later. 

I believe this is what should be done. Can you take a look at my sketch? https://github.com/Dirbaio/smoltcp/compare/ip-interface...lucaszanella:ip-interface?expand=1

I'm willing to do the `tun.rs` for you, since the only difference would be the `VirtualTunInterface` that implements the `Device` trait which is minimal. But I wanted to know if that's what you have in mind.

src/iface/ip.rs
```diff
@@ -0,0 +1,1030 @@
...
+                    },
+                    _ => Err(Error::Unrecognized),
+                }
+            }
+
+            // Ignore any echo replies.
+            Icmpv6Repr::EchoReply { .. } => Ok(None),
+
+            // Forward any NDISC packets to the ndisc packet handler
+            #[cfg(feature = "ethernet")]
```
dlrobertson, line 769: Is this really only needed for `ethernet`?

lattice0:
Hey @Dirbaio, if you look at https://github.com/Dirbaio/smoltcp/compare/ip-interface...lucaszanella:ip-interface?expand=1, you'll see several modifications. I sucessfully made a VirtualTun device that receives packets from a queue instead of a file descriptor, and sends to a queue (queue part to be done but the device receives the packet already). I'm working with example packets to debug things since it's going to take time for me to integrate this to OpenVPN.

So what you said you would do is kinda done, I just need to refactor + reactivate the tests that I commented.

Now you'd only have to create a real Tun device instead of using my VirtualTun one.

What you think? Can you make it so we can test my modifications better?

again: code is ugly, gonna refactor a lot

src/iface/ip.rs
lattice0: I also wanna know this. Had problems with this function on my code changes

lattice0:
Ok so I made the tun interface also. So now the project has a tap interface, a tun interface, and a virtual tun interface (reads packets from memory instead of from file descriptor).

I used Wireshark and could see the packet being sent to the tun device I created. However, I don't know how to connect this device to the world, so I can sucessfully test the HTTP client example. Right now I can see that it sends a TCP SYN but since the tun device is not connected to anything, it wont respond.

Any ideas on how to test it?

My changes: https://github.com/Dirbaio/smoltcp/compare/ip-interface...lucaszanella:ip-interface?expand=1

only thing missing now are mod tests that I commented + refactoring

lattice0:
I made the `tunhttpclient` example work!

```
sudo ip tuntap add dev tun1 mode tun user `id -un`
sudo ip link set dev tun1 up
sudo ip addr add dev tun1 local 192.168.69.0 remote 192.168.69.1
sudo iptables -t filter -I FORWARD -i tun1 -o eth0 -j ACCEPT
sudo iptables -t filter -I FORWARD -m state --state ESTABLISHED,RELATED -j ACCEPT
sudo iptables -t nat -I POSTROUTING -o eth0 -j MASQUERADE
sudo sysctl net.ipv4.ip_forward=1
```

`./tunhttpclient 172.217.28.238 http://google.com`

it sucessfully retrivies HTML from google using the tun interface :)

lattice0:
Great news!

tun example working and all non ethernet tests working!

https://github.com/Dirbaio/smoltcp/compare/ip-interface...lucaszanella:ip-interface?expand=1

Just need to know if `process_ndisc` is ethernet related or not. Anyone? 

After this, I can refactor the code to take comments out and then pull request to @Dirbaio's code

whitequark:
> Just need to know if `process_ndisc` is ethernet related or not. Anyone?

NDISC is the equivalent of ARP, AFAIK you don't need it for TUN interfaces (but I didn't check in depth).

lattice0:
thanks @whitequark, however it was not simple to remove ndisc. see:

I refactored everything. There are just 2 issues before I push to @Dirbaio's code:

Here:

https://github.com/lucaszanella/smoltcp/blob/071e416320cb2303e1c2b886550fc4169b3559f2/src/iface/tun.rs#L476

if I pass a `lower` that does not implement `ip::LowerProcessor` it won't compile because of the trait bound `ip::LowerProcessor` for the `lower` argument. This is the trait that requires `process_ndisc` so I left like this:

```
    fn process_ndisc<'frame>(&mut self, timestamp: Instant, ip_repr: Ipv6Repr,
                             repr: NdiscRepr<'frame>) -> Result<Option<ip::Packet<'frame>>> {
        self._processor.process_ndisc(timestamp, ip_repr, repr)
    }
```
where `process_ndisc` does nothing:

```
    fn process_ndisc<'frame>(&mut self, _timestamp: Instant, _ip_repr: Ipv6Repr,
        _repr: NdiscRepr<'frame>) -> Result<Option<ip::Packet<'frame>>> {
            Ok(None)
    }
```

and here:

https://github.com/lucaszanella/smoltcp/blob/071e416320cb2303e1c2b886550fc4169b3559f2/src/iface/tun.rs#L429

there's no `Packet` in `wire/ip.rs` just as there's `EthernetFrame` in `wire/ethernet.rs`

The problem is that I can't implement

`impl<T: AsRef<[u8]>> PrettyPrint for Frame<T> {`

in order to pretty print the packet in that line.

What should be done here? I see that there's methods to pretty print each individual packet (ipv6, igmp, udp, tcp etc) but there's no trait implementation for IP.

lattice0:
Hello, finally a pull requested to @Dirbaio's repo :)

https://github.com/Dirbaio/smoltcp/pull/1/files

Dirbaio:
@lucaszanella that looks great! Thanks for putting the work.

Sorry everyonw for the long delays adressing questions and review.

I've been integrating smoltcp with my changes into our product's firmware, and I've found some concerns about the current implementation.

## Concern 1: binary size bloat

Right now most of the `ip` code is generic over the `LowerProcessor` trait, which means they're emitted N times as binary code, monomorphized for each type of interface in use. This monomorphization propagates to all the sockets' `dispatch` methods because they take a closure that uses the `LowerProcessor`. And all these methods are definitely not small :( 

This means that if you want your firmware to support `ethernet` and `ip` devices at the same time you're going to pay quite a bit of flash space penalty. This is our case: we want to make a single device, where a single firmware supports pluggable Ethernet/Wifi modules (Ethernet medium), and cellular modules (PPP, IP medium).

You may think this is already a problem with the current code due to the `Device` trait: if you have two `Device` implementations, you're going to have two `iface::Ethernet` monomorphizations. That's not actually the case: you can use dynamic dispatch to the actual device code, so all the smoltcp code is only instantiated once. It's a bit [convoluted](https://gist.github.com/Dirbaio/7421499c5be944bf5e4e52689bef17d4) because `Device` is not object-safe, but it's doable.

## Concern 2: maintainability, code duplication

As seen in the current `EthernetInterface` and @lucaszanella 's `TunInterface`, there's quite a bit of code duplication that I don't think is easy to remove.

The tests should be refactored so all IP-specific tests are in `ip` module, but there's still a lot of duplication left. I really can't think of a way to remove it :( 

## Concern 3: the Device trait is ambiguous

You can create a `iface::TunInterface` with a `phy::TapInterface`, it compiles with zero complaints but blows up spectacularly at runtime.

## Alternative solution

I think I've found a way to solve these problems:

- Add a `medium` method to `Device`, that returns `Medium::Ethernet` or `Medium::IP`.
- Rename `iface::Ethernet` to `iface::Interface` (or similar. The point is that there won't be an interface struct per medium type)
- Add checks to `iface::Interface` that check the medium of the device, and behave differently accordingly. For example, if medium is IP, send IP payloads to the device. If it's Ethernet, do ARP/NDISC and send Ethernet payloads to the device.

This solves the binary bloat nicely:
- If you instantiate `Interface` with an Ethernet-only device, the `medium` method will return a constant of `Medium::Ethernet`. This will allow the compiler to inline it and statically remove all the checks, so the resulting code should be about as performant as the current one. Same with an IP-only device.
- You can instantiate `Interface` with a device impl that does dynamic dispatch. This way you get only a single instantiation of all the smoltcp code, at the cost of it doing checks at runtime to see if it's dealing with an Ethernet or IP medium.

It also solves code duplication since there'll be just only one Interface struct.

I'm going to write a prototype of it. This will allow us to do code-size comparisons and decide.

whitequark:
I'm excited to see your solution. In general I think code duplication isn't necessarily bad but I agree that in this specific case it probably makes things harder to maintain!

Dirbaio:
Closing this in favor of #346

PR #341, opened 2020-05-15
Specify concrete type for Bound::Unbounded

### LINKED ISSUE #340: E0277 - the trait bound `(core::ops::Bound<_>, core::ops::Bound<()>): core::ops::RangeBounds<wire::ip::Cidr>` is not satisfied
Hi,

Looks like 0.6 does not compile on rustc 1.45 :

```
error[E0277]: the trait bound `(core::ops::Bound<_>, core::ops::Bound<()>): core::ops::RangeBounds<wire::ip::Cidr>` is not satisfied
   --> /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/smoltcp-0.6.0/src/iface/route.rs:120:45
    |
120 |         for (prefix, route) in self.storage.range((Bound::Unbounded, Bound::Included(cidr))).rev() {
    |                                             ^^^^^ the trait `core::ops::RangeBounds<wire::ip::Cidr>` is not implemented for `(core::ops::Bound<_>, core::ops::Bound<()>)`
    |
    = help: the following implementations were found:
              <(core::ops::Bound<&'a T>, core::ops::Bound<&'a T>) as core::ops::RangeBounds<T>>
              <(core::ops::Bound<T>, core::ops::Bound<T>) as core::ops::RangeBounds<T>>

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: could not compile `smoltcp`.
```

Looks like it's doing the same thing on 1.43, am I doing something wrong or is there a supported version of rust I should be using ?

Thank you

zhaofengli:
Somehow type inference got confused. Fixes #340.

### DIFF
--- a/src/iface/route.rs
+++ b/src/iface/route.rs
@@ -117,7 +117,7 @@ impl<'a> Routes<'a> {
             _ => unimplemented!()
         };
 
-        for (prefix, route) in self.storage.range((Bound::Unbounded, Bound::Included(cidr))).rev() {
+        for (prefix, route) in self.storage.range((Bound::Unbounded::<IpCidr>, Bound::Included(cidr))).rev() {
             // TODO: do something with route.preferred_until
             if let Some(expires_at) = route.expires_at {
                 if timestamp > expires_at {


whitequark:
Is this a rustc bug? Should it be reported upstream?

zhaofengli:
Quite possibly, as the code compiled fine before. I would expect the compiler to correctly deduce the `T` in `Unbounded<T>`, as the only matching impl is `core::ops::Bound<T>, core::ops::Bound<T>` as far as I know.

whitequark:
I have no problem approving and merging this, but please do report it upstream, since it seems like potentially a serious issue. I know type inference changes are not generally considered a part of the stability guarantee, but this one seems especially heinous.

whitequark (approved):

jonas-schievink:
This is odd, managed 0.7.0 and 0.7.1 do not build on rustc <=1.27.0, and smoltcp 0.6.0 does not build on rustc >=1.28.0, so there is no Rust version that successfully compiles it according to the specified dependency requirements.

zhaofengli:
Ah sorry, was stuck with a paper deadline a week ago and forgot about this. I'll try to construct a minimal example that can trigger the problem. So far in very trivial cases, rustc seems to be able to correctly determine that my range tuple implements `RangeBounds`.

PR #342, opened 2020-05-17
SNTP wire protocol and client

plorefice:
Hi, this is an attempt at implementing a Simple Network Time Protocol (SNTP) over UDP sockets.

The client is pretty barebones: only unicast mode with a single server is supported, more advanced stuff like root delay and dispersion are skipped, and no adjustments are performed on the received timestamp.

Even so, I've tested it and it works pretty well, with an accuracy at least in the sub-second range. For timekeeping in embedded applications, which I plan to use it on, should be plenty good. However, let me know if more features need to be added to make it worth merging.

I've also taken the liberty to create an additional `apps` module for application-layer protocols, in case anyone decides to implement something else like TFTP or similar, to keep the crate more tidy. I'm not married to the idea though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,8 @@ matrix:
       env: FEATURES='std ethernet proto-ipv4 socket-icmp socket-tcp' MODE='test'
     - rust: nightly
       env: FEATURES='std ethernet proto-ipv6 socket-icmp socket-tcp' MODE='test'
+    - rust: nightly
+      env: FEATURES='std ethernet proto-ipv4 app-sntp' MODE='test'
     ### Test select feature permutations, chosen to be as aggressive as possible
     - rust: nightly
       env: FEATURES='ethernet proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp std'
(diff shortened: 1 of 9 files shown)


whitequark:
Should SNTP be a part of smoltcp? I feel like it would make more sense to publish application-layer stuff as separate crates, so that they wouldn't have to evolve in lockstep with the base stack, nor would they be tied to its release schedule.

plorefice:
The way I see it, if this were published as a separate crate, it would still be strongly tied to this particular stack, increasing the risk of it becoming abandonware or at least obsolete.

By including it here, one is "forced" to maintain it, in some way. I also don't see the high-level API that application layer protocols use changing very often, so I don't think it would be much of a burden.

Finally, I agree with you that most application protocols are too complex to live here, but there are a number of "building blocks" that are nice to have out of the box.

This being said, I have no issue whatsoever in moving this code to a dedicated crate, it would be very easy to do, as it doesn't rely on many (if any) internals. I leave the final decision to you.

whitequark:
> By including it here, one is "forced" to maintain it, in some way.

I think I would not like to be "forced" to maintain code by external contributors.

plorefice:
I see your point. In that case, I'll close this PR and move the code to an external crate.

Thanks for your time!

PR #351, opened 2020-06-11
Return `Error::Finished` in `recv()` on graceful close.

### LINKED ISSUE #349: How to distinguish FIN and RST closes in recv?
When reading data from a `TcpSocket` with the `recv_*` functions, `Error::Illegal` is returned when there's no more readable data. 

The issue is AFAICT there's no way to distinguish whether the remote end gracefully closed the connection with a FIN packet, or the connection was aborted with a RST packet, or other error cases (idle timeout, etc).

A common use case for this is when downloading a file. You want to know whether you have received the whole file (gracefully closed with FIN) or only part of it because the connection died for some other reason, so you don't silently truncate the file.

- Is it correct that this is currently impossible?
- How would the API for this look? Something like `recv_*` functions being able to signal `EOF` vs an error condition, like the Berkeley sockets API.

Dirbaio:
This PR depends on #350.

This allows applications to distinguish whether the remote end has
gracefully closed the connection with a FIN (in which case it has
received all the data intact), or the connection has failed due to e.g. a
RST or a timeout (in which case the received data may be truncated).

Fixes #349

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -134,6 +134,11 @@ pub enum Error {
     /// or a TCP connection attempt was made to an unspecified endpoint.
     Unaddressable,
 
+    /// The operation is finished.
+    /// E.g. when reading from a TCP socket, there's no more data to read because the remote
+    /// has closed the connection.
+    Finished,
+
     /// An incoming packet could not be parsed because some of its fields were out of bounds
     /// of the received data.
     Truncated,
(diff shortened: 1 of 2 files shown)


whitequark (dismissed):
LGTM

Dirbaio:
Added documentation and rebased.

whitequark (approved):

PR #356, opened 2020-06-15
added value for handle and fixed TcpSocket lifetime bug

lattice0:
I remember doing this pull request before but I didnt hear from you. Possibly I didn't press "create"?

as I talked to you, I changed the impl and also added a little value method for Handle because I couldnt get the value out on my code.

### DIFF
--- a/src/socket/set.rs
+++ b/src/socket/set.rs
@@ -19,6 +19,12 @@ pub struct Item<'a, 'b: 'a> {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
 pub struct Handle(usize);
 
+impl Handle {
+    pub fn value(self) -> usize {
+        self.0
+    }
+}
+
 impl fmt::Display for Handle {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "#{}", self.0)
--- a/src/socket/set.rs
+++ b/src/socket/set.rs
@@ -19,12 +19,6 @@ pub struct Item<'a, 'b: 'a> {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
 pub struct Handle(usize);
 
-impl Handle {
-    pub fn value(self) -> usize {
-        self.0
-    }
-}
-
 impl fmt::Display for Handle {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "#{}", self.0)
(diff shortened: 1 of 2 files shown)


whitequark:
1. Please don't do two unrelated changes in one commit.
2. The inner structure of the handle is an implementation detail and is intentionally not exposed.

lattice0:
ok changed it back to only the bug fix.

Getting the value from handle would be really useful though, because without it, to have sockets controled from C++ I'd have to store each Handle in Rust side and give to C++ the number for the handle object.

whitequark:
You can't create a handle even with your change, can you?

lattice0:
Sorry didn't understand. 

I want to create sockets in Rust by calling from C++. I cannot pass the handle to C++ so I wanted to pass the value of the handle. Then C++ could invoke methods on the socket because it knows its handle.

whitequark:
How would you call a method on the socket?

lattice0:
If C++ knows the handle number, it can call the C-callable Rust function with the handle number as an argument:

```
pub extern "C" fn socket_is_active(size_t handle) -> bool {
    let handle = Handle(handle);
    let mut socket = sockets.get::<TcpSocket>(handle);
    sockt.is_active()
}
```

whitequark:
But you can't use the constructor of `Handle` like that, see [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=19fd89154f0ea79a75c9ec4e5a3ec50d). If we were to expose the internals of `Handle` it would make a lot more sense to make the inner field `pub`. I'm not yet sure if we should do that though.

lattice0:
Hmmm right didn't antecipate that because I didnt even start this part of the C interface. I've just wrote the part that returns the socket handle to C++. 

Of course I can deal with it but since Rust does not have static variables I cannot simpy put these `Handle`s in a static list on Rust and give its index back to C++. There are workarounds like lazy_static but not the right way. I could also place it in a Box and return the Box to C++ (on C++ I receive it simply as a `void*` and pass this `void*` back to Rust and receive on Rust as `&mut HandleStruct`). 

Anyways, as you see, having access to the inner number of Handle (and having a way to create a Handle) would facilitate a LOT!

whitequark:
> I could also place it in a Box and return the Box to C++ (on C++ I receive it simply as a `void*` and pass this `void*` back to Rust and receive on Rust as `&mut HandleStruct`).

I would very much prefer you to go this route, since it leaves smoltcp free to adjust the internal structure of the handles however it needs to.

whitequark (approved):

whitequark:
Thanks for your contribution!

PR #360, opened 2020-06-25
Do not send window updates in states that shouldn't do so.

### LINKED ISSUE #358: Don't send window updates if remote side has closed the connection?
I've noticed that smoltcp is sending window updates after receiving a FIN. For example:

- Remote endpoint sends us a bunch of data.
- Remove endpoint sends FIN
- We ACK the data and the FIN, with reduced window size.
- Application reads the data from the TcpSocket.
- We send a window update ACK, with the now larger window.

It looks to me that this last ACK might be avoidable? If we've received a FIN, this means the remote side is not going to send more data, so they shouldn't care about the window size anymore.

I'm not sure whether the RFCs allow this though. I'm opening this issue to discuss. If we decide to do this, I can send a PR.

Dirbaio:
This fixes #358, but not only. I've noticed that window updates are also sent in CloseWait and Closed states, which is actually incorrect.

This PR adds two previously-failing tests.

Previously, the window update was sent anytime it changed. This is
not desirable for 2 reasons:

- It would send window updates in TimeWait or Closed states, which is incorrect.
- It would send window updates in states where we've already received the
  remote side's FIN, such as CloseWait. This is not necessarily incorrect, but
  is useless, since the remote side is not going to send us more data, therefore
  it's not interested in our window size anymore.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1417,8 +1417,11 @@ impl<'a> TcpSocket<'a> {
     }
 
     fn window_to_update(&self) -> bool {
-        (self.rx_buffer.window() >> self.remote_win_shift) as u16 >
-            self.remote_last_win
+        match self.state {
+            State::SynSent | State::SynReceived | State::Established | State::FinWait1 | State::FinWait2 =>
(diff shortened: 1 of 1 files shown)


whitequark (approved):

whitequark:
Thank you!

PR #367, opened 2020-08-10
Adjusting TCP connection 3-way handshake state management

### LINKED ISSUE #366: TcpSocket::connect does not properly handle an already connected socket
It appears that smoltcp does not properly handle an unexpected ACK when in the SYNC-SENT state. This results in the TCP socket being in an unusable state.

These conditions are described [here](https://serverfault.com/questions/733681/server-sends-ack-in-response-to-syn-causing-a-reset-in-tcp-request-from-the-clie), but the basics are that a previous connection (which was not properly terminated with a FIN) results in the server having some established TCP state (server is in ESTABLISHED).

When the client sends the SYN packet, the server acknowledges with an ACK#x, which is invalid to receive in the SYNC-SENT state. `smoltcp` should recognize this state and instead send a RST to the server to properly terminate the previous connection. It should then restart the SYN handshake to establish a connection with the server.

Sample of wireshark capture showing the error
image

ryan-summers:
This PR fixes #366 by adding an `abort()` if an `ACK` is received without the `SYN` flag set when in `SYN-SENT`.

The implications of this are that the user will have to re-`connect` the TCP socket after this occurrence.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -2500,6 +2500,18 @@ mod test {
     }
 
     #[test]
+    fn test_syn_sent_bad_ack() {
+        let mut s = socket_syn_sent();
+        send!(s, TcpRepr {
+            control: TcpControl::None,
+            seq_number: REMOTE_SEQ,
+            ack_number: Some(TcpSeqNumber(1)),
+            ..SEND_TEMPL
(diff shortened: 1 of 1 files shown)


whitequark:
Could you add a test please?

ryan-summers:
Test should now be added

PR #368, opened 2020-08-11
Allow for ARP retry during egress

### LINKED ISSUE #319: ARP requests not coming through UDP "flood"
I'm using smoltcp as the network stack on a STM32 embedded controller via https://github.com/stm32-rs/stm32-eth (cc @astro )

I've written a small test firmware that sends out UDP packets as fast as possible, saturating the 100 MBit link, and the device stops sending after 60 seconds, which is the timeout for the receiver's IP address in the ARP neighbor cache.

It looks like when the buffers are full (they are pretty small of course), smoltcp can't find anywhere to queue the ARP request and a kind of a deadlock is the result: no packets can go out anymore. Once I send a packet back to the device, this fills the cache and resolves the deadlock until the timeout hits again.

Do you think this is an issue with smoltcp, with the stm32-eth driver or is it to be expected and won't fix?

thalesfragoso:
The stack would send an ARP and put the `Neighbor` into the waiting state if it couldn't find a valid (non expired) hardware address, but it wouldn't allow for retries, stalling the socket indefinitely if we didn't get a response. These changes allow for retries after a `silent timer` has expired.

Closes https://github.com/smoltcp-rs/smoltcp/issues/319

CC @birkenfeld

### DIFF
--- a/src/socket/meta.rs
+++ b/src/socket/meta.rs
@@ -71,7 +71,7 @@ impl Meta {
                     self.neighbor_state = NeighborState::Active;
                     true
                 } else if timestamp > silent_until {
-                    net_trace!("{}: Retrying egress for neighbor {}", self.handle, neighbor);
+                    net_trace!("{}: neighbor {} silence timer expired, rediscovering", self.handle, neighbor);
                     true
                 } else {
                     false
--- a/src/socket/meta.rs
+++ b/src/socket/meta.rs
@@ -58,17 +58,20 @@ impl Meta {
         }
     }
 
-    pub(crate) fn egress_permitted<F>(&mut self, has_neighbor: F) -> bool
+    pub(crate) fn egress_permitted<F>(&mut self, timestamp: Instant, has_neighbor: F) -> bool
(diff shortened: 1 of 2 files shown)


src/socket/meta.rs
```diff
@@ -58,18 +58,21 @@ impl Meta {
...
                 true,
-            NeighborState::Waiting { neighbor, .. } => {
+            NeighborState::Waiting { neighbor, silent_until } => {
                 if has_neighbor(neighbor) {
                     net_trace!("{}: neighbor {} discovered, unsilencing",
                                self.handle, neighbor);
                     self.neighbor_state = NeighborState::Active;
                     true
+                } else if timestamp > silent_until {
+                    net_trace!("{}: Retrying egress for neighbor {}", self.handle, neighbor);
```
whitequark:
```suggestion
                    net_trace!("{}: neighbor {} silence timer expired, rediscovering", self.handle, neighbor);
```

whitequark (commented):
LGTM, with a small suggestion to improve clarity of the log message

PR #369, opened 2020-08-12
Only limit the neighbor cache rate after sending a request packet

### LINKED ISSUE #319: ARP requests not coming through UDP "flood"

thalesfragoso:
Prior to this change, the neighbor cache rate would get limited even when the device failed to process our arp packet, which could put the socket in the waiting state in the next poll.
This change only limits the cache rate if the device successfully consumed our packet.

Closes https://github.com/smoltcp-rs/smoltcp/issues/319 ~~again~~

For more information refer to [my comment on the issue](https://github.com/smoltcp-rs/smoltcp/issues/319#issuecomment-672062520).

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -1618,8 +1618,6 @@ impl<'b, 'c, 'e> InterfaceInner<'b, 'c, 'e> {
 
                     arp_repr.emit(&mut ArpPacket::new_unchecked(frame.payload_mut()))
                 })?;
-
-                Err(Error::Unaddressable)
             }
 
             #[cfg(feature = "proto-ipv6")]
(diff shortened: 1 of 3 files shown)


whitequark:
The change seems fine as-is, but I think that the `lookup_pure`/`lookup` distinction lost its meaning. Do you think you could merge those into just `lookup`? I'll leave the details to your judgement.

thalesfragoso:
Makes sense, I will try something.

thalesfragoso:
Ok, done. Let me know what you think of my approach.

whitequark (approved):
LGTM

PR #373, opened 2020-08-19
Changed Dhcpv4Client to use device checksum capabilities

### LINKED ISSUE #371: DHCP ignores device checksum capabilities
`Dhcpv4Client` ignores the device checksum capabiltities, and always uses the default value for `ChecksumCapabilities`. The code responsible can be seen here:
https://github.com/smoltcp-rs/smoltcp/blob/bdfa44270e9c59b3095b555cdf14601f7dc27794/src/dhcp/clientv4.rs#L133

As far as I can tell, this would basically be a one-line fix:
```rust
let checksum_caps = iface.device().capabilities().checksum;
```

This code breaks DHCP for any device that doesn't use the default checksum capabilities.

Spadi0:
Fixes #371 by fetching the checksum capabilities from the `Device`.

Previously, the checksum capabilities were always set to `ChecksumCapabilities::default()`, which could mess with the checksum generation.

### DIFF
--- a/src/dhcp/clientv4.rs
+++ b/src/dhcp/clientv4.rs
@@ -130,7 +130,7 @@ impl Client {
     where
         DeviceT: for<'d> Device<'d>,
     {
-        let checksum_caps = ChecksumCapabilities::default();  // ?
+        let checksum_caps = iface.device().capabilities().checksum;
         let mut raw_socket = sockets.get::<RawSocket>(self.raw_handle);
 
         // Process incoming


whitequark:
Hm, the tests are all broken, not sure why.

Spadi0:
@whitequark From what I can see, the broken tests have occurred on at least one more PR (#372).

I did some digging, and it seems to be something to do with the nightly toolchain. Running the tests locally using the [.test_like_travis.rb](https://github.com/smoltcp-rs/smoltcp/blob/bdfa44270e9c59b3095b555cdf14601f7dc27794/.test_like_travis.rb) file, it seems to work fine on `rustc 1.47.0-nightly (7e6d6e5f5 2020-08-16)`, but breaks on `rustc 1.47.0-nightly (792c645ca 2020-08-17)`.

Evidentally, something that was commited between those versions breaks something, but I'm not sure what or why.

PR #374, opened 2020-08-21
Add SocketHandle::inner

nbaksalyar:
Since `Hash` and `Display` are already implemented for `smoltcp::socket::set::Handle`, the numeric value of the handle is obtainable through these indirect means. I think it makes sense to expose a method to get this value explicitly.

### DIFF
--- a/src/socket/set.rs
+++ b/src/socket/set.rs
@@ -25,6 +25,13 @@ impl fmt::Display for Handle {
     }
 }
 
+impl Handle {
+    /// Return a numeric value of the handle.
+    pub fn inner(&self) -> usize {
+        self.0
+    }
+}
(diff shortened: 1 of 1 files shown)


whitequark:
I would specifically like to avoid exposing the internals of `SocketHandle` because this keeps the possibility of efficiency improvements in the future, e.g. tracking whether the socket was written to without O(n) iteration in `poll()`.

nbaksalyar:
Makes sense! How would you recommend passing it through the FFI boundary? Should it be treated as `*const SocketHandle`?

whitequark:
> Should it be treated as `*const SocketHandle`?

That definitely works, yeah.

PR #377, opened 2020-09-18
Subnet broadcasts

MabezDev:
I previously opened this before, but I think it got lost the repo ownership changes.

Currently smoltcp only responds to physical layer broadcasts (255.255.255.255), this PR enables the interface to respond to subnet broadcasts, as discussed in https://tools.ietf.org/html/rfc917 chapter 8.

## Questions

- Should this check be applied to all `ip_addrs` associated with the iface?

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -1023,7 +1023,9 @@ impl<'b, 'c, 'e> InterfaceInner<'b, 'c, 'e> {
 
         if !self.has_ip_addr(ipv4_repr.dst_addr) &&
            !ipv4_repr.dst_addr.is_broadcast() &&
-           !self.has_multicast_group(ipv4_repr.dst_addr) {
+           !self.has_multicast_group(ipv4_repr.dst_addr) &&
+           !self.is_subnet_broadcast(ipv4_repr.dst_addr) {
+
             // Ignore IP packets not directed at us, or broadcast, or any of the multicast groups.
             // If AnyIP is enabled, also check if the packet is routed locally.
             if !self.any_ip ||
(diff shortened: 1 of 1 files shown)


MabezDev:
Seems like the travis errors are unrelated, as they fail on current master too?

whitequark:
No idea what happened with tests, shouldn't that be impossible with protected branches, anyway?

Dirbaio:
Yes, master is broken, with ~~both stable and~~ nightly. EDIT see #378 What's weird is that commits that Travis previously said passed tests are now not passing.

whitequark:
Should probably be squashed, looks reasonable otherwise.

Dirbaio (changes requested):
> Should this check be applied to all ip_addrs associated with the iface?

I'd say yes. We process unicast packets for all our IPs, IMO it's quite strange that we only process broadcast packets for the first one.

Looks great otherwise! :)

Dirbaio (approved):
Looks great now that it checks all our addrs!

Commits got somehow duplicated when merging, can you rebase on top of master?

MabezDev:
> Looks great now that it checks all our addrs!
> 
> Commits got somehow duplicated when merging, can you rebase on top of master?

Will do, some tests are still failing so I'll fix those first and then squash.

MabezDev:
Should be good to go :)

Dirbaio:
There's still a clippy failure :(

MabezDev:
Whoops, must have missed that! Should be fixed now, sorry about that :)

MabezDev:
Fairly minor change, but I realised only ipv4 has the concept of broadcasts, so I removed a pointless match, and instead just pass the ip directly.

Dirbaio:
Awesome! Thank you!

PR #380, opened 2020-10-09
Fix `RawSocket` buffer size

vlad9486:
The buffer of `RawSocket` should account Ethernet header.

### DIFF
--- a/src/phy/raw_socket.rs
+++ b/src/phy/raw_socket.rs
@@ -7,6 +7,7 @@ use std::os::unix::io::{RawFd, AsRawFd};
 use Result;
 use phy::{self, sys, DeviceCapabilities, Device};
 use time::Instant;
+use wire::EthernetFrame;
 
 /// A socket that captures or transmits the complete frame.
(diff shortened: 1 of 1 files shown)


whitequark (changes requested):
I don't think this is right. Could you please explain the exact rationale for this change?

vlad9486:
> I don't think this is right. Could you please explain the exact rationale for this change?

`sys::RawSocketDesc::recv` gives the packet which might be up to `mtu` bytes long *plus* the Ethernet header (or some other hardware header), if the packet length is not fit, it just cutoff the last bytes of payload. I actually saw it in the debugging logs.

Dirbaio:
The root cause of the bug is that Linux's MTU is the IP mtu, while smoltcp's MTU is the Ethernet MTU (ie, including the Ethernet header). The fix is to update RawSocketDesc to return the correct MTU, done in #393.

I've also opened #392 to track improving the docs. I totally see how this can be very confusing the way it is now.

PR #381, opened 2020-10-21
igmp: centisecs are really decisecs

Dirbaio:
1 centisecond = 1/100th of a second
1 decisecond = 1/10th of a second

The field in the IGMP packets is in deciseconds.


whitequark:
Thanks!

PR #382, opened 2020-10-21
WIP Async/await support

Dirbaio:
This adds support for using TcpSocket with async/await code. If the approach proves good I'll add wakers to the other sockets too.

The approach is to do the minimum required changes to core smoltcp, which allow making higher-level wrappers that can be used with async/await. This way smoltcp can stay un-opinionated:
- The `no_std` ecosystem hasn't converged on any standard io traits yet. The Rust standard is `futures::io::{AsyncRead/AsyncWrite}`, but these aren't `no_std` compatible. 
- There are non-trivial tradeoffs when implementing the wrappers (how are sockets allocated, is there a single global tcp/ip stack or multiple, how are lifetimes handled).

I'm working on PoC wrappers, will report back soon!

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -10,6 +10,10 @@ use time::{Duration, Instant};
 use socket::{Socket, SocketMeta, SocketHandle, PollAt};
 use storage::{Assembler, RingBuffer};
 use wire::{IpProtocol, IpRepr, IpAddress, IpEndpoint, TcpSeqNumber, TcpRepr, TcpControl};
+#[cfg(feature = "async")]
+use socket::WakerStore;
+#[cfg(feature = "async")]
+use core::task::Waker;
 
 /// A TCP socket ring buffer.
 pub type SocketBuffer<'a> = RingBuffer<'a, u8>;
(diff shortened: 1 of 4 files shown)


whitequark (approved):
I think this is a great approach, and I'd like to merge this PR as soon as it's ready. There really are no downsides I see.

src/socket/tcp.rs
```diff
@@ -825,7 +856,17 @@ impl<'a> TcpSocket<'a> {
...
             }
         }
-        self.state = state
+
+        self.state = state;
+
+        #[cfg(feature = "async")]
+        {
+            // Wake all tasks waiting. Even if we haven't received/sent data, this
+            // is needed because return values of functions may change depending on the statem.
```
astro, line 865: Typo: `statem`?

astro:
FWIW, we have some async glue here: https://git.m-labs.hk/M-Labs/zynq-rs/src/branch/master/libasync/src/smoltcp

This new approach with wakers per socket looks more efficient.

Dirbaio:
Moved to https://github.com/smoltcp/smoltcp/pull/130

PR #384, opened 2020-10-23
Fixes unused MTU settings at TcpSocket dispatch

mustermeiszer:
This commit fixes a small bug, where the TcpSocket computed the maximum
size per payload in a tx_buffer without taking into account the MTU
defined by the underlying network device.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1559,7 +1559,7 @@ impl<'a> TcpSocket<'a> {
                 // from the transmit buffer.
                 let offset = self.remote_last_seq - self.local_seq_no;
                 let size = cmp::min(cmp::min(self.remote_win_len, self.remote_mss),
-                     caps.max_transmission_unit);
+                     caps.max_transmission_unit - ip_repr.buffer_len() - repr.mss_header_len());
                 repr.payload = self.tx_buffer.get_allocated(offset, size);
                 // If we've sent everything we had in the buffer, follow it with the PSH or FIN
                 // flags, depending on whether the transmit half of the connection is open.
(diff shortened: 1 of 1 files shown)


src/socket/tcp.rs
```diff
@@ -1558,7 +1558,8 @@ impl<'a> TcpSocket<'a> {
                 // Extract as much data as the remote side can receive in this packet
                 // from the transmit buffer.
                 let offset = self.remote_last_seq - self.local_seq_no;
-                let size = cmp::min(self.remote_win_len, self.remote_mss);
+                let size = cmp::min(cmp::min(self.remote_win_len, self.remote_mss),
+                     caps.max_transmission_unit);
```
Dirbaio: I'm not convinced this is correct. MTU is IP packet size (IP header + TCP header + payload). `size` here is only payload size.

You have to subtract the IP and TCP header sizes, as in here : https://github.com/smoltcp-rs/smoltcp/blob/1268fb3d8ddeb1814a6b3a49f1647d1bbf3debd8/src/socket/tcp.rs#L1624

Dirbaio (changes requested):

src/socket/tcp.rs
mustermeiszer: Ah, so much for a quick fix on my part. Haha. 

Is this fine?

```rust
let size = cmp::min(cmp::min(self.remote_win_len, self.remote_mss),
                     caps.max_transmission_unit - ip_repr.buffer_len() - repr.mss_header_len());
repr.payload = self.tx_buffer.get_allocated(offset, size);
```

Dirbaio: Sure!

Dirbaio (approved):

whitequark:
Thanks @mustermeiszer for the fix and @Dirbaio for the review!

PR #387, opened 2020-12-18
tcp: don't send data outside the remote window

Dirbaio:
If we receive a packet with `ack_seq=1000, win_len=10`, this means the remote end is granting us permission to send bytes up to `seq=1010` but no more.

The current implementation limits segment length to `win`, but does not limit the byte count. In the example above, it'll send many small packets:  `1000-1010`, `1010-1020` ... until the entire tx_buffer is sent.

This fixes the window handling so that only `1000-1010` is sent, and the rest isn't sent until the remote end sends a window update opening more space.

- Updated 1 test that tested for the previous behavior: `test_established_send_buf_gt_win`
- Updated a few tests that used the window to force sending small packets: they now use MSS.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -230,8 +230,8 @@ pub struct TcpSocket<'a> {
     /// The sending window scaling factor advertised to remotes which support RFC 1323.
     /// It is zero if the window <= 64KiB and/or the remote does not support it.
     remote_win_shift: u8,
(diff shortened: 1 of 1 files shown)


whitequark (approved):
At a glance the change seems reasonable, but I haven't thought about it in-depth, so I primarily trust that you have tested it well.

whitequark:
Thanks!

PR #388, opened 2020-12-18
tcp: allow sending ACKs in FinWait2 state.

Dirbaio:
in FinWait2 the connection is "half-closed": we may no longer send data, but we still can receive data. This data has to be ACKed.

The current code forbids anything to be sent in FinWait2, this has 2 problems:
- ACKs are not sent.
- A polling infinite busy loop happens, because `poll_at` still returns PollAt::Now (because `ack_to_transmit` returns `true`), but dispatch doesn't actually do anything.

This extends a test to cover the failure, then fixes it.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1574,8 +1574,9 @@ impl<'a> TcpSocket<'a> {
                 }
             }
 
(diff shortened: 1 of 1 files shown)


whitequark (approved):

whitequark:
Thanks!

Issue #389, opened 2020-12-18
Migrate CI to GitHub Actions


Dirbaio:
Travis is extremely slow lately due to them limiting testing resources for opensource projects. I sent a PRs 45min ago and it's still testing :(

It'd be nice to migrate to GHA which is still free and doesn't have this problem.

whitequark:
Agreed entirely.

Dirbaio:
Done!

PR #390, opened 2020-12-22
Add Github Actions for CI

adamgreig:
This is a first attempt at resolving #389. I expect it will need some fine-tuning; feedback welcome.

I believe I've included everything from the current Travis config except for checking the bench, which seems to be broken at the moment anyway.

Edit: It seems like building on 1.28.0 is broken as it can't parse the `Cargo.toml` file of `cfg-if` or other dependencies, although the README suggests that _should_ be the MSRV. It looks like even once that issue is resolved, the use of `MaybeUninit` requires 1.36, which does build OK, so I've set a check at 1.36 in the CI and updated README.

Edit 2: It looks like the test without std is failing due to a [whole bunch of errors](https://github.com/smoltcp-rs/smoltcp/runs/1594784110?check_suite_focus=true); I've removed it for now.

Edit 3: Tests just won't pass without std or alloc, so I've swapped it to only check the build for that feature-set.


jordens:
Nice!
Feel free to bump the msrv. 1.28 was because of artiq only. No need to block on that.

adamgreig:
Ok, this is ready for review. At the moment Travis is still enabled too, but we could remove it in this PR.

whitequark:
Thank you for tackling this! As far as I'm concerned this is ready. You can just get rid of Travis.

whitequark (approved):

whitequark:
I'm not sure who can actually merge this PR though; for some reason I don't have the permission bit for manually merging to protected branches, and the Travis required status will never pass now.

adamgreig:
I guess @jordens? I was previously an admin of the organisation but apparently no longer, so I can't disable the Travis required check either.

whitequark:
I'm pretty sure I had repo admin at one point but seemingly no longer? No idea what's up with that.

therealprof:
I also don't have the necessary permissions.

Dirbaio:
Given the hostile takeover we've started a new repo here -> https://github.com/smoltcp/smoltcp
I've merged this PR there, working great!

jordens:
I don't know why or who set these permissions up, I don't think it makes sense and it doesn't seem desirable to me. I have made the three of you owners.

therealprof:
> I don't know why or who set these permissions up, I didn't I don't think it makes sense and it doesn't seem desirable to me. I have made the three of you owners.

Thanks!

whitequark:
To ensure the long-term health of the project, I went ahead and did two things:
  1. I went through the GitHub organization audit log, determined the cause of the permission change that prevented this PR from being merged, and made the minimal amount of changes to ensure that this sort of disruption will not happen again.
  2. As you may have noticed, I've only had a limited amount of time to spend on smoltcp lately. As the original author of the codebase, I believe that @Dirbaio would make a great smoltcp maintainer, and he is also in a good position to be one. I have invited him to the organization, and he has my blessing in determining the future direction of the project to the same degree as I do.

I think smoltcp has a great future ahead of it, and I look forward to seeing it evolve in 2021 with a new maintainer.

Issue #267, opened 2018-12-19
Tokio integration


DemiMarie:
It would be very nice to be able to use smoltcp with the Tokio ecosystem.  For me, the first use case that comes to mind is extremely high-performance servers that use kernel bypass networking, but there are probably many, many others.

whitequark:
Sure. I am not personally invested in Tokio so I will not work on it myself.

Gymmasssorla:
This is probably a good idea, but it's strange that high-level abstractions like futures will be in such a low-level crate.

whitequark:
I think futures are pretty low level, it's just another way to represent control flow. In general, microcontroller firmware often greatly benefits from using async/await-like control flow rather than threads; there are many coroutine replacements (not really implementations) for C specifically for this task.

Gymmasssorla:
Agree with you.

stlankes:
Do you still working on a Tokio integration? Do you have already a test example?

dingelish:
very interested in tokio integration as well. do you need help?

whitequark:
Sure, would be helpful to have even a proof of concept.

Dirbaio:
I'd love to see async/await support in smoltcp too!

From some experiments I've done (#382), I believe the best approach is to:
- Add support in smoltcp to associate `Waker`s to sockets.
- Leave everything else to external higher-level crates.

The higher-level crates would implement wrappers for the smoltcp types that implement the desird async traits (AsyncRead, AsyncWrite, Stream for UDP packets..?).

I believe this is the best way to go, because:

- It is not clear which traits should smoltcp implement:
  - The `std` ecosystem has multiple competing incompatible traits, mainly `tokio` and `futures`.
  - The `no_std` ecosystem hasn't converged on any standard io traits yet. The `tokio` and `futures` traits aren't `no_std` compatible. 
- There are non-trivial tradeoffs when implementing async support: how are sockets allocated, is there a single global tcp/ip stack or multiple, how are lifetimes handled.

Closing this as we won't be integrating with Tokio specifically, but the PR adding `Waker` support is coming very soon! :)

PR #394, opened 2020-12-26
Async/await waker support.

Dirbaio:
This adds `register_recv_waker` and `register_send_waker` methods to all socket types. 

This is the minimum required changes to smoltcp to support being used in async/await code. This opens the door for higher-level crates that expose wrappers implementing async traits like AsyncRead/AsyncWrite, Future, Stream...

I believe this is the best way to go, because:

- It is not clear which traits should smoltcp implement:
  - The `std` ecosystem has multiple competing incompatible traits, mainly `tokio` and `futures`.
  - The `no_std` ecosystem hasn't converged on any standard io traits yet. The `tokio` and `futures` traits aren't `no_std` compatible. 
- There are non-trivial tradeoffs when implementing async support: how are sockets allocated, is there a single global tcp/ip stack or multiple, how are lifetimes handled. 

By delegating these choices to external crates, smoltcp can stay un-opinionated.

There's an example of using this to implement AsyncBufRead and AsyncWrite in `embassy` [here](https://github.com/embassy-rs/embassy/blob/master/embassy-net/src/tcp_socket.rs). Wrappers for the `tokio` or `futures` traits would look very similar. The use of `unsafe` there is not necessary in theory, AsyncRead and AsyncWrite can be implemented in 100% safe Rust. AsyncBufRead probably will require unsafe in most cases, unfortunately.

Previously discussed on #382 and #267 .

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           - std ethernet proto-ipv6 socket-icmp socket-tcp
 
           # Test features chosen to be as aggressive as possible.
-          - std ethernet proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp
+          - std ethernet proto-ipv4 proto-ipv6 socket-raw socket-udp socket-tcp socket-icmp async
 
         include:
           # Test alloc feature which requires nightly.
@@ -66,7 +66,7 @@ jobs:
 
         features:
           # These feature sets cannot run tests, so we only check they build.
-          - ethernet proto-ipv6 proto-ipv6 proto-igmp proto-dhcpv4 socket-raw socket-udp socket-tcp socket-icmp
+          - ethernet proto-ipv6 proto-ipv6 proto-igmp proto-dhcpv4 socket-raw socket-udp socket-tcp socket-icmp async
 
     steps:
       - uses: actions/checkout@v2
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,12 +42,14 @@ ethernet = []
 "socket-udp" = []
 "socket-tcp" = []
 "socket-icmp" = []
+"async" = []
 default = [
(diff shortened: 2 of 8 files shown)


src/socket/waker_store.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+impl WakerStore {
+    pub const fn new() -> Self {
+        Self { waker: None }
+    }
+
+    pub fn register(&mut self, w: &Waker) {
+        match self.waker {
+            Some(ref w2) if (w2.will_wake(w)) => {}
```
whitequark: I think it's worth adding a comment here explaining this logic.

whitequark (commented):
LGTM other than inline comments. Nice work!

src/socket/waker_store.rs
Dirbaio: Done!

Charles-Schleich:
Looks like the AsyncBufRead and AsyncWrite example for  `embassy` mentioned earlier has been moved [here](https://github.com/embassy-rs/embassy-net/blob/main/embassy-net/src/tcp_socket.rs). 
For those having trouble finding it.

Dirbaio:
Now moved here https://github.com/embassy-rs/embassy/blob/master/embassy-net/src/tcp_socket.rs

cameronelliott:
All the previous embassy links are broken. 
Here is an example of using await with embassy, but this is with a serial port not a TCP socket, but the principals may be the same:
https://github.com/embassy-rs/embassy/blob/53fbd0efb3e77e1e3de948afde2b5bf1a5a9735f/examples/std/src/bin/serial.rs#L25

cameronelliott:
This example link shouldn't break unless they re-write history:

https://github.com/embassy-rs/embassy/blob/79cbad501c6e6166d0e53ab33b2401a674ee9f31/embassy-net/src/tcp_socket.rs#L23

PR #395, opened 2020-12-26
Various cleanup suggested by cargo clippy

crawford:
I noticed that PRs tend to get cluttered with a lot of warnings from `cargo clippy`, so I went through and started addressing some of the suggestions where I thought they make sense. Here's a collection of changes that I thought were non-controversial, but I have a bunch more locally, if folks think these sorts of changes are helpful.

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -91,7 +91,7 @@ fn main() {
             let client = match socket.recv() {
                 Ok((data, endpoint)) => {
                     debug!("udp:6969 recv data: {:?} from {}",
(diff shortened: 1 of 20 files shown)


whitequark (approved):

Dirbaio:
Thanks for this. Please do send the others! If we get all the lints cleaned up we could set clippy to hard-fail CI so they aren't reintroduced.

Some should be disabled instead. For example `match expression looks like `matches!` macro`: the `matches!` macro doesn't work in Rust 1.36

PR #396, opened 2020-12-26
Update to Rust 2018

Dirbaio:


Dirbaio:
Merging now, as this is all boring boilerplate that should be uncontroversial and is highly likely to to create annoying merge conflicts.

PR #398, opened 2020-12-27
tcp: fix racey simultaneous close not sending FIN.

Dirbaio:
This fixes the issue identified in PR #279 

In CLOSING state we may still have data or FIN to retransmit, or to transmit for the first time if we received a FIN before we had a chance to transmit it. 

This fixes the issue by treating CLOSING similarly to FinWait1 and LastAck. This ensures data/FIN is always sent, and retransmitted in the case of timeout.


whitequark (approved):
LGTM, though it's been a while since I worked on TCP and so for the most part I trust that you are making the correct change here.

PR #346, opened 2020-06-04
Add support for IP mediums, v2

Dirbaio:
This is a new attempt at implementing #334, addressing the shortcomings found in the first version (#336) detailed in https://github.com/smoltcp-rs/smoltcp/pull/336#issuecomment-638520316.

So far I'm quite pleased by how it turned out. The only big structural change in `Interface` is checking for `Device.medium()` in send and receive paths and adjusting the behavior accordingly. The rest of the changes are mostly boilerplate and updating the tests.

This includes the Linux tun work by @lucaszanella from Dirbaio/smoltcp#1, squashed and updated for the v2 changes.

Things that need fixing:
- [ ] The prettyprinter is broken (it doesn't take `medium` into account, always interprets as Ethernet)

Possible future work:
- [X] Adding a `medium_ip` feature DONE
- [X] Renaming `ethernet` feature to `medium_ethernet` for consistency? DONE
- [ ] Doing a PoC of instantiating an `Interface` with dynamic dispatch to the device, check how much improvement in binary size is there vs static dispatch.

PR #399, opened 2020-12-27
Add IP medium support, part 1

Dirbaio:
Previously #336 #346

I'm sending this in several parts since the PR is very big otherwise. This first part refactors iface::Packet.

- Remove "None" variant from Packet, use Option<Packet> instead.

- Split Packet into EthernetPacket and IpPacket.

    Functions that only deal with IP packets take/return IpPacket's. IpPacket's
are wrapped into EthernetPacket's as late as possible.

    This will later allow generalizing Interface to handle both Ethernet
and pure-IP mediums.


whitequark (approved):
LGTM, nothing to object to.

PR #400, opened 2020-12-27
Add support for IP mediums, part 3

Dirbaio:
Previously #336 #346. includes #399, review that one first

TODO: 
- [ ] Prettyprinter doesn't work

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,23 +23,23 @@ jobs:
           - default
 
           # Test features chosen to be as orthogonal as possible.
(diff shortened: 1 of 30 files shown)


jan4984:
how about the progress of tun support? I want use it in Android VPN service app which use tun interface.

Dirbaio:
#401 has working code, people have used it successfully with tun including on android :)

It should be on master soon hopefully

JonForShort:
Support for tun interface would be great to have.  I am also looking into integrating smoltcp into an Android VPN service.  Thanks @Dirbaio for working on this.

PR #401, opened 2020-12-27
Add IP medium support, part 2

Dirbaio:
Previously #336 #346. ~~includes #399, review that one first~~

TODO: 
- [x] Add a way for examples to use TAP or TUN. For example with --tap / -- tun. Not sure how to do this because `Device` is not object-safe.
- [x] Prettyprinter doesn't work
- [x] pcap doesn't work

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,23 +26,23 @@ jobs:
           - std proto-ipv4
 
           # Test features chosen to be as orthogonal as possible.
-          - std ethernet phy-raw_socket proto-ipv6 socket-udp
-          - std ethernet phy-tap_interface proto-ipv6 socket-udp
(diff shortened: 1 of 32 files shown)


Dirbaio:
- Changed tun/tap code so there's now a single TunTapDevice instead of TapDevice and TunDevice. Less code duplication, nicer for code that wants to use both.
- Added --tun and --tap options to examples so they can use either.
- Updated PcapWriter to write the right LinkType based on the medium.
- Updated Tracer to prettyprint with the right type based on the medium.

@whitequark this is ready to go :) 

This is now stuck with 22 pending tests that will never happen because the Cargo features have been renamed... Best solution is to probably tweak repo settings to mark the new checks as required, then merge..?

Issue #397, opened 2020-12-27
Use #[non_exhaustive] instead of manual __Nonexhaustive variants


whitequark:
It used to not be available in our MSRV, but that got bumped recently.

Dirbaio:
`#[non_exhaustive]` was introduced in 1.40, our MSRV is 1.36. We shouldn't do this yet, right?

whitequark:
1.40 is a year old by now, I would just bump MSRV to 1.40.

niclashoyer:
I added a PR for this one. I'm not sure how MSRV is handled, though.

PR #402, opened 2020-12-27
Some more cleanup suggested by cargo clippy

crawford:
This is a follow-on to https://github.com/smoltcp-rs/smoltcp/pull/395. I still have some more changes locally, but I can follow up with another pull request once I get them cleaned up a bit.

### DIFF
--- a/src/wire/icmpv6.rs
+++ b/src/wire/icmpv6.rs
@@ -420,7 +420,7 @@ impl<T: AsRef<[u8]> + AsMut<[u8]>> Packet<T> {
             Message::MldQuery => {
                 let data = self.buffer.as_mut();
                 NetworkEndian::write_u16(&mut data[field::QUERY_RESV], 0);
-                data[field::SQRV] = data[field::SQRV] & 0xf;
+                data[field::SQRV] &= 0xf;
             },
             Message::MldReport => {
                 let data = self.buffer.as_mut();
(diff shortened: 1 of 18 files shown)


src/wire/icmpv6.rs
```diff
@@ -68,6 +69,7 @@ impl Message {
     ///
     /// [MLD]: https://tools.ietf.org/html/rfc3810
     pub fn is_mld(&self) -> bool {
+        #[allow(clippy::match_like_matches_macro)]
```
Dirbaio: Given that our MRSV is 1.36 and `matches!` is not supported there, it'd make more sense to allow it for the entire crate (`#![allow(clippy::match_like_matches_macro)]` in `lib.rs`)

src/wire/icmpv4.rs
```diff
@@ -366,6 +366,7 @@ impl<T: AsRef<[u8]>> AsRef<[u8]> for Packet<T> {
 
 /// A high-level representation of an Internet Control Message Protocol version 4 packet header.
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[allow(clippy::manual_non_exhaustive)]
```
Dirbaio: I'd update these to use  `#[non_exhaustive]`, instead of adding an `#[allow]` for the old way:  #397

Dirbaio (changes requested):

src/wire/icmpv4.rs
crawford: That was introduced in Rust 1.40 and I didn't want to bump the minimum version.

Dirbaio: Oh okay, then I don't know what's up with #397 . Could you change it to a crate-wide allow then? Thank you!

Dirbaio:
Very nice! Thanks!

PR #403, opened 2020-12-29
Last round of suggestions by clippy

crawford:
This is a follow-on to https://github.com/smoltcp-rs/smoltcp/pull/402 and is the last in the series. Now that all of the warnings have been addressed, I've also set `cargo clippy` to fail when it encounters warnings going forward.

Should we pin to a specific toolchain or do folks think it's feasible to keep up-to-date with the latest stable? I don't have a sense for how often new lints are added.

### DIFF
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::collapsible_if)]
+
 mod utils;
 
 use std::cmp;
--- a/src/phy/loopback.rs
+++ b/src/phy/loopback.rs
@@ -44,7 +44,7 @@ impl<'a> Device<'a> for Loopback {
 
     fn receive(&'a mut self) -> Option<(Self::RxToken, Self::TxToken)> {
         self.queue.pop_front().map(move |buffer| {
-            let rx = RxToken { buffer: buffer };
+            let rx = RxToken { buffer };
             let tx = TxToken { queue: &mut self.queue };
             (rx, tx)
         })
--- a/src/phy/loopback.rs
+++ b/src/phy/loopback.rs
@@ -19,6 +19,7 @@ pub struct Loopback {
     queue: VecDeque<Vec<u8>>,
 }
 
+#[allow(clippy::new_without_default)]
 impl Loopback {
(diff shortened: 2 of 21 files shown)


examples/benchmark.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::collapsible_if)]
```
whitequark, line 1: Should we just disable clippy in examples?

src/phy/loopback.rs
```diff
@@ -14,7 +14,7 @@ use crate::phy::{self, Device, DeviceCapabilities};
 use crate::time::Instant;
 
 /// A loopback device.
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
whitequark: I don't think this is desirable; end users should use `Loopback::new`. `Loopback` in this case behaves like an OS resource, not just an object.

whitequark (changes requested):
I've reviewed individual commits and I agree with almost all of your judgement calls (except where noted). Good work!

examples/benchmark.rs
crawford: I tend to look at the examples first when I'm looking at a new project, so my _personal_ preference would be to make sure that the examples are just as tidy. But I'm happy to drop this if you'd rather not bother.

crawford:
I want to double check that the clippy check actually fails this time. I've got one final commit to address the last lint.

examples/benchmark.rs
whitequark: No objection if you prefer it this way.

whitequark (approved):

crawford:
Today I learned about [`pulL_request_target`](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/):

> we’ve added a new pull_request_target event, which behaves in an almost identical way to the pull_request event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request.

That's why the job isn't failing. I'm reasonably confident this will work as-is.

PR #405, opened 2021-01-04
clippy fixes

Dirbaio:


src/parsers.rs
```diff
@@ -88,11 +88,11 @@ impl<'a> Parser<'a> {
 
     fn accept_digit(&mut self, hex: bool) -> Result<u8> {
         let digit = self.advance()?;
-        if digit >= b'0' && digit <= b'9' {
+        if (b'0'..=b'9').contains(&digit) {
```
Sympatron, line 91: This might not be such a big deal, but does this generate code that is equally efficient?

crawford: Not quite.I just tried the following patch in my project  and it resulted in a small decrease in size:

```diff
diff --git a/src/parsers.rs b/src/parsers.rs
index bf87cdc..36bad29 100644
--- a/src/parsers.rs
+++ b/src/parsers.rs
@@ -88,11 +88,11 @@ impl<'a> Parser<'a> {

     fn accept_digit(&mut self, hex: bool) -> Result<u8> {
         let digit = self.advance()?;
-        if (b'0'..=b'9').contains(&digit) {
+        if digit >= b'0' && digit <= b'9' {
             Ok(digit - b'0')
-        } else if hex && (b'a'..=b'f').contains(&digit) {
+        } else if hex && digit >= b'a' && digit <= b'f' {
             Ok(digit - b'a' + 10)
-        } else if hex && (b'A'..=b'F').contains(&digit) {
+        } else if hex && digit >= b'A' && digit <= b'F' {
             Ok(digit - b'A' + 10)
         } else {
             Err(())
```

Before this change, the size was as follows:

```
section                 size        addr
.vector_table            336           0
.text                  41528         336
.rodata                 8496       41864
```

And after:

```
section                 size        addr
.vector_table            336           0
.text                  41404         336
.rodata                 8512       41740
```

PR #407, opened 2021-01-04
Pin clippy check to 1.49.0

crawford:
This also includes a handful of other clippy-inspired cleanups.


Dirbaio:
Sounds great! This way we can decide when to update it and deal with the breakage there, instead of in a random PR. Thanks!

PR #409, opened 2021-01-08
Use #[non_exhaustive] instead of manual variant

### LINKED ISSUE #397: Use #[non_exhaustive] instead of manual __Nonexhaustive variants

niclashoyer:
first issue :tada: 

Fixes #397

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,11 +11,11 @@ jobs:
     continue-on-error: ${{ matrix.rust == 'nightly' }}
     strategy:
       matrix:
-        # Test on stable, MSRV 1.36, and nightly.
+        # Test on stable, MSRV 1.40, and nightly.
         # Failure is permitted on nightly.
         rust:
           - stable
-          - 1.36.0
+          - 1.40.0
           - nightly
 
         features:
@@ -57,11 +57,11 @@ jobs:
     continue-on-error: ${{ matrix.rust == 'nightly' }}
     strategy:
       matrix:
-        # Test on stable, MSRV 1.36, and nightly.
+        # Test on stable, MSRV 1.40, and nightly.
         # Failure is permitted on nightly.
         rust:
           - stable
-          - 1.36.0
+          - 1.40.0
           - nightly
 
         features:
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ include complicated compile-time computations, such as macro or type tricks, eve
 at cost of performance degradation.
 
 _smoltcp_ does not need heap allocation *at all*, is [extensively documented][docs],
-and compiles on stable Rust 1.36 and later.
+and compiles on stable Rust 1.40 and later.
 
 _smoltcp_ achieves [~Gbps of throughput](#examplesbenchmarkrs) when tested against
 the Linux TCP stack in loopback mode.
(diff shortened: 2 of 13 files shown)


Dirbaio:
Very nice! 

Can you update `.github/workflows/test.yml` and `README.md`?

niclashoyer:
I'll try to, but I think there are other issues here. I just noticed that locally the tests run fine (I'm not sure if other feature combinations might fail, I can check that). But `cargo clippy` fails because I removed all `_ => ...` match arms.

I think clippy demands these wildcard arms, but `#![cfg_attr(deny(unused))]` demands removal of these match arms, since the code is unreachable. I'm not sure how to proceed here.

Dirbaio:
You can make the match arms themselves conditional. For example instead of  `_ => None`, you can do:

```rust
IpCidr::Ipv6 => None
```

niclashoyer:
ok, I'll check that. README and workflow is updated now :crossed_fingers:

crawford:
Are we okay bumping the minimum Rust version? I believe @whitequark and https://github.com/m-labs/ have a fork of Rust that they use with this project.

niclashoyer:
@crawford this was also mentioned in the issue: https://github.com/smoltcp-rs/smoltcp/issues/397#issuecomment-751544675

Meanwhile I need help again :sweat_smile: 

Adding the `cfg` attr to other match arms helped, but now I'm stuck at this macro in `src/socket/mod.rs`:

```rust
macro_rules! from_socket {
    ($socket:ty, $variant:ident) => {
        impl<'a, 'b> AnySocket<'a, 'b> for $socket {
            fn downcast<'c>(ref_: SocketRef<'c, Socket<'a, 'b>>) ->
                           Option<SocketRef<'c, Self>> {
                match SocketRef::into_inner(ref_) {
                    &mut Socket::$variant(ref mut socket) => Some(SocketRef::new(socket)),
                    _ => None,
                }
            }
        }
    }
}
```

This wildcard matches all other arms that are not given by the macro. Adding all other arms here would complicate this a lot, wouldn't it?

Dirbaio:
@whitequark ACKed the MSRV bump https://github.com/smoltcp-rs/smoltcp/issues/397#issuecomment-751544675 .

I'm also for bumping the MSRV. 1.36 is *ancient* and 1.40 is already a year old.

Additionally, this change is not only cosmetic. The extra dummy variants impact code size and struct layout.

Dirbaio:
> Adding the `cfg` attr to other match arms helped, but now I'm stuck at this macro in `src/socket/mod.rs`:

I think using `if let` should work there?

```rust
if let Socket::$variant(ref mut socket) = SocketRef::into_inner(ref_) { 
    Some(...)
} else {
    None
}
```

Dirbaio:
Maybe that still triggers some warnings if the `else` branch is unreachable, in this case I'd be OK with some `#[allow(..)]` on that `if`...

niclashoyer:
Better now. I'm still a bit lost here:

https://github.com/smoltcp-rs/smoltcp/pull/409/checks?check_run_id=1671886625

It seems that `TcpState` is behind `socket-tcp` which seems to trigger unused lifetimes. Is this new in 1.40?

Edit: no, `master` compiles fine on 1.40 using `cargo test --no-default-features --features "std ethernet proto-ipv6 socket-tcp"`.

Dirbaio:
Oh, that's very unfortunate.

With conditional compilation `'b`  *is* unused when just `socket-tcp` is enabled... Not sure what's the best solution.

niclashoyer:
ah this issue seemed like a low hanging fruit :cherries: I'll go to bed now, maybe I've got more ideas tomorrow...

Dirbaio:
Actually it turns out that having 2 lifetimes in sockets isn't really needed. 

I've opened PR #410 to join them in just `'a`. With that, the problem goes away :)

Dirbaio:
Thank you for the PR @niclashoyer !

PR #410, opened 2021-01-09
Simplify Socket lifetimes

Dirbaio:
`PacketBuffer` currently requires 2 different lifetimes for metadata and payload buffers. This causes a lot of cascading complexity causing Sockets to require 2 lifetimes and SocketSet to require 3 (!) lifetimes.

I can't think of any usage requiring 2 separate lifetimes in `PacketBuffer`. (ie something you *could* do with separate lifetimes but *can't* do with just 1). 

This PR removes it, yielding a nice simplification of lifetimes everywhere :)

### DIFF
--- a/src/dhcp/clientv4.rs
+++ b/src/dhcp/clientv4.rs
@@ -94,8 +94,7 @@ impl Client {
     ///     Instant::now()
     /// );
     /// ```
-    pub fn new<'a, 'b, 'c>(sockets: &mut SocketSet<'a, 'b, 'c>, rx_buffer: RawSocketBuffer<'b, 'c>, tx_buffer: RawSocketBuffer<'b, 'c>, now: Instant) -> Self
-    where 'b: 'c,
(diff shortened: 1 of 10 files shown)


whitequark:
Are you *completely sure*? Our examples don't test `no_std` lifetimes nearly enough.

Dirbaio:
Yes!

With the new code, when creating a PacketBuffer with buffers that could have different lifetimes (for example, one `'static` and one `'foo`), its `'a` lifetime becomes the smaller lifetime, which is `'foo`. The end result is the PacketBuffer can only live for `'foo`.

With the previous code, the PacketBuffer would get `'a` = `'static` and `'b` = `'foo`. The struct can only live for the min of all its lifetime params. The end result is the PacketBuffer can only live for `'foo`, again.

The only way single-vs-multiple lifetimes can make a difference is if there are APIs to "take out" the `&mut`s, or to "move" them somewhere else that's not the struct. If there's a single lifetime, all the `&mut`s are "downgraded" to the smallest lifetime because the information is "lost", if there are multiple each individual lifetime is kept. We don't have such APIs in PacketBuffer or Socket though.

Dirbaio:
I also think similar simplifications can be made for `Interface` and `SocketSet` too.

whitequark:
I thought it was necessary, but it seems like I misunderstood exactly how lifetimes work.

PR #411, opened 2021-01-10
Various cleanup

crawford:
This builds on 633e7c25 (#409), adding the `#[non_exhaustive]` attribute to applicable struct definitions and de-duplicates and (hopefully) simplifies a few if-else blocks. I've also verified that these changes don't result in larger binaries.


crawford:
cc @dlrobertson

Dirbaio (approved):
Very nice, thanks!

Issue #412, opened 2021-01-11
New crate.io release?


stlankes:
The latest crates.io release (version 0.6.0) is from Januar 2020. Are there any plans to release a new version?

whitequark:
At the moment only I have crates.io permissions. @Dirbaio could you set up a team in the GitHub org that I could add to crate owners?

Dirbaio:
[done](https://github.com/orgs/smoltcp-rs/teams/crates-io)!

whitequark:
Hm, I'm having trouble with that crates.io interface again. It is so horribly unusable.
![Screenshot_20210119_102705](https://user-images.githubusercontent.com/54771/105021796-e0272280-5a40-11eb-91b2-a01758723904.png)

stlankes:
I thought that `cargo owner --add github:smoltcp-rs:crates-io` is the way to do it.

whitequark:
That produces the exact same error.

stlankes:
Hm, does crates.io has oauth access to smoltcp-rs?

whitequark:
Thanks for the pointer. Done now ([smoltcp](https://crates.io/crates/smoltcp), [managed](https://crates.io/crates/managed)).

stlankes:
Thanks!

whitequark:
For the onlookers, what I had to do was to revoke permissions for crates.io for my account, then re-login while granting access to smoltcp-rs. I have no idea whether that's documented anywhere other than in https://github.com/rust-lang/cargo/issues/5297.

whitequark:
Actually, there's no release yet, I just set up the permissions, so reopening this.

Dirbaio:
smoltcp 0.7 is now out! :tada:

PR #413, opened 2021-01-19
Simplify more lifetimes

Dirbaio:
Follow-up to #410 

I'd like to get this in before the 0.7 release (#412) for consistency: if we do breaking changes simplifying some lifetimes we might as well simplify them all.


whitequark (approved):
Very nice!

PR #420, opened 2021-02-12
Make wire/ipv4 functions const where easy.

stabler:
This is a small PR that makes functions in `src/wire/ipv4.rs` const. Specifically targeting `Address::new` and `Cidr::new`, but also updating other functions where there is no impact to the function body.

`const fn` was [stabilized in 1.31](https://doc.rust-lang.org/edition-guide/rust-next/const-fn.html), so I believe this is zero impact to existing code (given current minimum required rust version of 1.40)

The motivation for this change is to simplify static initialization on embedded platforms. E.g. initialize an arbitrary IP address, or empty storage for a CIDR.

Let me know if this is misaligned with project goals/direction! It's useful to my projects, but I'm very open to feedback!

### DIFF
--- a/src/wire/ipv4.rs
+++ b/src/wire/ipv4.rs
@@ -39,7 +39,7 @@ impl Address {
     pub const MULTICAST_ALL_ROUTERS: Address = Address([224, 0, 0, 2]);
 
     /// Construct an IPv4 address from parts.
-    pub fn new(a0: u8, a1: u8, a2: u8, a3: u8) -> Address {
+    pub const fn new(a0: u8, a1: u8, a2: u8, a3: u8) -> Address {
         Address([a0, a1, a2, a3])
     }
 
@@ -54,7 +54,7 @@ impl Address {
     }
 
     /// Return an IPv4 address as a sequence of octets, in big-endian.
-    pub fn as_bytes(&self) -> &[u8] {
+    pub const fn as_bytes(&self) -> &[u8] {
         &self.0
     }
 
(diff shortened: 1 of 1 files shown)


src/wire/ipv4.rs
```diff
@@ -125,8 +125,11 @@ impl Cidr {
     ///
     /// # Panics
     /// This function panics if the prefix length is larger than 32.
-    pub fn new(address: Address, prefix_len: u8) -> Cidr {
-        assert!(prefix_len <= 32);
+    #[allow(clippy::no_effect)]
+    pub const fn new(address: Address, prefix_len: u8) -> Cidr {
+        // Replace with const panic (or assert) when stabilized
+        // see: https://github.com/rust-lang/rust/issues/51999
+        ["Prefix length should be <= 32"][(prefix_len > 32) as usize];
```
whitequark, line 132: This is a bit sketchy--is this a generally accepted workaround?

stabler: From what I've read, yes. There's a reference to this technique in [this comment](https://github.com/rust-lang/rust/issues/51999#issuecomment-687624116) on the issue ticket referenced (in relation to the RustCrypto crate).

It's also the method used by 3rd party crates [static_assertions](https://github.com/nvzqz/static-assertions-rs/blob/master/src/bool.rs#L57) and [const_fn_assert](https://github.com/powlpy/const_fn_assert/blob/master/src/lib.rs#L86). I considered bringing one of them in- static_assertions seems to be more popular- but for such a simple macro it didn't seem worth the overhead. Those crates use the simplified `[()][bool_expr as usize]` form, but I've seen people use `["debug message"][bool_expr as usize]` to provide better feedback if the assert is triggered.

whitequark (approved):

PR #424, opened 2021-02-23
Build fixes

astro:
Is there something wrong with my local rustc-1.49.0? How did these pass CI?

Still looking for more elegant ideas regarding the `wire::ethernet_header_len()` workaround.

### DIFF
--- a/src/iface/ethernet.rs
+++ b/src/iface/ethernet.rs
@@ -1720,7 +1720,7 @@ mod test {
 
     use super::{EthernetPacket, IpPacket};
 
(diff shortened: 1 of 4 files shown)


whitequark (changes requested):
Why do we even try to support TAP interfaces with Ethernet feature turned off? That makes no sense.

Dirbaio:
The ethernet issue is solved in #447. 

Why is the `'static` lifetime needed for that SocketSet?

Dirbaio:
I'm assuming the issues this PR fixes are no longer relevant. Please reopen if not.

PR #427, opened 2021-02-28
Fix BPF header length

### LINKED ISSUE #426: BPF header length incorrect on OpenBSD
The recent PR https://github.com/smoltcp-rs/smoltcp/pull/421 still does not support OpenBSD. I incorrectly read the length as decimal when it was hex.

It's currently written with a length of 18 bytes: https://github.com/smoltcp-rs/smoltcp/blob/master/src/phy/sys/bpf.rs#L19.
In my captures, there is an additional 8 bytes before the actual Ethernet frame.

The actual struct length is 24 bytes (192 bits): https://github.com/openbsd/src/blob/37ecb4d066e5566411cc16b362d3960c93b1d0be/sys/net/bpf.h#L144.
The missing 2 bytes (8 - (24-18)) are from alignment: https://github.com/openbsd/src/blob/37ecb4d066e5566411cc16b362d3960c93b1d0be/sys/net/bpf.c#L1649.

pk-j-xyz:
The actual header length may be larger than the `bpf_hdr` struct due to aligning: see https://github.com/openbsd/src/blob/37ecb4d066e5566411cc16b362d3960c93b1d0be/sys/net/bpf.c#L1649
and https://github.com/apple/darwin-xnu/blob/8f02f2a044b9bb1ad951987ef5bab20ec9486310/bsd/net/bpf.c#L3580.

This change does not affect `macos` (aligning 18 bytes + 14 bytes for the ethernet header is still 18 bytes on 32 and 64 bit machines, but does affect `openbsd` (aligning 24 bytes + 14 bytes for the ethernet header requires 26 for both 32 and 64 bit machines). I am ignoring other platforms, but can guard the test with a `cfg(target_pointer_width)` if you'd like. Keep in mind this only affects `macos` and `openbsd` until other `bpf` targets are added.

smoltcp currently only supports Ethernet, but this code will need to adapt if that changes: see use of the `ETHERNET_HEADER_LEN` constant.

Fixes https://github.com/smoltcp-rs/smoltcp/issues/426

### DIFF
--- a/src/phy/sys/bpf.rs
+++ b/src/phy/sys/bpf.rs
@@ -1,8 +1,10 @@
 use std::io;
+use std::mem;
 use std::os::unix::io::{AsRawFd, RawFd};
 
 use libc;
 
+use crate::wire::ETHERNET_HEADER_LEN;
 use super::{ifreq, ifreq_for};
 
 /// set interface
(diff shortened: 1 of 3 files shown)


whitequark:
Thanks!

PR #437, opened 2021-03-12
Adding DHCP lease management

### LINKED ISSUE #307: DHCP lease renewal time not respected
The DHCP client implementation ignores the renewal (T1) and rebinding (T2) times set by the server, and instead has a fixed renewal interval of 60s ([here](https://github.com/m-labs/smoltcp/blob/1ada3da5aa33917daf891dda2242d7f5beadfb10/src/dhcp/clientv4.rs#L16)).

Renewing earlier than T1 is explicitly allowed by the RFC, but renewing this frequently is a little extreme, and can set off intrusion detection systems (sample size 1...).

### LINKED ISSUE #436: Detecting DHCP lease renewal timeout
If the DHCP server cannot be contacted within the lease time, [RFC 2131](https://tools.ietf.org/html/rfc2131#page-22) states that the user of the address must immediately cease usage. Currently, there's no means of determining if a DHCP lease has expired, so it's impossible for clients to cease usage of an allocated address.

ryan-summers:
This PR fixes #436 by adding an API to check if the current DHCP lease has expired, which allows external users to deconfigure the IP address from the interface.

This fixes #307 by updating DHCP to utilize the lease duration provided by the DHCP server.

TODO:
- [x] Add tests for parsing DHCP lease duration
- [x] Test updates on hardware


src/dhcp/clientv4.rs
```diff
@@ -114,6 +117,25 @@ impl Client {
...
+    /// # Note
+    /// RC 2131 requires that a client immediately cease usage of an
+    /// address once the lease has expired.
+    ///
+    /// # Args
+    /// * `now` - The current network time instant.
+    ///
+    /// # Returns
+    /// True if a lease exists and it has expired. False otherwise.
+    pub fn lease_expired(&self, now: Instant) -> bool {
```
ryan-summers: Instead of having a manual API to check, we could have the client manually return a config with an UNSPECIFIED Ipv4 address so that the config is stuck. This is similar to what happens if the DHCP Retry mechanism is exceeded.

src/dhcp/clientv4.rs
```diff
@@ -199,12 +221,18 @@ impl Client {
...
         // once we receive the ack, we can pass the config to the user
         let config = if dhcp_repr.message_type == DhcpMessageType::Ack {
-               let address = dhcp_repr.subnet_mask
-                   .and_then(|mask| IpAddress::Ipv4(mask).to_prefix_len())
-                   .map(|prefix_len| Ipv4Cidr::new(dhcp_repr.your_ip, prefix_len));
-               let router = dhcp_repr.router;
-               let dns_servers = dhcp_repr.dns_servers
-                   .unwrap_or([None; 3]);
+            let lease_duration = dhcp_repr.lease_duration.unwrap_or(DEFAULT_RENEW_INTERVAL * 2);
+            self.lease_expiration = Some(now + Duration::from_secs(lease_duration.into()));
```
ryan-summers, line 204: It looks like the DHCP server isn't currently respecting this lease expiration - it will unconfigure the interface after X number of retries instead, so I need to update this.

src/dhcp/clientv4.rs
Dirbaio: IMO returning UNSPECIFIED is better than a manual `lease_expired` API. I'd say most users care mostly about whether DHCP is up or not (do we have a valid lease or not), and care less about the reason. Also, it's less work for the user.

ryan-summers:
CC @whitequark / @Dirbaio - This should now be ready for review when either of you get a chance. I've verified that it's functioning as intended on hardware by sniffing the DHCPACK and verifying that with a 10 minute lease time, the client properly re-requests after 5 minutes

Dirbaio:
Very nice! Is my understanding on what changed correct?

Previous behavior:
- Start renew 60 seconds after ACK
- retry every 60 seconds, give up after 3 retries

New behavior:
- Start renew lease_time/2 after ACK
- retry every 60 seconds, give up when lease expires

If that's the case, :+1:  from me.

There's one concern: routers typically set very high lease times, so a smoltcp device could now become offline for *hours* if a network changes its IP range or some other configuration. I'd like to see some option to set a "ceiling" on the renew/lease time, so users can choose to mitigate this. We can add this later, I'm fine with merging this PR without it.

ryan-summers:
> Very nice! Is my understanding on what changed correct?
> 
> Previous behavior:
> 
> * Start renew 60 seconds after ACK
> * retry every 60 seconds, give up after 3 retries
> 
> New behavior:
> 
> * Start renew lease_time/2 after ACK
> * retry every 60 seconds, give up when lease expires
> 
> If that's the case, 👍 from me.

This summary is correct. `lease_time / 2` is recommended by RFC 2131

> There's one concern: routers typically set very high lease times, so a smoltcp device could now become offline for _hours_ if a network changes its IP range or some other configuration. I'd like to see some option to set a "ceiling" on the renew/lease time, so users can choose to mitigate this. We can add this later, I'm fine with merging this PR without it.

As far as my reading of the RFC, even if the router changes the port range, the assigned IP address to the device is still respected for the entire lease duration, so the device isn't "off-line". The DHCP server will be checking to see if addresses are already assigned before handing them out even between configuration changes. As far as I understand it, the RFC essentially states that the provided IP + lease is guaranteed to be valid for the whole duration, regardless of server reboots/configurations (it explicitly states the server should store IP assignments in persistent storage)

Additionally, reconfiguration of DHCP servers is (as I understand it) very much not the norm, so I don't think we should be designing intentionally around that. Reconfiguration is common in development environments, but pretty rare in production imo.

ryan-summers:
@Dirbaio / @whitequark Is there any feedback that needs to be incorporated before this can be merged?

src/dhcp/clientv4.rs
```diff
@@ -357,6 +353,7 @@ impl Client {
         net_trace!("DHCP reset");
         self.state = ClientState::Discovering;
         self.next_egress = now;
+        self.lease_expiration = None;
```
ryan-summers: I believe that we actually shouldn't lose track of the lease expiration date through a reset (at least, there should be a way to return to discovering without losing the expiration date of our lease) because we are allowed to continue using the IP even if the ethernet link connection is lost.

Dirbaio (approved):
Looks great!

Regarding the DHCP server reconfiguration: In my company we've had problems where a customer changed their address ranges and our devices stopped working and had to be powercycled because the leases were hours long. 

A "max lease time" where it caps the lease time if the dhcp server's is longer would be handy. We can always add this later though.

jordens:
Not honoring the lease time and instead renewing significantly more frequently is somewhat problematic. It can reasonably be seen as non-compliant and causing unnecessary or unwanted traffic. To migrate DHCP configurations, you'd do something similar to DNS migration, decrease the lease time/TTL on the server, then migrate, then increase again.

Dirbaio:
I know, but our customers have no idea about that, are using crappy consumer-grade routers where these options may not be even configurable, and expect our devices to Just Work. :)

ryan-summers:
Yeah, but I would argue that should be baked into the application logic (e.g. you have a ping service going every minute) and not in the underlying DHCP implementation

Issue #439, opened 2021-03-15
Need a "packet queue" phy


gopakumarce:
Today smoltcp has full control over the "device" (phy) - it reads and writes from/to it and removes/adds encapsulations etc.. But in many scenarios the packet handling has to be a lil more flexibly done outside the control of smoltcp. So provide a phy which is just a simple Rx/Tx queue

PR #440, opened 2021-03-15
Re #439:

gopakumarce:
Today smoltcp has full control over the "device" (phy) - it reads and writes from/to it
and removes/adds encapsulations etc.. But in many scenarios the packet handling has to
be a lil more flexibly done outside the control of smoltcp. So provide a phy which is
just a simple Rx/Tx queue

For more details, please see examples/README.md section "smoltcp as a proxy"

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,156 @@
+# Usage notes
+
+smoltcp expects a packet path as below, just like any standard networking path.
+
+Rx Pkt--phy[l2 pkts]-->iface[l3 pkts]-->Rx tcp/udp socket[l4 data]-->application
+
+application-->[l4 data]smoltcp tcp/udp socket-->[l3 pkts]iface-->[l2 pkts]phy--> Tx Pkt
+
+There are some ready to use phy modules available (see src/phy) like ethernet raw socket,
+tun/tap interfaces, file descriptors, generic IP packet queues etc.. 
+
+## smoltcp as a server
+
+### TCP
+
+In the server.rs example, this is the general pattern followed
+
+  create-unix-tap-interface
+  phy = Fd (file descriptor of tap)
+  iface = create-iface(phy, <ip address of iface>, <mac address of phy>);
+  socket1 = create-tcp-socket(rx-buffer, tx-buffer)
+  socket.listen(6969) // the port on which we expect clients to connect to
+  socket_set = create-set(socket1, <more sockets if any ..>) // socket set will be explained down below
+
+  loop {
+    // The poll API does the below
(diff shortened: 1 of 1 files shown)


Dirbaio:
I'm not convinced this is something that should be in core smoltcp, because:

1. smoltcp provides the `Device` trait so users can implement their own, with any buffer management strategy they desire.  You could simply have this impl in your crate. There's no requirement that `Device` implementations be in the smoltcp crate, so having this in smoltcp doesn't unlock any new use cases (it doesn't allow doing something that was impossible before).
2. It is specific for a very particular use case: it requires alloc, it provides for a fixed-size lower-layer header which seems to be tracked _per packet_, queues can grow indefinitely.

smoltcp does provide tap/rawsocket Device impls because they're required for examples and benchmarks and are likely useful for a big fraction of the userbase, but this is not the case for `PacketQ`.

whitequark:
I agree with @Dirbaio.

gopakumarce:
Makes sense, I will just use the packetQ from my code base. 

Would you like the README.md (I can reword the proxy section mentioning how to add a device trait to do that etc..) - I think it will save a new user at least a day or two trying to investigate and re-discover the same things mentioned in the readme.

whitequark:
I think this would work better if you documented in your project how you're using smoltcp. Then we could link to that from smoltcp's README.

gopakumarce:
I have it already documented in my project :), but my project is not planned to be open source at the moment. No worries, we can discard this MR, at least the one thing if you could take from this readme and put it back in some readme somewhere in smoltcp is the below that I mentioned in the UDP section - could save a couple of hours of debugging for someone else

```
2. The Rx buffer for UDP is basically a ring buffer .. But if we call send_slice(), smoltcp will
   try to find a 'contiguous' are of the ring to enqueue the entire data passed in send_slice().
   Remember this is UDP and we cant sent partial data, so we need to send all or nothing, but the
   ring buffer contiguity enforcement means that if we allocate say 1500 bytes of buffer and then
   send 200 bytes of data, after the 200 bytes is transmitted the ring buffer advanced by 200 bytes,
   and now the only contiguous buffer available is 1300 bytes. So the next send_slice() call to
   send 1500 bytes will NEVER succeed how many ever times we retry. So for UDP, make sure we
   allocate two times the size of the largest data packet we plan to send, so we always have one
   max-data-packet worth of contiguous buffer always available.
```

gopakumarce:
And by the way I have to really thank smoltcp and the people behind it again - before I knew about smoltcp, I was initially thinking I will have to wrap lwip in rust and use that, that would have taken me a good couple of weeks worth of effort at the least, with smoltcp I had to spend maybe two or three days to get a hang of how to use it and then I dint face any further issues

PR #442, opened 2021-03-17
Update managed to the latest version

dutchmartin:
Makes it possible to use alloc on stable, see https://github.com/smoltcp-rs/rust-managed/pull/16.


Dirbaio:
Why didn't we do this earlier? hahaha

Thank you!

PR #447, opened 2021-03-23
Fix feature-related compilation issues.

### LINKED ISSUE #428: 0.7 build error
I've just updated to `0.7` but I cannot build anymore:

```
error[E0004]: non-exhaustive patterns: type `(&Cidr, &Cidr)` is non-empty
   --> /Users/jiayihu/.cargo/registry/src/github.com-1ecc6299db9ec823/smoltcp-0.7.0/src/wire/ip.rs:334:15
    |
334 |         match (self, subnet) {
    |               ^^^^^^^^^^^^^^
    |
    = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
    = note: the matched value is of type `(&Cidr, &Cidr)`
```

These are the enabled features:

```
smoltcp = { version = "0.7.0", default-features = false, features = ["socket-udp", "socket-icmp"]}
```

Rust version:

```
rustc 1.50.0-nightly (593fe977a 2020-11-20)
```

Dirbaio:
The goal is that every combination of features either:
- compiles cleanly, or
- fails to compile with a human-readable error message

I've opted to make "no ipv4 and no ipv6" an unsupported combination. It *could* be made to work but I don't think it's worth it. 

Fixed cases: 

- `(no features)` -> now fails with `You must enable at least one of the following features: proto-ipv4, proto-ipv6`. Fixes https://github.com/smoltcp-rs/smoltcp/issues/428#issue-821009561
- `proto-ipv4,socket` -> now fails with `If you enable the socket feature, you must enable at least one of the following features: socket-raw, socket-udp, socket-tcp, socket-icmp`. Fixes https://github.com/smoltcp-rs/smoltcp/issues/428#issuecomment-800541276
- `ethernet` -> now correctly enables `socket` feature, and then fails with the error above.
- `proto-dhcpv4` -> now correctly enables `ethernet` feature.
- `proto-ipv4,phy-tap_interface` -> now correctly enables `ethernet` feature.

fixes #428 

Will be cherry-picked for 0.7.1 release


crawford (approved):

whitequark:
> I've opted to make "no ipv4 and no ipv6" an unsupported combination. It _could_ be made to work but I don't think it's worth it.

I think this is a reasonable decision. That's what I aimed to do anyway, but I think at the time `compile_error!` was not a thing.

PR #448, opened 2021-03-24
Remove support table from docs. Fixes #361

### LINKED ISSUE #361: Document table broken at `docs.rs`
This is a minor document issue: https://docs.rs/smoltcp/0.6.0/smoltcp/#packet-and-representation-layer-support

图片

The markdown table here seems failed to render.

Dirbaio:
The table is no longer very informative since it's all "yes" now.


whitequark:
The table was there in case we add more protocols (and so it'll be extended with a few more rows), but I don't object.

PR #455, opened 2021-03-31
Add defmt logging support

### LINKED ISSUE #376: Support for defmt logging
https://github.com/knurling-rs/defmt is a new approach to tracing/logging from embedded devices which dramatically reduces bandwidth use by deferring formatting to the host, and using indexed format strings.

I think this is a very good fit for embedded smoltcp, since other than in low-traffic situations it is currently basically impossible to use smoltcp's (extensive) logging.  With defmt, this might become feasible.

Unfortunately, format strings are not compatible since for compactness, `defmt` format strings need to contain more information about the formatted type (e.g. `{:u8}`).

Since `defmt` isn't meant for non-embedded use, smoltcp would have to switch between them using a feature flag.  The easiest way I could see to support both syntaxes is to use a custom proc macro (`net_trace` and `net_debug` are already custom declarative macros anyway), which outputs two versions of the logging call depending on the feature flag selected (and none if logging is switched off).  However, it's also conceivable for `defmt` to implement that, so I opened https://github.com/knurling-rs/defmt/issues/149 there.

cc @thalesfragoso @adamgreig (stm32-eth developers)

Dirbaio:
This adds basic support for optionally logging with `defmt` instead of `log`, controlled via Cargo features.

Any struct that can implement `defmt::Format` now does so. There's a few left, due to:
- Missing impls for Cell, RefCell that should be added in `defmt`.
- Missing impls for `managed` types.

The left structs are not used in log statements anyway, and are usually "big" structs, like sockets. Since this is already fully functional for integration in a firmware that uses `defmt` I would merge this now, and add the missing impls later when they're unblocked.

Fixes #376

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ byteorder = { version = "1.0", default-features = false }
 log = { version = "0.4.4", default-features = false, optional = true }
 libc = { version = "0.2.18", optional = true }
 bitflags = { version = "1.0", default-features = false }
+defmt = { version = "0.2.0", optional = true }
 
 [dev-dependencies]
 env_logger = "0.5"
@@ -46,6 +47,13 @@ verbose = []
 "socket-tcp" = ["socket"]
 "socket-icmp" = ["socket"]
 "async" = []
+
+defmt-trace = []
+defmt-debug = []
+defmt-info = []
+defmt-warn = []
+defmt-error = []
+
 default = [
   "std", "log", # needed for `cargo test --no-default-features --features default` :/
   "medium-ethernet", "medium-ip",
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,7 +70,12 @@ jobs:
         features:
           # These feature sets cannot run tests, so we only check they build.
           - medium-ip medium-ethernet proto-ipv6 proto-ipv6 proto-igmp proto-dhcpv4 socket-raw socket-udp socket-tcp socket-icmp async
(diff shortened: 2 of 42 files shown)


Cargo.toml
```diff
@@ -46,6 +47,13 @@ verbose = []
 "socket-tcp" = ["socket"]
 "socket-icmp" = ["socket"]
 "async" = []
+
+defmt-trace = []
+defmt-debug = []
+defmt-info = []
+defmt-warn = []
+defmt-error = []
+
```
therealprof, line 56: Hm, what's the point of these, are they needed by `defmt`?

therealprof (approved):
Sweet! Any comparison available between typical binary size and `defmt`enabled?

Cargo.toml
Dirbaio: Yes, the defmt macros only emit code if the corresponding log level (or lower) is enabled. The end-user is supposed to depend on smoltcp with `defmt, defmt-debug` features to see up to `debug` level, for example. This is so log level is configurable per-crate.

therealprof: But don't you need to forward the features then? I'm confused about how this would work as-is in this PR.

Dirbaio:
`defmt` won't build with 1.40, which is our MSRV. The lowest I can get this to build is with 1.46, see https://github.com/knurling-rs/defmt/issues/443

I guess the best option is to declare our MSRV is 1.40 except if you're using defmt, in which case the limiting factor is defmt's MSRV.

Dirbaio:
> Any comparison available between typical binary size and defmtenabled?

Unfortunately no :( I've been using this internally for a while already and no longer have a firmware that uses `log` handy...

Issue #457, opened 2021-04-01
DHCP replies with ICMP Port Unreachable when RawSocket buffer full


Dirbaio:
Right before the ACK, a random UDP broadcast arrives which occupies the only slot in the RawSocket buffer. When the ACK arrives, the RawSocket is full, so it doesn't accept the packet. smoltcp then replies with a "port unreachable". This probably shouldn't happen.

![screenshot-2021-04-01_15-42-35](https://user-images.githubusercontent.com/1247578/113304348-3b5b7800-9302-11eb-84c4-c3ba239f0637.png)

(DHCP doesn't retry the request after that and the device goes offline, but I think that's another bug...)

ryan-summers:
I was observing this as well, thanks for figuring out the root cause here!

Dirbaio:
Fixed in #459

PR #459, opened 2021-04-01
DHCP refactor

### LINKED ISSUE #446: Convert DHCP client to a socket
Currently the DHCP client uses RawSocket to send/receive packets. The RawSocket needs TX and RX buffers, which need to be 900 + 600 bytes = 1.5KB.

If the DHCP client could send and receive packets "on the fly", inside the Interface poll loop, it would use 1.5KB less of RAM.

The way to do this is make the DHCP client be a new socket type.

Thoughts? @whitequark

### LINKED ISSUE #464: DHCP retransmits request too fast, breaks if server is slow.
The DHCP client retransmit REQUEST packets if no ACK arrives after 1 second. Retransmits are done with a new XID and ACKs with old XID are ignored.

Therefore, if the router takes more than 1s to reply to ACKs, smoltcp *never* acquires the IP address.

This particular router seems to delay for a bit intentionally to send ARP requests to check if the IP is in use (?). It's the Askey routers from Movistar's fiber (Spain's biggest ISP).

screenshot-2021-04-07_21-29-28

Dirbaio:
- DHCP is now a Socket, so it can process and emit packets on the fly instead of buffering them. Reduces RAM requirements by 1.5kb. Fixes #446, #457 
- No longer requires access to the Interface, it stores all state internally. Poll now has no params.
- Now handles DHCP NAK packets
- Now handles ACK packets only in states where we expect them.
- Renew retransmission algorithm correctly implements RFC2131: waits for `(expires_at-now)/2` before next retry.
- Retry REQUESTs slower and with exponential backoff. Fixes #464 
- Remove 0.0.0.0s from the DNS serevr list. (TP-link routers seem to pad the DNS server list with 0.0.0.0 to a fixed size.. wtf?)

Tested against these DHCP servers:

- mikrotik hAP AC2, RouterOS v6.45.9 (long-term)
- dnsmasq 2.84rc2 (linux router)
- Askey rtf8115vw (Movistar fiber, Spain)
- TP-Link TD-W8970
- Linksys E2000 with Tomato 1.28
- Pixel 4a Android 11 USB-Ethernet tethering
- Ubiquiti EdgeRouter X, many thanks to @adamgreig
- TP-Link Archer C7 v2 on OpenWrt 19.07.3, many thanks to @GrantM11235

Testing on more routers would be **highly appreciated**, especially crappy ISP routers. It requires a linux machine with wired Ethernet connection, and only 5 minutes of your time! Instructions [here](https://gist.github.com/Dirbaio/86bfa62873fc6375c914a5ede0a4698f)


Dirbaio:
Marking as draft until I remove the "fast renew" hack commit

src/wire/udp.rs
```diff
@@ -230,25 +231,26 @@ impl<'a> Repr<'a> {
         Ok(Repr {
             src_port: packet.src_port(),
             dst_port: packet.dst_port(),
-            payload:  packet.payload()
         })
     }
 
     /// Return the length of a packet that will be emitted from this high-level representation.
```
ryan-summers: Is this comment still accurate? I imagine a higher level representation would actually emit the payload as well?

src/wire/udp.rs
```diff
@@ -28,6 +28,8 @@ mod field {
     }
 }
 
+pub const HEADER_LEN: usize = field::CHECKSUM.end;
```
ryan-summers, line 31: Should we use this constant below instead of `field::CHECKSUM.end` for clarity?

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+const DISCOVER_TIMEOUT: Duration = Duration::from_secs(10);
+
+// timeout doubles every 2 tries.
+// total time 5 + 5 + 10 + 10 + 20 = 50s
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+const REQUEST_RETRIES: u16 = 5;
+
+const MIN_RENEW_TIMEOUT: Duration = Duration::from_secs(60);
+
+const DEFAULT_LEASE_DURATION: u32 = 120;
```
ryan-summers: What are the units on this? Should we specify this as a `Duration`? That works nicely above to show the units.

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// IP address that we're trying to request.
+    requested_ip: Ipv4Address,
+}
+
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+struct RenewState {
+    /// Server that gave us the lease
+    server: ServerInfo,
+    /// Active networkc config
```
ryan-summers: Typo: `network config`

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+    config: Config,
+
+    /// Renew timer. When reached, we will start attempting
+    /// to renew this lease with the DHCP server.
+    /// Must be less or equal than `expires_at`.
+    renew_at: Instant,
```
ryan-summers, line 84: Would it make more sense to encode this as a `Duration` before the expiration that we will renew at? That will enforce that the renewal time always occurs before expiration by design.

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// Requesting an address
+    Requesting(RequestState),
+    /// Having an address, refresh it periodically.
+    Renewing(RenewState),
+}
+
+/// Return value for the `Dhcpv4Socket::poll` function
+pub enum Event<'a> {
+    /// No change has occured to the configuration.
+    NoChange,
```
ryan-summers: Should we just change the return value to be `Option<Event<'a>>` and drop the `NoChange` variant (since it could be encoded as `None`)? It would make it clearer that there's no action that needs to be taken by the end user, and then they can use option chaining like below:

```
self.dhcp_client.poll().and_then(|event| handle_configuration(event))
```

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+            ClientState::Requesting(state) => state.retry_at,
+            ClientState::Renewing(state) => state.renew_at.min(state.expires_at),
+        };
+        PollAt::Time(t)
+    }
+
+    pub(crate) fn process(&mut self, now: Instant, ethernet_addr: EthernetAddress, ip_repr: &Ipv4Repr, repr: &UdpRepr, payload: &[u8]) -> Result<()> {
+        let src_ip = ip_repr.src_addr;
+
+        if repr.src_port != DHCP_SERVER_PORT || repr.dst_port != DHCP_CLIENT_PORT {
```
ryan-summers: Shouldn't we log an error if we get a request on the DHCP client port that isn't from the expected DHCP server port instead of silently ignoring it? It might indicate some improperly-configured DHCP server (although how that would occur, no idea).

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub(crate) fn poll_at(&self) -> PollAt {
+        let t = match &self.state {
+            ClientState::Discovering(state) => state.retry_at,
+            ClientState::Requesting(state) => state.retry_at,
+            ClientState::Renewing(state) => state.renew_at.min(state.expires_at),
+        };
+        PollAt::Time(t)
+    }
+
+    pub(crate) fn process(&mut self, now: Instant, ethernet_addr: EthernetAddress, ip_repr: &Ipv4Repr, repr: &UdpRepr, payload: &[u8]) -> Result<()> {
```
ryan-summers, line 161: I'm less familiar with all the way smoltcp does routing and error handling, but it seems to me like we should try to propagate strange DHCP packet reception up to the application layer. If we're getting invalid representations, that may be grounds for an application to want to reset (in case of internal buffer corruption. etc). What are your thoughts here?

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+                        self.config_changed = true;
+                        state.config = config;
+                    }
+                }
+            }
+            (ClientState::Renewing(_), DhcpMessageType::Nak) => {
+                self.reset();
+            }
+            _ => {
+                net_debug!("DHCP ignoring {:?}: unexpected in current state", dhcp_repr.message_type);
```
ryan-summers, line 238: Should we explicitly decline incoming offers when in the "BOUND" state? I presume silently ignoring them will also be sufficient in most practical use cases.

src/socket/dhcpv4.rs
```diff
@@ -0,0 +1,462 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    fn parse_ack(now: Instant, _ip_repr: &Ipv4Repr, dhcp_repr: &DhcpRepr, max_lease_duration: Option<Duration>) -> Option<(Config, Instant, Instant)> {
```
ryan-summers: Why take in the IP representation if we don't need it at all here?

ryan-summers (commented):
This is all looking much nicer - thanks for putting in the work to get this refactored.

I'm going to test integrating this into our hardware platform and verify that things function as intended shortly, and will let you know how things go. Other feedback is attached in-line below - most is general questions on design decisions. I don't believe I found anything problematic during review.

ryan-summers:
Follow-up, I have now tested this on [`Stabilizer`](https://github.com/quartiq/stabilizer) and monitored network traffic with Wireshark. I can confirm that the DHCP client renewed after 4 minutes (on an 8 minute lease) and the ICMP message is absent as well.

I also want to note that the buffer removal is also quite significant and this refactor makes DHCP _much_ easier to use. Thanks for the hard work here!

ryan-summers:
@Dirbaio Ping on this - is there any chance we can work towards getting this merged? This is a very nice improvement to DHCP.

src/socket/dhcpv4.rs
Dirbaio: Done!

src/wire/udp.rs
Dirbaio: Nice catch, fixed!

src/wire/udp.rs
Dirbaio: Done

src/socket/dhcpv4.rs
Dirbaio: Done!

src/socket/dhcpv4.rs
Dirbaio: I'm not sure on this. The renew check would then be something like  `now > expires_at - renew_at_before_expiration` instead of `now > renew_at`, which is less clear. I think the current way is more "symmetric": At "renew_at", renew, and at "expires_at", expire.

src/socket/dhcpv4.rs
Dirbaio: Makes a lot of sense, yup! Done.

src/socket/dhcpv4.rs
Dirbaio: Removed!

src/socket/dhcpv4.rs
Dirbaio: Yes, the way "malformed" and similar errors are currently handled is propagating them up to user code. It seems however that it's a bad idea, see #281 . 

I still have to take a deep look at it, but it's quite likely that I'll refactor error handling to *remove* propagating these up, as detailed in https://github.com/smoltcp-rs/smoltcp/issues/281#issuecomment-805445430 , so I'm actively avoiding propagating Malformed errors in new code.

src/socket/dhcpv4.rs
Dirbaio: This check is actually impossible to hit, because it's checked before here https://github.com/smoltcp-rs/smoltcp/blob/e82ef25e9108425ea8ea6ff9c4682187e9dbe421/src/iface/interface.rs#L1092

Not sure if it should be removed, or maybe turned into an `assert!`. Removing it feels dangerous.

src/socket/dhcpv4.rs
Dirbaio: It's somewhat tricky. It's not as simple as "reply with DECLINE here".

For example, an OFFER packet might be duplicated. On first one we'll change from Discovering to Requesting. On second one we'd reply with DECLINE, which might cause us to lose the lease that we indeed want.

I think it's safer to just ignore them. Less chances of getting code wrong and less chances to hit whacky buggy DHCP server behavior. :rofl:

Dirbaio:
@ryan-summers Thank you for the review! and sorry for the delay, I've had a really bad month. 

Addressed all the comments, could you take another look?

src/socket/dhcpv4.rs
ryan-summers: That's an interesting take indeed. I think I can get on-board with that approach, as long as the error is being logged somehow

src/socket/dhcpv4.rs
ryan-summers: That's a good counter point. This should be good as-is then. Thanks for the analysis

src/socket/dhcpv4.rs
ryan-summers: I like keeping it here, as if it got removed at the upper layer, we could get in trouble. Maybe make it a hard assert as you note instead so this gets caught quickly

src/socket/dhcpv4.rs
ryan-summers: This is largely just a style choice. I'm fine with either. Let's leave it as-is then, since we have testing on the current implementation.

ryan-summers:
No worries! I appreciate the great work on the refactor. I believe there's only one discussion still in place (re: DHCP source/dest ports), but other than that, the changes look good

src/socket/dhcpv4.rs
Dirbaio: Changed to hard assert.

Dirbaio:
Changed the port check to a hard assert, I believe this should be now good to go.

... except the CI now fails due to a defmt semver-breaking change :( https://github.com/knurling-rs/defmt/issues/494

PR #469, opened 2021-04-29
Add IEEE 802.15.4/6LoWPAN support

thvdveld:
This adds the IEEE 802.15.4 frame representation.
Still a work in progress and interested to know what you think about this.

I really would like to add 6LowPAN as well, however I'm not sure where to put this in smoltcp, since 6LowPAN is kind of weird.


Dirbaio:
Very interesting! ieee802154/6lowpan support would be definitely great to have in smoltcp, but it's not a small task :) 

I believe the 6lowpan frame format is quite different than standard ipv6, so it might be a separate wire Packet and emit/parse methods. Support could be added in `Interface`, with a new `Medium` variant.

There's also 6lowpan over BLE (bluetooth low energy), which would use 6lowpan but not ieee802154 framing, passing the 6lowpan frames to a separate BLE stack.

How "independent" is 6lowpan from ieee802154? It would be great if it could be implemented at a layer lower than `Interface` instead of inside `Interface` itself, but I'm fearing that may not be possible, like with Ethernet. Ethernet support is super interlinked with IP, since it has to do ARP and NDP and silence the sockets until neighbors are discovered, etc. I tried splitting it in #336 and the result wasn't pretty. Is there similar neighbor discovery / MAC addressing issues with 6lowpan over ieee802154?

thvdveld:
6lowpan needs to set the addresses in the ieee802154 header (ieee802154 doesn't really do a lot, it has a sequence number per frame, and the source/destination address/pan ID and some security feature).

The _problem_ with 6lowpan is that it is hugely interlinked with IP, doing IP layer compression and fragmentation, since ieee802154 only allows 127 octets per frame.  It also needs to perform the ARP and NDP, etc.

I'm not sure what you mean with your last question.

thvdveld:
Just added 6LoWPAN IPHC parsing in wire, as well as the 6LoWPAN NHC parsing for UDP.

thvdveld:
Most of the wire implementation is done (need some more work on resolving the Address and probably other stuff that will come up when implementing the higher layers).

Currently, only IEEE 802.15.4 addresses are used (and thus not the BLE addresses). 

I think I now understand your last question. I think 6LoWPAN also needs to handle the ARP and NDP.

thvdveld:
Hi @Dirbaio, I think the _wire_ part is mostly done (probably still needs some more tests etc.).
I started in `interface.rs` but then I was wondering what the best approach was.
Would you handle each LOWPAN packet with new functions or should we first convert the incoming LOWPAN packet to an IPv6 packet and then handle it with the already existing functions?
The response is then converted back to a LOWPAN packet before sending it to the lower layers.
Not sure what the best approach is here.

thibautvdv:
Also, is there a specific reason why rust-fmt is not used for formatting the code?

thvdveld:
The current state of this PR is that I can use UDP sockets over 6LoWPAN for receiving.
Sending is also possible although only when an address is manually added to the neighbour cache.

thvdveld:
OK, still need correct some tests :'-).
The next step is probably updating ICMP and NDP since 6LoWPAN also has it's [RFC](https://datatracker.ietf.org/doc/html/rfc6775) for that...

Dirbaio:
This is so cool! 

How do you test it out? Is there a way to create a "6lowpan tun/tap" to get Linux to talk to it?

thvdveld:
I'm testing it on real hardware. Our research group at the university uses Contiki on Zolertia RE-Motes to research protocols used in IEEE802.15.4. Now we are starting to test protocols written in Rust.

The tests I have are really not that great yet: using a Contiki implementation + using a Rust implementation + an IEEE802.15.4 sniffer...

Dirbaio:
Fun! It seems like it should be possible to get it going with fakelb (now deprecated? the replacement is ieee82154_hwsim?): https://www.netdevconf.org/2.1/slides/apr8/aring_virtual_mesh.pdf

thvdveld:
It seems so that ieee82154_hwsim with wpan-tools is the way to go for testing, however my knowledge about those tools (especially modprobe, ip, etc.) is non-existent.

Dirbaio:
I've got hwsim working! I've pushed it to your branch (hope you don't mind). The `sixlowpan` example has the commands to set it up. Once it's running it sends/receives packets to the Linux 6lowpan stack, so you can talk to it with regular Linux `ping` etc.

I've tried pinging it (with `ping fe80::180b:4242:4242:4242%lowpan0`) but it seems the ICMP checkum fails because IPHC doesn't decompress to the right dst addr (smoltcp says `fe80::00ff:fe00:ffff`, wireshark says `ff02::1:ff42:4242`. it's a broadcast Neighbor Solicitation)

thvdveld:
Thank you for the test!

I didn't test NDISC, since I manually added the neighbour address into the cache in my tests.
The reason why I didn't test it yet is because the Contiki stack uses the compressed version of ndisc (which I didn't implement yet for smoltcp). However, 6LoWPAN should be able to work with uncompressed formats of ndisc. I'll try to debug the example now.

thvdveld:
14d553e fixes the addressing issue. I'll now try to adapt ICMP for 6LoWPAN (uncompressed version).

thvdveld:
`ping` should now work. Next step is to polish the code a little since a lot is still spaghetti code.

thvdveld:
Rebased on master.

1 test is failing but I have not yet figured out what causes this.

thvdveld:
Apparently the UDP API has changed, so need to update the Sixlowpan part.

Cargo.toml
```diff
@@ -35,12 +35,16 @@ alloc = ["managed/alloc"]
 verbose = []
 "medium-ethernet" = ["socket"]
 "medium-ip" = ["socket"]
-"phy-raw_socket" = ["std", "libc", "medium-ethernet"]
+"medium-sixlowpan" = ["socket", "sixlowpan", "proto-sixlowpan"]
+"phy-raw_socket" = ["std", "libc"]
 "phy-tuntap_interface" = ["std", "libc", "medium-ethernet"]
+"ieee802154" = []
```
Dirbaio: features `ieee802154` and `sixlowpan` should probably be `proto-blahblah`

src/dhcp/clientv4.rs
```diff
@@ -0,0 +1,468 @@
+use super::{UDP_CLIENT_PORT, UDP_SERVER_PORT};
```
Dirbaio: This file accidentally slipped through a rebase :)

src/phy/mod.rs
```diff
@@ -275,15 +277,36 @@ pub enum Medium {
     /// Examples of devices of this type are the Linux `tun`, PPP interfaces, VPNs in tun (layer 3) mode.
     #[cfg(feature = "medium-ip")]
     Ip,
+
+    /// 6LoWPAN medium. Devices of this type send and receive IPv6 frames encoded using 6LoWPAN.
```
Dirbaio: As implemented, the "Sixlowpan" medium is really the "Ieee802154" medium, because the Device impl sends/receives full Ieee802154 packets.

I would rename it (and the Cargo feature) to ieee802154. We could add a "Sixlowpan" medium in the future (not needed in this PR :smile: ) that really sends/receives only 6lowpan packets, without the ieee802154 header.

This would be consistent with the IP/Ethernet mediums:

- Medium::Ethernet: Ethernet header + IP packet
- Medium::Ip: IP packet
- Medium::Ieee802154: Ieee802154 header + 6lowpan packet
- Medium::Sixlowpan: 6lowpan packet

(6lowpan over BLE would require Medium::Sixlowpan, it doesn't have Ieee802154 headers)

src/wire/ieee802154.rs
```diff
@@ -0,0 +1,931 @@
...
+
+    for b in buffer {
+        crc = crc_byte(crc, *b);
+    }
+
+    crc
+}
+
+#[derive(Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
+pub struct Ieee802154Conf {
```
Dirbaio: This struct is only used from Interface, perhaps it should be there instead of in `wire`. It looks like "internal runtime state", especially `sequence_no`.

src/wire/sixlowpan.rs
```diff
@@ -0,0 +1,1720 @@
...
+
+/// A wrapper around the address provided in the 6LoWPAN_IPHC header.
+/// This requires some context to convert it the an IPv6 address in some cases.
+/// For 802.15.4 the context are the short/extended addresses.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Address<'a> {
+    Complete(ipv6::Address),
+    WithContext(&'a [u8]),
+    Elided,
+    Reserved,
```
Dirbaio, line 27: Is this meant to represent "this bit combination is reserved, invalid for now"? If so it might make more sense to remove it and have src_addr/dst_addr return a Result, and represent this with Malformed.

In general it's nicer to represent "invalid" values with Result/Error, so the error is handled in the correct place, and no `unreachable!()` are required in places such as `resolve()`, which might cause crashes if someone forgets some check :)

Dirbaio (commented):
Thanks for your hard work! Just, wow, why is 6lowpan so complex with all the header compression and stuff? :rofl: 

I (finally, sorry for the slowness :sweat_smile:) got to do a first review. The structure and organization fits really well the existing code, so I see no major concerns there.

I've commented a few nits, mostly minor.

Something that needs fixing is the build fails when enabling `medium-sixlowpan` together with other mediums. This should be supported, to enable use cases like multi-medium devices (like an Ethernet + ieee802154 gateway). This is the reason for the `Medium` enum and all the `match`es on it.

src/phy/mod.rs
thvdveld: I renamed the feature to `medium-ieee802154` for now :smile:. I think that the features are more consistent now.

src/wire/sixlowpan.rs
thvdveld: The `WithContext` is defined in the [RFC](https://datatracker.ietf.org/doc/html/rfc6282#section-3.1.1), however I don't know how to really use it.
Instead of returning `Malformed` I added the `NotSupported` error. This way, users will know that the behaviour is defined in some standard or RFC, but that smoltcp just does not support it yet. I think this can reduce debug time for users, since we already confirm that there is well-defined behaviour, just not supported.

src/dhcp/clientv4.rs
thvdveld: Oops :p Removed this!

thvdveld:
> why is 6lowpan so complex with all the header compression and stuff?

I know right :rofl: . It's because IEEE802.15.4 as a MTU of 127 I think, and the MAC header of 802156 is already 25 bytes long, so the payload can only be 102 bytes... Then you have the IPv6 header size, which is minimum 40 bytes, so we have 62 bytes left. Finally, the UDP header is 8-bytes, so we would have only 54 bytes for a UDP payload... So I guess the compression needs to be very good :rofl:

thvdveld:
For some reason, the UDP checksum in `src/wire/sixlowpan.rs` is wrong, however, it is close to the expected value. Still trying to figure out how that's possible.

src/wire/ieee802154.rs
thvdveld: I removed it completely. Now the IEEE802.15.4 is stored in `hardware_addr` in the Interface. Not sure if an interface can have both an Ethernet address and a IEEE802.15.4 address, however, I can separate them in to to fields in `Interface` when needed. The source PAN ID and destination PAN ID are now stored in the context, since it is some kind of address I guess. But I'm not sure if it is the correct place to store them, since it is dynamic. Maybe it belongs in `Interface` as well.

thvdveld:
Now trying to get rid of all the `unreachable!()`, because it seems it _is_ possible to reach them :rofl:

oddstr13:
Can confirm that I managed to hit some unreachable code by sending a raw packet containing "Hello, World!" :laughing:

I've managed to get it (udp rx/tx) mostly working on the nRF52840 chip, with the exception of missing two characters on the rx side.
Wireshark looks happy with the on-air packets, so that's most likely an issue *somewhere* in the RX pipeline, between the radio receiving and the data getting printed to the terminal, but I haven't looked closer into why it happens yet.

image
image


My test code so far: https://github.com/oddstr13/usb-dongle-smoltcp-6lowpan-test (working, but probably not the best example of good Rust)

thvdveld:
@oddstr13 The two bytes that are missing are the FCS bytes from the IEEE802.15.4 layer. The cc2548 from TI calculates this in hardware and appends it to the frame. We should probably add this to the capabilities struct of the device such that we can calculate it in smoltcp when needed.

Thanks for the test code for the nRF!

oddstr13:
Ah, so you're assuming the received packet contains the crc, and stripping it from the end of the packet then?
I figured that was a likely cause, but I just haven't debugged it yet.

The nRF52 is actually handling the crc internally, so there's no need to generate it in rust just for the nRF52
https://github.com/oddstr13/nrf-hal/blob/148ab0f78f235bf65108d604bb792e87d26d50ed/nrf-hal-common/src/ieee802154.rs#L799-L809
The crc is available, but not usually included in the slice given by the Packet api
https://github.com/oddstr13/nrf-hal/blob/148ab0f78f235bf65108d604bb792e87d26d50ed/nrf-hal-common/src/ieee802154.rs#L894-L899

I am currently discarding packets that fail hardware crc validation
https://github.com/oddstr13/usb-dongle-smoltcp-6lowpan-test/blob/8ce7d313c6df09b31701082c8bccc6f074cf8396/ieee802154socket/src/lib.rs#L50

thibautvdv:
I am trying to test your implementation on the nRF, but I can't seem to build it: there is some lifetime issue I think and I didn't get enough coffee this morning to figure out the issue:
```
error[E0599]: the method `finalize` exists for struct `InterfaceBuilder<'_, IEEE802154Socket<'_>>`, but its trait bounds were not satisfied

   --> src/main.rs:132:25
    |
132 |     let mut iface = ifb.finalize();
    |                         ^^^^^^^^ method cannot be called on `InterfaceBuilder<'_, IEEE802154Socket<'_>>` due to unsatisfied trait bounds
    |
   ::: /tmp/usb-dongle-smoltcp-6lowpan-test/ieee802154socket/src/lib.rs:12:1
    |
12  | pub struct IEEE802154Socket<'a> {
    | ------------------------------- doesn't satisfy `IEEE802154Socket<'_>: smoltcp::phy::Device<'d>`
    |
    = note: the following trait bounds were not satisfied:
            `IEEE802154Socket<'_>: smoltcp::phy::Device<'d>`

```

thvdveld:
> Ah, so you're assuming the received packet contains the crc, and stripping it from the end of the packet then?
> I figured that was a likely cause, but I just haven't debugged it yet.

Indeed, I implemented it this way since the FCS is part of the IEEE802.15.4 frame format. However, currently I do not check if the FCS is correct.  When `payload()` is called on the IEEE802.15.4 frame, a buffer is returned with the payload of the MAC frame, so the implementation removes 2 bytes at the end from the frame. I think that the buffer passed to `smoltcp` should still contain the FCS field, since I think not all devices calculate the FCS field in hardware. For example, `example/sixlowpan.rs` uses a device that doesn't calculate the FCS.

oddstr13:
> I am trying to test your implementation on the nRF, but I can't seem to build it: there is some lifetime issue I think and I didn't get enough coffee this morning to figure out the issue:

Right you are – in my excitement of having gotten it working, I remembered to commit my changes, but forgot to push them last night – apologies :sweat_smile: 

I would have expected one of `smoltcp::Error::Trunkated` or `smoltcp::Error::Checksum` with part of the data missing, but I guess that isn't checked yet (not implemented in the UDP parser?).

thvdveld:
Yeah, I still had to add the checksum check for the UDP packet. I quickly add this in 1ff210c (not tested yet), so hopefully this works. I really need add tests in smoltcp for this :sweat_smile:

oddstr13:
I can confirm that the crc check is working, at least for truncated data.
Including the 802.15.4 crc field in the passed data does allow the udp crc check to pass.

oddstr13:
Am I supposed to handle Neighbor discovery in my application code?
Without manually filling the neighbor cache (and after the cache has timed out), I get this log message, and no packets get sent (stuck in tx buffer).
```
TRACE:smoltcp::socket::meta -- #0: neighbor fe80::180b:4242:dead:beef missing, silencing until t+3.000s
```

thvdveld:
Not sure how NDISC is handled in smoltcp. I know smoltcp responds to it, but not sure it is automated in smoltcp. @Dirbaio probably knows more about this. In my tests I added the neighbours in the neighbor cache.

The next thing to add to smoltcp after 6LoWPAN is probably RPL (Routing Protocol for Low power networks).

Dirbaio:
It [does](https://github.com/thibautvdv/smoltcp/blob/1ff210c9037e3b377a56b674969c42581978f9a7/src/iface/interface.rs#L2231) send NDISC requests for Ethernet mediums.

The issue is the 802154 code is [looking it up in `neighbor_cache`](https://github.com/thibautvdv/smoltcp/blob/1ff210c9037e3b377a56b674969c42581978f9a7/src/iface/interface.rs#L2325) and doing nothing if it's not present.

oddstr13:
I found a relevant RFC;
[RFC 6775 – Neighbor Discovery Optimization for IPv6 over Low-Power Wireless Personal Area Networks (6LoWPANs)](https://datatracker.ietf.org/doc/html/rfc6775)

thvdveld:
@oddstr13 You are right, that's the correct RFC for NDISC in 6LoWPAN. However, this is not yet implemented in this branch. I am using Contiki to test this branch. However, I don't think they implemented the RFC for NDISC; they use RPL instead. So I'll need to see what RIOT OS is doing before implementing this RFC.

I'm not at all an expert in this field, but I think NDISC is handled differently in 6LoWPAN. It is used differently here, because neighbours can be in a sleep state for example. Therefore, a 6LoWPAN router (6LR) is used, which periodically sends Router Advertisements. A 6LoWPAN host will ask the 6LR for the link layer address for a specific IP address by using a NS and NA message. That's why it's not implemented yet. However, we could do it the Ethernet way for the moment, but this is (I think) totally against the RFC, which says the following:

> A host MUST NOT multicast an NS message.

oddstr13:
I think enabling ethernet NDISC is fine for now, and a reasonable stopgap until RFC6775 can be implemented fully.

[6.5.5.  Address Resolution between Routers](https://datatracker.ietf.org/doc/html/rfc6775#section-6.5.5)

>    There needs to be a mechanism somewhere for the routers to discover
   each other's link-layer addresses.  If the routing protocol used
   between the routers provides this, then there is no need for the
   routers to use the ARO between each other.  Otherwise, the routers
   SHOULD use the ARO.  When routers use the ARO to register with each
   other and multihop DAD (Section 8.2) is in use, then care must be
   taken to ensure that there isn't a flood of ARO-carrying messages
   sent to the 6LBR as each router hears an ARO from their neighboring
   routers.  The details for this scenario are out of scope of this
   document.
>
>    **_Routers MAY also use multicast NSs as in [RFC4861]_** to resolve each
   others link-layer addresses.  Thus, routers MAY multicast NSs for
   other routers, for example, as a result of receiving some routing
   protocol update.  Routers MUST respond to multicast NSs.  This
   implies that routers MUST join the solicited-node multicast addresses
   as specified in [RFC4861].

I have not read the whole document yet (gotten to section 4), but I did get the impression that RFC4861, while not optimal due to potentially sleeping hosts and multicast, *is* allowed in parallel with RFC6775.

oddstr13:
That being said, RPL would also be nice to have, as I intend to eventually deploy a mesh (for similar use case as ZigBee).
[RFC 6550 – RPL: IPv6 Routing Protocol for Low-Power and Lossy Networks](https://datatracker.ietf.org/doc/html/rfc6550)

But, both may be considered out of scope of this PR, as they'd probably take quite a bit of work to implement properly?

thvdveld:
Yeah, I guess it's fine to have RFC4861, I'll do the necessary tomorrow or Friday, don't think it's a lot of work.

RPL (RFC6550) has a higher priority than RFC6775, since my research group is working with it. So I'm planning to implement it after this PR in a separate one, since the current one is already quite big (and also not finished yet). I think RPL is as big and even a bigger RFC to implement than 6LoWPAN.

oddstr13:
That sounds good to me!

I may not be of much use for actually implementing RPL in rust, but I do have a few nRF52840 based boards, and a desire to get a mesh up and running, so I could get out the duct-tape and poke around & test a bit when you get that far.

If you've got a known good hardware+software stack to test 6LoWPAN and RPL against, I'd be interested in the details of that, especially if said hardware is <$100.

thvdveld:
@oddstr13 Our research group is using [Contiki-NG](https://github.com/contiki-ng/contiki-ng) with the [Zolertia RE-Mote](https://github.com/Zolertia/Resources/wiki/RE-Mote). However, I don't recommend the Zolertia. I think the nRF52840 is supported in Contiki (I think @Dirbaio tested it).

Another OS that you could use is [RIOT OS](https://api.riot-os.org/group__net__gnrc__sixlowpan.html), but I don't have experience with it.

thvdveld:
Those features are getting out of hand :sweat_smile:

Dirbaio:
@thibautvdv I've squashed, rebased and added some fixes. I think this is almost ready to merge! It works great when tested with `mac802154_hwsim` (haven't had time to test with Contiki). Ping works, UDP works, NDISC works in both directions.

The issue with the ndisc SourceLinkLayerAddr/TargetLinkLayerAddr was due to Linux brokenness: Using a raw socket on `monitor0` causes weird results: packets we receive include FCS, packets we send are parsed as if they didn't have FCS, except by wireshark which always expects a FCS?? Turns out the sane way is to use raw sockets on normal `wpanX` interfaces, in which case all packets we send/receive are without FCS.

Due to that, and for consistency with Ethernt, I've changed the `wire` module to expect/emit packets without FCS too.

I'd like to merge this soon, as I want to do some refactors on Interface regarding error handling, and it would cause giant conflicts. If it works it can be merged, we can always improve/fix later.

I have a few outstanding questions, not sure if you have any before merging?

- Linux allows you to set one pan_id per interface, while we allow setting src_pan_id, dst_pan_id. What's the use case for allowing different src/dst IDs? Or should we just allow setting a single pan_id?
- Should we filter incoming packets with the wrong pan_id? Linux does it.

thvdveld:
> Linux allows you to set one pan_id per interface, while we allow setting src_pan_id, dst_pan_id. What's the use case for allowing different src/dst IDs? Or should we just allow setting a single pan_id?

It's defined this way in the specs of 802.15.4, however, it is never used together with 6LoWPAN (I think), so I guess we can just accept a single `pan_id`.

> Should we filter incoming packets with the wrong pan_id? Linux does it.

The spec specifies that a device must be able to filter frames when requested, so every transceiver should be able to filter the wrong pan_id's out. But I guess it doesn't harm anyone when we do it in smoltcp as well.

oddstr13:
Updated my test app, ndisc is working now :smiley: (once I figured out my address derp that is :sweat_smile:)

image

src/iface/interface.rs
```diff
@@ -989,6 +1101,103 @@ impl<'a> InterfaceInner<'a> {
...
+
+        let payload = iphc_packet.payload();
+        let ip_repr = IpRepr::Sixlowpan(iphc_repr);
+
+        // Currently we assume the next header is a UDP, so we mark all the rest with todo.
+        match iphc_repr.next_header {
+            SixlowpanNextHeader::Compressed => {
+                match SixlowpanNhcPacket::dispatch(payload)? {
+                    SixlowpanNhcPacket::ExtensionHeader(ext_header) => {
+                        todo!("{:?}", ext_header)
```
Dirbaio: `todo!` makes smoltcp crash. Could you log some "this is unsupported" message and then ignore the packet instead?

src/iface/interface.rs
```diff
@@ -989,6 +1101,103 @@ impl<'a> InterfaceInner<'a> {
...
+                            {
+                                Ok(()) => return Ok(None),
+                                Err(e) => return Err(e),
+                            }
+                        }
+
+                        // The packet wasn't handled by a socket, send an ICMP port unreachable packet.
+                        match ip_repr {
+                            #[cfg(feature = "proto-ipv6")]
+                            IpRepr::Ipv6(ipv6_repr) => {
```
Dirbaio: ip_repr is always Sixlowpan, so this seems unreachable

Dirbaio:
@oddstr13  I got your firmware working against Linux native stack with [this dongle](http://shop.sysmocom.de/products/atusb). It just worked! :)

src/iface/interface.rs
```diff
@@ -989,6 +1101,103 @@ impl<'a> InterfaceInner<'a> {
...
+                                let icmpv6_reply_repr = Icmpv6Repr::DstUnreachable {
+                                    reason: Icmpv6DstUnreachable::PortUnreachable,
+                                    header: ipv6_repr,
+                                    data: &payload[0..payload_len],
+                                };
+                                Ok(self.icmpv6_reply(ipv6_repr, icmpv6_reply_repr))
+                            }
+                            IpRepr::Unspecified { .. } => Err(Error::Unaddressable),
+                            IpRepr::Sixlowpan(_) => Err(Error::Malformed), // XXX(thvdveld): this is just wrong;
+                            r => todo!("{:?}", r),
```
Dirbaio: same as above

src/iface/interface.rs
```diff
@@ -989,6 +1101,103 @@ impl<'a> InterfaceInner<'a> {
...
+                        }
+                    }
+                }
+            }
+            SixlowpanNextHeader::Uncompressed(nxt_hdr) => match nxt_hdr {
+                IpProtocol::Icmpv6 => {
+                    self.process_icmpv6(cx, sockets, ip_repr, iphc_packet.payload())
+                }
+                hdr => todo!("{:?}", hdr),
```
Dirbaio: same as above

src/wire/ip.rs
```diff
@@ -500,6 +502,8 @@ pub enum Repr {
     Ipv4(Ipv4Repr),
     #[cfg(feature = "proto-ipv6")]
     Ipv6(Ipv6Repr),
+    #[cfg(feature = "proto-sixlowpan")]
+    Sixlowpan(SixlowpanIphcRepr),
```
Dirbaio: Adding `Sixlowpan` to `IpRepr` feels a bit strange to me. 6lowpan packets are just ipv6, but with a different on-wire format. It seems it should be possible to convert 6lowpan packets to/from the normal `IpRepr::Ipv6` when (de)compressing, so code that doesn't deal with 6lowpan doesn't have to care whether it's on 6lowpan or normal ipv6.

src/iface/interface.rs
```diff
@@ -229,32 +272,49 @@ let iface = InterfaceBuilder::new(device)
...
+                #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
                 neighbor_cache,
                 #[cfg(feature = "proto-igmp")]
                 ipv4_multicast_groups: self.ipv4_multicast_groups,
                 #[cfg(feature = "proto-igmp")]
                 igmp_report_state: IgmpReportState::Inactive,
+                #[cfg(feature = "medium-ieee802154")]
+                sequence_no: self.sequence_no,
+                #[cfg(feature = "medium-ieee802154")]
+                src_pan_id: self.src_pan_id,
```
Dirbaio: As discussed above, it'd be nice to have just a single `pan_id` instead of src/dst

Dirbaio (changes requested):
@thibautvdv I've been playing with the code, to learn more how it works, so here's the final "in depth" review :) 

The only "design issue" I see is adding the `IpRepr::Sixlowpan` variant, everything else is minor stuff.

thvdveld:
Still need to remove some other `todo!`s, but those are part of the `IpRepr::Sixlowpan`.
I'll first need to think about how to remove this representation.
I'll try to do it today, such that it can be merged ASAP (I promised to test the implementation, but didn't have the time to do it because of teaching).

src/wire/ip.rs
thvdveld: Voila, removed `IpRepr::Sixlowpan`!

oddstr13:
Did a quick test with 59e8604, looks like ndisc requests are no longer processed?
image
(Note `-- Started…` which indicates when I reset the board)

image

I have not done any further debugging yet.

---

**Edit:** Also panicked after receiving traffic from what's presumably some of my ZigBee devices;
image
image
https://github.com/smoltcp-rs/smoltcp/blob/59e8604e55fa77d82a363665d61d21995fa8583a/src/wire/ieee802154.rs#L362-L388

thvdveld:
@oddstr13 , my bad, didn't test my changes, shouldn't have pushed them. This week and next week is a busy week because of teaching, but I'll try to test it when I have some free time.

oddstr13:
@thibautvdv no worries, I can just figure out how to specify a specific git commit if I wanna play around, I ran my test app now specifically to test whether your PR was still working as expected or not :smiley:

thvdveld:
@oddstr13 I tested the implementation on my device, and it is responding to NDISC requests. Also tested it on Linux, which was also working. So not sure why it is not working for you :thinking: 

The panicking part is because the received data frame was a Beacon. For now we will just ignore frames that are not of the Data type. So we also don't respond to Acks (not sure if that is needed). I guess packets other than Data types are for lower layers that are not handled by smoltcp.

src/iface/interface.rs
```diff
@@ -1111,6 +1096,12 @@ impl<'a> InterfaceInner<'a> {
         let ieee802154_frame = Ieee802154Frame::new_checked(sixlowpan_payload)?;
         let ieee802154_repr = Ieee802154Repr::parse(&ieee802154_frame)?;
 
+        if ieee802154_repr.dst_pan_id != cx.pan_id {
```
oddstr13: This check is what stopped my code from working; I was not setting the PAN ID, so it defaults to `0x0000`.

This if is currently comparing `0x0000` with `None`, which presumably is not equal, and thus dropped.

Also not taking into consideration the `0xffff` broadcast PAN ID, which I don't know if is relevant to 6lowpan, but if it is, then the code should probably be capable of setting source and dest PAN ID separately, so it can broadcast and identify the source at the same time.
More research is needed about this.

thvdveld: After some searching and looking for information in the IEEE802.15.4 standard it seems that a PAN id is required (so it cannot be `None` I guess). However, the PAN ID that is used by a network is selected by a PAN ID coordinator (a device that will assign a PAN ID and that handles PAN ID collisions etc.). This is not implemented here in `smoltcp` and I don't think it should be part of it. That should be some physical/mac crate for IEEE802.15.4.

I looked in the 6LoWPAN RFC but didn't find anything about PAN id's. So do we actually need to filter on the PAN id? It is not specified in the RFC. @Dirbaio looked in the Linux kernel and there frames are filtered, but I think those are filtered in `net/mac802154/rx.c` and thus not in the 6LoWPAN implementation of Linux (or I didn't look closely enough).

I think we shouldn't filter them? What do you guys think? We could write a note in the README whenever we merge this into master to tell them that we accept all PAN id's (however not all hardware IDs).

For now I pushed a commit where we filter on our own PAN id AND broadcast.

Dirbaio: I'm pretty sure the right behavior is to filter on pan_id. This is what Linux does. It makes sense, since it allows having multiple independent PANs where a broadcast on one isn't received on the others for example, just like independent WiFi networks sharing the air.

Dirbaio (approved):
@thibautvdv I'm going to merge this. It's complete, works great, code is well structured! :+1: The few remaining TODOs don't seem very critical, they can be addressed in future PRs.

I want to do some big changes to Interface code, which would send this to conflict hell otherwise.

PR #482, opened 2021-05-31
Adding UDP socket close functionality

### LINKED ISSUE #475: Udp rebinding / closing
It doesn't appear that smoltcp provides a mechanism to close or un-bind a UDP port (unless I'm missing something?).

As a result of this, once a UDP port has been bound, it may not be used for any other activities.

It would be beneficial to allow UDP ports to be re-bound to different sockets

ryan-summers:
This PR fixes #475 by adding a `close()` method to UDP sockets to unbind from the current endpoint.

### DIFF
--- a/src/socket/udp.rs
+++ b/src/socket/udp.rs
@@ -148,7 +148,18 @@ impl<'a> UdpSocket<'a> {
 
     /// Close the socket.
     pub fn close(&mut self) {
+        // Clear the bound endpoint of the socket.
         self.endpoint = IpEndpoint::default();
+
+        // Reset the RX and TX buffers of the socket.
+        self.tx_buffer.reset();
+        self.rx_buffer.reset();
+
+        #[cfg(feature = "async")]
+        {
+            self.rx_waker.wake();
+            self.tx_waker.wake();
+        }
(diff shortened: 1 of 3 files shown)


src/socket/udp.rs
```diff
@@ -146,6 +146,11 @@ impl<'a> UdpSocket<'a> {
         Ok(())
     }
 
+    /// Close the socket.
+    pub fn close(&mut self) {
+        self.endpoint = IpEndpoint::default();
```
Dirbaio, line 152: This should wake rx_waker, tx_waker.

Dirbaio (changes requested):
Sounds reasonable!

What happens with the packets in rx_buffer/tx_buffer when the socket is closed and reopened (possibly with a different endpoint)? Should they be cleared?

ryan-summers:
I've added a new `PacketBuffer::clear()` API to allow us to flush the RX/TX buffer of the socket when closing it. Good catch there.

Dirbaio (approved):
Very nice! Thanks :)

Dirbaio (approved):

PR #485, opened 2021-06-08
impl std::error::Error for smoltcp::Error

### LINKED ISSUE #484: Why is `std::error::Error` not implemented for `smoltcp::Error`?
I'm trying to use `anyhow` to catch all errors in a simple example. However, 
```rust
error[E0277]: the trait bound `smoltcp::Error: std::error::Error` is not satisfied
   --> devourer/examples/dhcp_client.rs:110:41
    |
110 |         let x = tcp.recv_slice(&mut buf)?;
    |                                         ^ the trait `std::error::Error` is not implemented for `smoltcp::Error`
    |
    = note: required because of the requirements on the impl of `From<smoltcp::Error>` for `anyhow::Error`
    = note: required because of the requirements on the impl of `FromResidual<Result<Infallible, smoltcp::Error>>` for `Result<(), anyhow::Error>`
    = note: required by `from_residual`

```
Is it because `smoltcp::Error` is oversimplied and discouraged to expose to external libraries? However, I see that smoltcp::Error is already `Display+Debug`, which `std::error::Error` only requires

JakkuSakura:
closes #484

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,6 +160,9 @@ pub enum Error {
     Dropped,
 }
 
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
+
 /// The result type for the networking stack.
 pub type Result<T> = core::result::Result<T, Error>;
 


Dirbaio (approved):
Thank you!

PR #494, opened 2021-06-14
Adding support for retrieving socket handle from DHCP sockets

ryan-summers:
This PR adds support for getting the socket handle from a DHCP socket similar to the support provided by other socket types (e.g. UDP and TCP)


Dirbaio (approved):
Nice catch, thank you!

PR #495, opened 2021-06-14
Bumping smoltcp version to 0.8

ryan-summers:
This PR bumps the stored version in Cargo.toml to 0.8 so that any cargo-patch will utilize the `master` branch of smoltcp instead of published 0.7.4 versions. The `master` branch is incompatible with existing 0.7.x code already, so this will force dependency updates as well.


Dirbaio (approved):
:rocket:

PR #496, opened 2021-06-14
Update loopback.rs

gz:
Fixes a compiler error which imports `Vec` and `VecDeque` twice if for example we have an app that has smoltcp as a dependency and links it with feature `std`, and then also links with a no-std library that uses `smoltcp` with the `alloc` feature. 

The fix is to import `Vec` and `VecDeque` using the `alloc::*` path which works in both (`std` and `alloc`) configurations.

The easiest way to reproduce the error is to do a
`cargo build --features alloc`
(compiles smoltcp with alloc and std features)


Dirbaio (approved):
Makes sense, thank you!

PR #513, opened 2021-07-01
pcap timestamp bugfix

### LINKED ISSUE #512: Incorrect time format for pcap leads to wierd wireshark graph
This is a tcptrace graph(receiver side) produced by wireshark. Why does it always have a 1-sec gap? 
image

An linux socket is like this
image

JakkuSakura:
fixes #512

### DIFF
--- a/src/phy/pcap_writer.rs
+++ b/src/phy/pcap_writer.rs
@@ -73,7 +73,7 @@ pub trait PcapSink {
         assert!(length <= 65535);
 
         self.write_u32(timestamp.secs() as u32); // timestamp seconds
-        self.write_u32(timestamp.millis() as u32); // timestamp microseconds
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Nice catch! Thank you

JakkuSakura:
Consider use `microsecond` as internal unit?

Dirbaio:
~Instant is in ms because with us it'd overflow after aprox 70 minutes, which might be too short for some use cases. Sometimes you want TCP timeouts in the order of hours.~

Edit: they're in u64 so no, they wouldn't overflow that fast. Maybe!

JakkuSakura:
> Edit: they're in u64 so no, they wouldn't overflow that fast. Maybe!

I'm using nanoseconds in i64 in my program, and it will overflow after 292 years since 1970. us would give you 292471 years.

I would like to file a PR on it.

PR #514, opened 2021-07-04
use micros in Instant and Duration

JakkuSakura:
This improves precision everywhere, especially pcap
https://github.com/smoltcp-rs/smoltcp/pull/513


Dirbaio (approved):
Thank you! Looks great.

I'll probably write a follow-up PR to make the tick rate configurable. I have devices where the time source is 32768hz, having to convert to/from millis/micros is annoying.

bors r+

PR #521, opened 2021-08-19
Remove trailing commas in macros

thvdveld:
This is going to become a hard error in future releases of the compiler (https://github.com/rust-lang/rust/issues/79813).


Dirbaio (approved):
Nice catch! Thank you

PR #524, opened 2021-08-24
Expose underlying device from PcapWriter

bdbai:
`PcapWriter` should expose an interface for access to the lower `Device`, similar to `Tracer`.


Dirbaio (commented):
Makes sense, thank you! :)

Dirbaio (approved):

PR #526, opened 2021-09-09
Update MSV of Rust

thvdveld:
Force the MSV of Rust as the toolchain to use. This reduces the chance of using new features, such as or-pattern syntax (which is a really nice feature to use).

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.49.0
+          toolchain: 1.53.0
           override: true
           components: clippy
       - uses: actions-rs/clippy-check@v1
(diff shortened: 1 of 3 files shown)


Dirbaio:
Hmmm. This is nice to have, but I'm not sure if there are negative side effects of this. Using an older Rust for local dev can negatively impact the dev UX. For example it won't catch lints added in newer rusts. CI tests with a newer clippy, so it could be possible you don't see some clippy lints locally that later pop up in CI.

On the other hand, we could bump the MSRV. 1.46 is ancient. I have some refactors on my mind that would benefit from using `heapless` with const generics (Rust 1.51). Or-patterns are added in Rust 1.53. We can bump the MSRV to that.

thvdveld:
> Using an older Rust for local dev can negatively impact the dev UX. For example it won't catch lints added in newer rusts. 

Yeah, I didn't think about that, good point!

I think it's a good idea to update to Rust 1.53, just because it's nice to have the new features!

Dirbaio:
Let's do that then :rocket:

thvdveld:
I'll do the necessary if you want!

Dirbaio (approved):
Nice! Thank you

PR #528, opened 2021-09-13
Fix clippy because of MSV change

thvdveld:
This fixes the clippy warnings from the MSV changes in #526. Should have added this to #526.


thibautvdv:
Eeuuhm, not sure why CI is failing here (only in nightly) :open_mouth: @Dirbaio any idea here?

thibautvdv:
It has something to do with the features not set correctly for `ethernet_address` I think, but it's strange that the stable compiler version is not throwing an error about this.

Dirbaio (approved):
Thank you!

bors r+

Dirbaio:
goddamnit

bors r+

Dirbaio:
`bors` enforces master is always green: it merges on a staging branch, runs tests on the merge result, and only then merges the PR. That'll avoid what happened with the MSRV bump, which somehow??? made master red even though the PR was green...

thvdveld:
CI was probably using the old configuration... Once merged into master, CI is fired again, because of a new commit, and there the updated config was used, resulting in red probably. Honestly, this should not happen...

Dirbaio:
bors try

PR #540, opened 2021-10-05
make bors checks much simpler

Emilgardis:

PR #541, opened 2021-10-05
Bors test

Dirbaio:
Testing how #540 reacts to a test failing


Dirbaio (commented):
bors r+

Dirbaio:
indeed it works! I was somewhat scared github wouldn't report a status for `tests` instead of reporting failure, but seems like that's not the case

PR #547, opened 2021-10-11
Basic rand infrastructure.

Dirbaio:
See [previous discussion](https://github.com/smoltcp-rs/smoltcp/pull/465#pullrequestreview-774487285). Opening a separate PR so it can be discussed separately.

- Add `smoltcp::rand`.
  - On `std` targets, `OsRng` is used by default.
  - The user can supply a custom impl by enabling the `rand-custom-impl` Cargo feature and using the `rand_custom_impl!()` macro.
  - Specifying a custom impl is mandatory when `std` is not enabled.
- Make TCP initial sequence numbers actually random.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -1,0 +1,67 @@
+#![allow(unsafe_code)]
+#![allow(unused)]
+
+#[cfg(not(any(test, feature = "std", feature = "rand-custom-impl")))]
+compile_error!("None of the Cargo features `std` or `rand-custom-impl` is enabled. smoltcp needs a `rand` implementation to work. If your target supports `std`, enable the `std` feature to use the OS's RNG. Otherwise, you must enable the `rand-custom-impl` Cargo feature, and supply your own custom implementation using the `smoltcp::rand_custom_impl!()` macro");
+
+pub fn rand_u32() -> u32 {
+    let mut val = [0; 4];
+    rand_bytes(&mut val);
+    u32::from_ne_bytes(val)
+}
+
+/// Fill `buf` with random bytes.
+pub fn rand_bytes(buf: &mut [u8]) {
+    extern "Rust" {
+        fn _smoltcp_rand(buf: &mut [u8]);
+    }
+
+    unsafe { _smoltcp_rand(buf) }
+}
+
+/// Methods required for a custom rand implementation.
+///
+/// This trait is not intended to be used directly, just to supply a custom rand implementation to smoltcp.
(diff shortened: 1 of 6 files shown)


src/rand.rs
```diff
@@ -0,0 +1,67 @@
+#![allow(unsafe_code)]
+#![allow(unused)]
+
+#[cfg(not(any(test, feature = "std", feature = "rand-custom-impl")))]
+compile_error!("None of the Cargo features `std` or `rand-custom-impl` is enabled. smoltcp needs a `rand` implementation to work. stdCritical section is not implemented for this target. Make sure you've specified the correct --target. You may need to supply a custom critical section implementation with the `custom-impl` feature");
```
dnadlinger: Presumably the references to stdCritical here are just copypasta?

Dirbaio: Oops my bad. Fixed

src/socket/tcp.rs
```diff
@@ -751,18 +746,24 @@ impl<'a> TcpSocket<'a> {
...
+
+        let seq = Self::random_seq_no();
+        self.local_seq_no = seq;
+        self.remote_last_seq = seq;
         Ok(())
     }
 
+    fn random_seq_no() -> TcpSeqNumber {
+        #[cfg(test)]
+        return TcpSeqNumber(10000);
```
ryan-summers, line 762: Is the intent to not implement a "bad" rand generator in `test` so that a user doesn't accidentally enable it for their app? It seems like we should be able to use the OS's rand for test purposes?

ryan-summers (commented):
One comment, otherwise looks good to me :) Nice approach to avoid templatizing everything over some `impl RandCore`

src/socket/tcp.rs
Dirbaio: Many tests rely on the starting sequence number to be 10000. Changing them to make them work with a random starting seq number is annoying, and wouldn't really add much value.

One alternative I tried is making the actual RNG a mock when building in test mode, so the tests can "magically influence" what numbers come out of the RNG. The problem is tests would have to be aware of how and when does TCP use the RNG, which is kind of an implementation detail (ie when does it request bytes, how many bytes does it request, are they little/big endian to form the u32 sequence number...?)

ryan-summers: Ah, I see the limitation now. It just feels like addressing it at the `tcp.rs` layer may be introducing tech-debt for the future, but it looks extremely lightweight, so I'm not worried. Thanks for the explanation!

PR #559, opened 2021-10-24
Added PXE DHCP options

Qubasa:
I added three new DHCP options for PXE boot defined in this [RFC4578](https://datatracker.ietf.org/doc/html/rfc4578)
Appropriate tests have been added too. I also added two option fields vendor id and time offset from [RFC2132]( https://datatracker.ietf.org/doc/html/rfc2132)


.github/workflows/clippy.yml
```diff
@@ -23,4 +23,4 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --tests --examples -- -D warnings
+          args: --tests --examples -- -D warnings -A clippy::large-enum-variant
```
crawford: Can you use an attribute on the specific enum instead? I think that would help everyone running `cargo clippy` locally.

Qubasa: @crawford  You are right I changed it.

src/wire/dhcpv4.rs
```diff
@@ -98,6 +205,31 @@ impl<'a> DhcpOption<'a> {
...
                         option = DhcpOption::RequestedIp(Ipv4Address::from_bytes(data));
                     }
+                    (field::OPT_TIME_OFFSET, 4) => {
+                        option = DhcpOption::TimeOffset(u32::from_be_bytes([
+                            data[0], data[1], data[2], data[3],
+                        ]));
+                    }
+                    (field::OPT_CLIENT_ARCH, _) => option = DhcpOption::ClientArchTypeList(data),
+                    (field::OPT_CLIENT_MACHINE_ID, _) => {
+                        let id_type = MachineIdType::try_from(data[0])?;
```
Dirbaio, line 274: This panics if `data.len() = 0`.

Dirbaio (changes requested):
Thanks for the PR! Code looks good.

The DHCP fuzzer (#562) found a few issues though.
- Out of bounds at line 215 if option data length is 0.
- emit panics at line 306, maybe something wrong in `buffer_len`.

Could you fix them, and then run the fuzzer yourself to check there are no more? If you get the fuzzer to run for 10-15min without finding anything it should be good to go.

src/wire/dhcpv4.rs
```diff
@@ -21,6 +22,107 @@ enum_with_unknown! {
...
+    /// Intel Lean Client
+    IntelLeanClient = 5,
+    /// EFI IA32
+    EfiIA32 = 6,
+    /// EFI BC
+    EfiBc = 7,
+    /// EFI Xscale
+    EfiXscale = 8,
+    /// EFI x86-64
+    EfiX86_64 = 9,
```
nbdd0121: This is wrong. 7 is EFI x86-64 while 9 is EFI BC, see errata https://www.rfc-editor.org/errata_search.php?rfc=4578.

According to the errata this field should follow the IANA assignments for DHCPv6 options (https://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml#processor-architecture), which includes a few additional types (boot from http, other architecture like ARM, RISC-V etc), and deprecates a few ones.

Qubasa: @nbdd0121  Ohh thanks interesting! I never thought of looking for an errata, yeah I will fix that tomorrow~

src/wire/dhcpv4.rs
Luis-Hebendanz: @Dirbaio Um I don't seem to get the fuzzer to work. I tried running `cargo install cargo-fuzz` &&  `cargo fuzz run dhcp_header` 
``` 
Finished release [optimized] target(s) in 0.01s
     Running `target/x86_64-unknown-linux-gnu/release/dhcp_header -artifact_prefix=/home/lhebendanz/Projects/smoltcp/fuzz/artifacts/dhcp_header/ /home/lhebendanz/Projects/smoltcp/fuzz/corpus/dhcp_header`
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3421684815
INFO:        0 files found in /home/lhebendanz/Projects/smoltcp/fuzz/corpus/dhcp_header
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
#2	INITED exec/s: 0 rss: 31Mb
ERROR: no interesting inputs were found. Is the code instrumented for coverage? Exiting.
────────────────────────────────────────────────────────────────────────────────

Error: Fuzz target exited with exit status: 1
```

with rustc `nightly-2021-09-19` `clang --version`
```
clang version 13.0.0
Target: x86_64-unknown-linux-gnu
Thread model: posix
```

Luis-Hebendanz: Ah I had to update my nightly version

Qubasa: Fixed

src/wire/dhcpv4.rs
Qubasa: Fixed

src/wire/dhcpv4.rs
```diff
@@ -166,13 +362,39 @@ impl<'a> DhcpOption<'a> {
             }
             _ => {
                 skip_length = self.buffer_len();
+
+                assert!(skip_length <= buffer.len());
```
Qubasa, line 368: @whitequark @Dirbaio I have a problem I can't return an error in this function, so in the case someone puts an 'Other' option with a very large payload into the function and the supplied buffer is smaller, then this function panics. I don't want to silently return and ignore the option either. But the fuzzer will find this crash and stop there.  What I did now is insert an assert so the panic is easier to reason about

Dirbaio: It's OK for `emit` to panic if the supplied buffer len is not exactly equal to the length provided by `buffer_len`.It's the caller's responsibility to ensure that. 

I guess it's either a bug in the fuzzer (but from a quick look it looks OK to me) or a mismatch in the length you return from `buffer_len` and the length you require in `emit`...

Qubasa: @Dirbaio Ah alright yeah the Repr returns the incorrect size and the fuzzer allocates exactly that size, I see! As I'm looking into it that bug already existed beforehand

Qubasa:
If someone wants to open up a new reworked pull request go ahead :-) I sadly won't be able to work on this for at least a couple of months sadly

PR #560, opened 2021-10-26
IEEE802154: getters for security fields

thvdveld:
This adds getters for the security fields of the IEEE802154 frame. Also added a test for it.

`src/wire/ieee802154.rs` also contained a function for calculating the CRC. However, this is not needed and thus I also remove it in this PR.


Dirbaio (approved):
Thank you!

bors r+

Dirbaio:
it took 1 hour to merge, wtf?

thvdveld:
Yeah, really weird, I was wondering when it was going to merge 😅

PR #562, opened 2021-10-28
fuzz: DHCP header parser

alexandrasandulescu:
Hi. This fuzz target tests the [DHCPRepr](https://docs.rs/smoltcp/0.7.5/smoltcp/wire/struct.DhcpRepr.html) `parse` and `emit` functions ([context](https://github.com/smoltcp-rs/smoltcp/issues/305#issuecomment-530327153)).
The fuzz target requires a small change in the `DHCPOption` parser because the option type `OPT_DOMAIN_NAME_SERVER ` requires the payload representing the DNS server list to have a length multiple of 4. Otherwise the `chunks` call places the remainder in the last chunk (length < 4). This panics when the last chunk is parsed into an `IPv4Address`. Do you have a better suggestion for handling this situation?

I ran the fuzzer for a couple of days but it didn't find any crash yet. To increase the chances, I used the oss-fuzz seed corpora of `dnsmasq` DHCP fuzzer and `systemd` DHCP server fuzzer.
Please share if you have suggestions on how to improve coverage.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -28,3 +28,9 @@ name = "tcp_headers"
 path = "fuzz_targets/tcp_headers.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "dhcp_header"
+path = "fuzz_targets/dhcp_header.rs"
+test = false
+doc = false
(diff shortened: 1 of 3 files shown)


Dirbaio (approved):
> Otherwise the chunks call places the remainder in the last chunk (length < 4). This panics when the last chunk is parsed into an IPv4Address. Do you have a better suggestion for handling this situation?

Returning `Malformed` is the right behavior, yes. This is an actual bug you found thanks to fuzzing! Packet parsing should never panic. Nice! :D

Thank you!

bors r+

alexandrasandulescu:
@Dirbaio thank you very much for your review and thank you for smoltcp!
If you have suggestions for other sensitive components that should have fuzz targets, I am happy to continue working on this :D.

PR #563, opened 2021-10-30
Print log to stdout in test

deantvv:
Use simple [trick](https://github.com/rust-lang/log/issues/106) to allow print log to stdout in test

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,7 +1,15 @@
+#[cfg(not(test))]
 #[cfg(feature = "log")]
 macro_rules! net_log {
     (trace, $($arg:expr),*) => { log::trace!($($arg),*) };
     (debug, $($arg:expr),*) => { log::debug!($($arg),*) };
+}
+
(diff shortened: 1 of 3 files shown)


Dirbaio:
Interesting.

Some [existing](https://github.com/smoltcp-rs/smoltcp/blob/24fa299529df13a78a294e1b321dab15c8e0528d/src/socket/dhcpv4.rs#L655) [tests](https://github.com/smoltcp-rs/smoltcp/blob/24fa299529df13a78a294e1b321dab15c8e0528d/src/socket/tcp.rs#L2654) explicitly set a logger that delegates to `println!`, but not all (I guess you were seeing this in `iface` tests right?). 

Changing the macros to `println!` is definitely cleaner, it'll apply to all tests without them having to explicitly init `log` which is very nice.

Dirbaio (approved):
I've removed the manual log init since it's no longer needed now!

Thanks for the PR :) 

bors r+

Issue #543, opened 2021-10-06
arp: flush ARP cache when disconnecting or changing address


Dirbaio:
ARP cache should be flushed when

1. Our IP address changes
2. The interface goes down

Otherwise, the device may get plugged in into a network where eg router is still 192.168.1.1, but is another MAC addr. Communication might not work until the ARP cache entry expires.

elpiel:
I suppose only the second point of this issue remains to be fixed after #564 was merged?

PR #564, opened 2021-10-31
arp: flush neighbor cache after IP update

deantvv:
Related to #543


Dirbaio (approved):
Thank you! Nice code, nice tests too :) 

bors r+

PR #565, opened 2021-11-04
Add fuzzing for IEEE802.15.4 frame

thvdveld:
Because IEEE802.15.4 uses compression in its frame, fuzzing it is maybe a good idea. Adding this fuzz target showed that some frame methods were panicking. `check_len` now checks if accessors will panic or not.

I ran the fuzzer for about 15 minutes and nothing showed up after the changes in `check_len`.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -34,3 +34,9 @@ name = "dhcp_header"
 path = "fuzz_targets/dhcp_header.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "ieee802154_header"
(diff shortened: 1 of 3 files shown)


Dirbaio (approved):
Very nice! Fuzzing is awesome. LGTM

It'd be cool to test buffer_len + emit, like in the DHCP fuzzer. Can be in this PR or in a future one, feel free to merge this one in that case. Up to you :) 

bors d+

thvdveld:
Added `buffer_len` and `emit` (seems to be working as well, no panicks after ~10 minutes).

bors r+

PR #567, opened 2021-11-10
IEEE802.15.4 decryption

thvdveld:
Decryption of IEEE802154 frames. Still very much a work in progress.

I am currently able to decrypt and check the authentication for a 32-bit authentication tag (the EncMic32 mode). Other modes are implemented, except for the encryption only mode (Enc mode), since this requires a modified AES-CCM construction (IEEE802.15.4 uses AES-CCM*).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,9 @@ bitflags = { version = "1.0", default-features = false }
 defmt = { version = "0.2.0", optional = true }
 rand_core = { version = "0.6.3", optional = true, default-features = false }
 
-ccm = "0.4.4"
-cipher = "0.3.0"
-aes = "0.7.5"
+ccm = { version = "0.4.4", optional = true }
+cipher = { version = "0.3.0", optional = true }
+aes = { version = "0.7.5", optional = true }
 
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -24,6 +24,11 @@ bitflags = { version = "1.0", default-features = false }
 defmt = { version = "0.2.0", optional = true }
 rand_core = { version = "0.6.3", optional = true, default-features = false }
 
+ccm = "0.4.4"
```
Dirbaio: crypto usually causes giant code size bloat, it should be optional.

It should be excluded If 802154 is not used at least. In this case I think the linker will remove the bloat anyway, so maybe it's just a code size optimiation.

I don't know how often is unencrypted 802154 used in the real world? If the answer is "never" we could force "if you use 802154 you have crypto", but otherwise it should be possible to build 802154 support without crypto.

thvdveld: I also have no idea how often unencrypted is used in the real world.
I put it behind a feature flag (another one ;p).

thvdveld:
I don't think this should be in smoltcp, so I'm closing this.

PR #568, opened 2021-11-14
Expose `self.sockets` in `Interface`

korken89:
This is needed to not loose access to methods on sockets, e.g. iterating
over them and closing or checking which ports are being used.

Not sure if there is some other way this is meant to be done.

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -555,6 +555,16 @@ where
         &mut self.device
     }
 
+    /// Get an iterator to the inner sockets.
+    pub fn sockets(&self) -> impl Iterator<Item = &Socket<'a>> {
+        self.sockets.iter()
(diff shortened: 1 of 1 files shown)


Dirbaio (changes requested):
I was hoping to make SocketSet private later on after the initial "simplify" PR. If we expose it, it's no longer possible.

Could you try adding individual methodsto Interface to iterate that call the socketset internally? I think that should work as nicely for your usecase wihtout requiring SocketSet public.

```rust
fn sockets(&self) -> impl Iterator<Item=&Socket> {}
fn sockets_mut(&mut self) -> impl Iterator<Item=&mut Socket> {}

Dirbaio (approved):
Thank you!

bors r+

PR #569, opened 2021-11-15
Support defmt version span 0.2 to 0.3

korken89:
A trick to support multiple versions of `defmt`. This allows the user crate to select the version in the end.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
         # Failure is permitted on nightly.
         rust:
           - stable
-          - 1.53.0
+          - 1.56.0
           - nightly
 
         features:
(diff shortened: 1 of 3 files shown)


korken89:
Huh, running exactly the failing test locally does not trigger an error

korken89:
Ahh, it's the 1.53 version that is the issue.

Maybe this feature is a bit too early?

Dirbaio:
I do want defmt 0.3 too, so :+1: to bumping the MSRV.

One downside is that we still have to keep the `defmt-trace, defmt-debug` etc Cargo features to support 0.2, but they do nothing in 0.3. This could be a bit confusing to users. I wonder if dropping support for 0.2 would be better, since this is going to be released in 0.8 which is breaking anyway.

korken89:
Sure, I see no problem in bumping except that during the `0.2 -> 0.3` transition period people might have issues.

I've not stumbled on any yet, but yeah. I'll follow however you'd like it :)

Dirbaio:
I think it's best to not support 0.2...

Another downside is if 0.3.x introduces new features (like new format specifiers) we won't be able to use them because they won't work in 0.2.

Also, the released 0.7.x versions don't have defmt support, so this would only be an issue for people using git upgrading to a newer git or to 0.8.

korken89:
Fixed!

Dirbaio (approved):
Neat! Thank you :) 

bors r+

PR #572, opened 2021-11-24
GAT-based Device trait.

Dirbaio:
The current `'a` lifetime in the `Device` trait is essentially a workaround for lack of GATs.
I'm just experimenting how this would look like, it'll have to wait until GATs are stable to go in.

The main benefit is structs implementing `Device` can now borrow stuff. This wasn't possible
before because the `for<'d> T: Device<'d>` bounds would essentially imply `T: 'static`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,8 +20,8 @@ jobs:
         # Test on stable, MSRV, and nightly.
         # Failure is permitted on nightly.
         rust:
-          - stable
-          - 1.61.0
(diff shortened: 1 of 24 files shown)


gopakumarce:
@Dirbaio looks like GAT is gonna be in 1.65 (https://github.com/rust-lang/rust/pull/96709) - hopefully this gets merged then !

gopakumarce:
@Dirbaio awesome to see this merged, thanks! I hope you will also cut out a new smoltcp release in crates.io that has this changeset

PR #573, opened 2021-11-25
Simpler rand

Dirbaio:
Requires #571 

Revisits #547, because it turns out the `rand_custom_impl!` macro stuff was less nice than I thought. It's now a simple PRNG owned by Interface. The builder allows setting the random seed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,7 @@ jobs:
         include:
           # Test alloc feature which requires nightly.
           - rust: nightly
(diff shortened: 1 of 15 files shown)


crawford:
Having just implemented against the earlier Rand interface, I like this better. My device RNG (and I imagine others) requires some setup before it can be used, so this ownership model removes the runtime checks needed in the previous interface (making sure that the RNG was configured before first-use). 👍

thvdveld:
The sequence number for IEEE 802.15.4 needs to be random and not equal to zero when the interface is build. I guess we can now set it when building the interface, but then we should remove the `sequence_no` function in the interface builder.

Made a PR for it #575

PR #575, opened 2021-12-03
ieee802154: random sequence number

thvdveld:
Set the sequence number to a random value when building the interface (builds on #573).


Dirbaio (approved):
oh, nice! Thank you!

PR #579, opened 2021-12-11
Remove IpRepr::Unspecified

Dirbaio:
Implement simplifications unlocked by adding the `Context` parameter to socket methods. 

- tcp: immediately choose src addr on connect. Now that we have access to `Context` from `.connect()`, we don't have to delay setting it anymore.
- Remove IpRepr::Unspecified and lowering. Choosing the right source IP address is now responsibility of the individual sockets.


Dirbaio:
3 months later... :snail: 

bors r+

PR #580, opened 2021-12-17
6LoWPAN fragmentation

thvdveld:
This adds the fragmentation part of 6LoWPAN from [RFC4944 5.3](https://datatracker.ietf.org/doc/html/rfc4944#section-5.3). This is still a work in progress:

- [x] Add wire format of the fragmentation header of 6LoWPAN.
- [x] Update documentation in wire.
- [x] Use packet assembler (based on the IPv4 assembling #236).
- [x] Use assembling in the interface for incoming packets.
    - [x] Assemble packets.
    - [x] Correct checksum (still need to find out why my checksums are incorrect).
- [x] Check the interval between incoming fragments.
- [x] Fragment outgoing packets (still not sure how to do this part).
- [x] Add overlap detection in `Assembler`.
- [x] Handle stale fragments.
- [x] Do the correct thing when encountering fragment overlaps.
- [x] Finish address resolution for 6LoWPAN.
- [x] Try not to use a temporary buffer when trying to fragment outgoing packets.
    ~~**NOTE**: This requires more changes. For example: the emit function for UDP requires that the whole buffer is passed to it such that even the payload fits in it. However, IEEE802154 only has buffers of 127 bytes (125 bytes without the CRC).~~

### DIFF
--- a/src/wire/sixlowpan.rs
+++ b/src/wire/sixlowpan.rs
@@ -7,6 +7,408 @@ use crate::wire::ieee802154::Address as LlAddress;
 use crate::wire::ipv6;
 use crate::wire::IpProtocol;
 
+/// The representation of an unresolved address. 6LoWPAN compression of IPv6 addresses can be with
+/// and without context information. The decompression with context information is not yet
+/// implemented.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum UnresolvedAddress<'a> {
+    WithoutContext(AddressMode<'a>),
+    WithContext(AddressMode<'a>),
+    Reserved,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum AddressMode<'a> {
+    /// The full address is carried in-line.
+    FullInline(&'a [u8]),
+    /// The first 64-bits of the address are elided. The value of those bits
+    /// is the link-local prefix padded with zeros. The remaining 64 bits are
+    /// carried in-line.
+    InLine64bits(&'a [u8]),
+    /// The first 112 bits of the address are elided. The value of the first
+    /// 64 bits is the link-local prefix padded with zeros. The following 64 bits
+    /// are 0000:00ff:fe00:XXXX, where XXXX are the 16 bits carried in-line.
+    InLine16bits(&'a [u8]),
+    /// The address is fully elided. The first 64 bits of the address are
+    /// the link-local prefix padded with zeros. The remaining 64 bits are
+    /// computed from the encapsulating header (e.g., 802.15.4 or IPv6 source address)
+    /// as specified in Section 3.2.2.
+    FullyElided,
+    /// The address takes the form ffXX::00XX:XXXX:XXXX
+    Multicast48bits(&'a [u8]),
+    /// The address takes the form ffXX::00XX:XXXX.
+    Multicast32bits(&'a [u8]),
+    /// The address takes the form ff02::00XX.
+    Multicast8bits(&'a [u8]),
+    /// The unspecified address.
+    Unspecified,
+    NotSupported,
+}
+
+const LINK_LOCAL_PREFIX: [u8; 2] = [0xfe, 0x80];
+const EUI64_MIDDLE_VALUE: [u8; 2] = [0xff, 0xfe];
+
+impl<'a> UnresolvedAddress<'a> {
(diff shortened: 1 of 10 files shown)


thvdveld:
Shouldn't we remove `proto-sixlowpan` from the default features? When I would use `smoltcp` with the default features, then it wouldn't be for an embedded device (since this would require `alloc` and `std`). I think `proto-sixlowpan` is only useful for embedded devices and since I added the 6LoWPAN fragments cache to the `Interface`, everyone should pass a cache for it now, even though it is not used for Ethernet or IP but only for `6LoWPAN`.

thvdveld:
Fragmentation is currently working as expected. Now I think it needs some clean-up.

src/wire/sixlowpan.rs
```diff
@@ -51,13 +51,271 @@ impl<'a> Address<'a> {
 
                 Ok(ipv6::Address::from_bytes(&bytes))
             }
-            Address::WithContext(_) => Err(Error::NotSupported),
+            Address::WithContext(ctx) => {
+                if ctx.is_empty() {
+                    let mut bytes = [0; 16];
+                    bytes[0] = 0xfe;
```
dlrobertson: Could we have constants for these, a note about them, or a mention of the relevant RFC section?

src/wire/sixlowpan.rs
```diff
@@ -51,13 +51,271 @@ impl<'a> Address<'a> {
...
+                    bytes[1] = 0x80;
+
+                    match ll_addr {
+                        Some(LlAddress::Short(ll)) => {
+                            bytes[11] = 0xff;
```
dlrobertson: Same as above.

dlrobertson (commented):
Big fan of the 6LoWPAN work you've been doing. Sorry I wasn't around to review the initial implementation. On a quick overview the only note I have is that I'd like a few more references to the RFC and relevant sections. For a less experienced 6LoWPAN user like me, more comments would be helpful. Thanks for the work on this!

thvdveld:
I added constants where I could and some more documentation.

For the resolving of the address in `src/wire/sixlowpan.rs`, the part with the `WithContext` I don't really understand yet how that works from the RFC. Still need to take a better look at it.

thvdveld:
I tested this in hardware (CC2538 from TI) and everything seems to work like expected.

Dirbaio:
could you rebase instead of merging? the history is a bit messy otherwise

thvdveld:
Sorry my bad. Fixed now!

thvdveld:
Still need to change some API stuff. I don't like it right now.

Dirbaio:
> Still need to change some API stuff. I don't like it right now.

@thvdveld what's the status? Have you finished these changes, it it ready for review?

thvdveld:
Didn't find the time yet, only found time to rebase and fix some things. It's ready for review. But just to let you know, I'm not happy with how the buffer is passed to the interface for the outgoing and incoming fragments. Also, there is still the thing of the buffer in the dispatching of the outgoing fragments, which I'm not happy about either and didn't find the time yet to make it better.

https://github.com/smoltcp-rs/smoltcp/blob/c11e4bb6a7a1ca78d3a45a6e28237407ed0fce9d/src/iface/interface.rs#L2788

thvdveld:
Also, something that I am not happy with is the following:

https://github.com/smoltcp-rs/smoltcp/blob/c11e4bb6a7a1ca78d3a45a6e28237407ed0fce9d/src/iface/interface.rs#L844-L853

This is for sending out the fragmented packets. This is not ideal I guess and should be changed IMO.

Dirbaio:
I see! So, currently, you:

- `Interface` owns a fragment queue
- in `dispatch_ieee802154`, emit the packet to a temp buffer
- in `dispatch_ieee802154`, fragment it all upfront into `out_fragments`
- in `poll`, send the queued fragments

What about fragmenting "lazily" instead?

- `Interface` has `out_packet: &[u8], out_packet_sent_bytes: usize`
- in `dispatch_ieee802154`, emit the packet into `out_packet`, set `out_packet_sent_bytes=0`.
- in `poll`, if `out_packet_sent_bytes < out_packet.len()`, consume a tx token, emit one fragment into the tx token buffer, update sent_bytes`.

This way you only need one buffer instead of the previous 2 (the unfragmented packet and the fragment queue).

Also, if you send pending fragments the first thing in `poll`, in dispatch you no longer have to worry about the out_packet buffer being full (because it wouldn't have been called otherwise).

thvdveld:
Good idea! I'm working on that and will probably have this finished next week.

thvdveld:
I think it is ready for review!

Dirbaio (approved):
Looks awesome, let's do this! Thanks for the great work @thvdveld!

bors r+

PR #586, opened 2022-01-28
Update dependencies

alexandrasandulescu:
Hi. This PR bumps smoltcp dependencies version to `latest` according to [crates.io](https://crates.io/).
The `env_logger` update brings an API change, the [parse](https://docs.rs/env_logger/0.5.0/src/env_logger/lib.rs.html#380-383) Builder method was replaced with [parse_filters](https://docs.rs/env_logger/0.9.0/env_logger/struct.Builder.html#method.parse_filters).

Apart from `parse_filters`, I noticed that the original code parsed `RUST_LOG` **after** setting the log level to `trace` and parsing the filters. My understanding is that this overwrites the builder setup with what `RUST_LOG` contains.
My change parses the [default environment](https://docs.rs/env_logger/0.9.0/env_logger/struct.Env.html#default-environment-variables) first with `from_default_env` and then applies the filter changes.

If the desired behaviour is the original one (overwrite `setup_logging_with_clock` filters with the ones defined by `RUST_LOG`) let me know.

Tnx :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,10 @@ autoexamples = false
 
 [dependencies]
 managed = { version = "0.8", default-features = false, features = ["map"] }
-byteorder = { version = "1.4", default-features = false }
-log = { version = "0.4.14", default-features = false, optional = true }
-libc = { version = "0.2.115", optional = true }
-bitflags = { version = "1.3", default-features = false }
+byteorder = { version = "1.0", default-features = false }
+log = { version = "0.4.4", default-features = false, optional = true }
+libc = { version = "0.2.18", optional = true }
+bitflags = { version = "1.0", default-features = false }
 defmt = { version = "0.3", optional = true }
 rand_core = { version = "0.6.3", optional = true, default-features = false }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,18 +17,18 @@ autoexamples = false
 
 [dependencies]
 managed = { version = "0.8", default-features = false, features = ["map"] }
-byteorder = { version = "1.0", default-features = false }
-log = { version = "0.4.4", default-features = false, optional = true }
-libc = { version = "0.2.18", optional = true }
-bitflags = { version = "1.0", default-features = false }
+byteorder = { version = "1.4", default-features = false }
+log = { version = "0.4.14", default-features = false, optional = true }
+libc = { version = "0.2.115", optional = true }
+bitflags = { version = "1.3", default-features = false }
 defmt = { version = "0.3", optional = true }
 rand_core = { version = "0.6.3", optional = true, default-features = false }
 
 [dev-dependencies]
-env_logger = "0.5"
+env_logger = "0.9"
 getopts = "0.2"
-rand = "0.3"
-url = "1.0"
+rand = "0.8"
+url = "2.0"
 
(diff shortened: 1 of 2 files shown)


alexandrasandulescu:
One [test fails](https://github.com/smoltcp-rs/smoltcp/runs/4980497275?check_suite_focus=true#step:3:32) but it looks unrelated to the changes. I cannot rerun the jobs in the test workflow.

crawford:
What’s the benefit of raising the minimum required version for these dependencies? E.g. is there a specific behavior that is being corrected?

As for the parsing of RUST_LOG, I think it’s generally desirable to allow environment variables to override defaults. This is how rustc and cargo handle the variable, for example.

alexandrasandulescu:
> What’s the benefit of raising the minimum required version for these dependencies? E.g. is there a specific behavior that is being corrected?

The current dependencies version specification do not allow updates for `env_logger`, `rand` and `url` (according to [Dependencies specification](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements)).
I use smoltcp in a setting where only certain dependencies versions are available (eg. env_logger > 0.9).

> As for the parsing of RUST_LOG, I think it’s generally desirable to allow environment variables to override defaults. This is how rustc and cargo handle the variable, for example.

`env_log` defines "RUST_LOG" as default environment, which led me to the (wrong) assumption that it should be parsed first. Thank you for clarifying this issue.

adamgreig:
> The current dependencies version specification do not allow updates for env_logger, rand and url (according to Dependencies specification).

That makes sense for those three dev-dependencies; is there any need to (or benefit from) updating the four dependencies byteorder/log/libc/bitflags?

alexandrasandulescu:
> That makes sense for those three dev-dependencies; is there any need to (or benefit from) updating the four dependencies byteorder/log/libc/bitflags?

No. I added another commit based on your feedback. The PR now updates only `env_logger`, `rand` and `url`. Moreover, the `RUST_LOG` env overrides the default logging setup.

crawford:
> The current dependencies version specification do not allow updates for `env_logger`, `rand` and `url`

I’m confused by this. My understanding of the documentation is that `env_logger` and `rand` _could_ be updated, since the major version doesn’t change. Am I mistaken?

adamgreig:
In Cargo, versions `0.x.y` are effectively treated with `x` as the major and `y` as the minor version, so `^0.4.5` will use `0.4.6`, but not `0.5.0`. Once the first 1.0 version is published, `^1.2.3` would use `1.3.4` but not `2.0`.

So, in this case, `env_logger ^0.5` wouldn't use `0.9`, and `rand ^0.3` wouldn't use `0.8`.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

> This compatibility convention is different from SemVer in the way it treats versions before 1.0.0. While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

crawford:
Ah, I forgot about that caveat. Thanks for explaining.

crawford (approved):

Dirbaio (approved):
This is nice, thanks for the PR! :) 

bors r+

Issue #587, opened 2022-02-08
Panic in src/phy/sys/raw_socket.rs


witchof0x20:
I'm using smoltcp to ferry packets between two interfaces via raw sockets (with some custom filtering), and ran into this panic when stress-testing my program:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }',
 /home/me/.cargo/registry/src/github.com-1ecc6299db9ec823/smoltcp-0.8.0/src/phy/sys/raw_socket.rs:107:49
```
I see an explicit unwrap at https://github.com/smoltcp-rs/smoltcp/blob/master/src/phy/sys/raw_socket.rs#L107 :
```rust
if len == -1 {
    Err(io::Error::last_os_error()).unwrap()
}
```
and in the outer function that calls this at https://github.com/smoltcp-rs/smoltcp/blob/master/src/phy/raw_socket.rs#L114 :
```rust
lower..send(&buffer[..]).unwrap();
```

Is there a reason to explicitly panic here, or would it be reasonable to propagate that error up to the caller?

Dirbaio:
Oh, raw socket send can WouldBlock under load? huh.

The cleanest solution is probably returning an error from `consume`, yes. This way the stack knows the send failed and can retry later.

Perhaps translate `Exhausted` to `WouldBlock`, and the others to `Illegal`

PR #588, opened 2022-02-08
Propagate phy::RawSocket send error to caller

witchof0x20:
For #587

### DIFF
--- a/src/phy/raw_socket.rs
+++ b/src/phy/raw_socket.rs
@@ -111,12 +111,10 @@ impl phy::TxToken for TxToken {
         let mut lower = self.lower.borrow_mut();
         let mut buffer = vec![0; len];
         let result = f(&mut buffer);
-        if let Err(err) = lower.send(&buffer[..]) {
-            return match err.kind() {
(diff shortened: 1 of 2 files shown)


Dirbaio:
Thanks for the PR!

Now that I look at it again (sorry!) in `receive`, WouldBlock errors are propagated and everything else panics (instead of returning `Illegal`): https://github.com/smoltcp-rs/smoltcp/blob/61639d59211024f88ba1004740a1fe3cfd842b81/src/phy/raw_socket.rs#L75-L76

`send` and `receive` should probably be consistent, either both panic or both returning an error. I don't know in which cases can a RawSocket send/recv fail, I think they should be fairly rare since it's a "virtual" network thingy instead of actually talking to a peer, so maybe panicking is better?

witchof0x20:
Updated with consistent behavior. I generally lean towards propagating any error, but in the name of consistency, it makes sense to have the same behavior on both. If I run into other errors, I can file a separate issue!

Dirbaio (approved):
Thank you!

bors r+

PR #591, opened 2022-02-22
Fragmentation mechanism

thvdveld:
This contains the fragmentation mechanism used in #580 (which is heavily based on #236 taking into account the remarks that were made there). I tried to decouple it from the 6LoWPAN implementation such that it can be used for other protocols as well.

I opened a separate PR for this such that #580 will only contain the mechanisms that are part of 6LoWPAN only.

### DIFF
--- a/src/iface/fragmentation.rs
+++ b/src/iface/fragmentation.rs
@@ -466,10 +466,7 @@ mod tests {
 
         p_assembler.add(&data[..], 0, Instant::now()).unwrap();
 
-        assert_eq!(
-            p_assembler.add(&data[..], 1, Instant::now()),
-            Ok(true),
-        );
+        assert_eq!(p_assembler.add(&data[..], 1, Instant::now()), Ok(true));
     }
 
     #[test]
--- a/src/iface/fragmentation.rs
+++ b/src/iface/fragmentation.rs
@@ -1,0 +1,553 @@
+use managed::{ManagedMap, ManagedSlice};
+
+use crate::storage::Assembler;
+use crate::time::Instant;
+use crate::Error;
+use crate::Result;
+
+pub trait PacketAssemblerInfo: PartialEq {
+    /// Calculate a new offset based on some other information.
+    fn calc_offset(&self, offset: usize) -> usize;
+}
+
+#[derive(Debug, PartialEq)]
+pub struct NoInfo;
+
+impl PacketAssemblerInfo for NoInfo {
+    #[inline]
+    fn calc_offset(&self, offset: usize) -> usize {
+        offset
+    }
+}
+
+/// Holds different fragments of one packet, used for assembling fragmented packets.
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PacketAssembler<'a, Info: PacketAssemblerInfo = NoInfo> {
+    buffer: ManagedSlice<'a, u8>,
+    assembler: AssemblerState<Info>,
+}
+
+/// Holds the state of the assembling of one packet.
+#[derive(Debug, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+enum AssemblerState<Info: PacketAssemblerInfo = NoInfo> {
+    NotInit,
+    Assembling {
+        assembler: Assembler,
+        total_size: usize,
+        info: Info,
+        last_updated: Instant,
+        started_on: Instant,
+    },
+}
+
+impl<'a, Info: PacketAssemblerInfo> PacketAssembler<'a, Info> {
+    /// Create a new empty buffer for fragments.
+    pub fn new<S>(storage: S) -> Self
+    where
+        S: Into<ManagedSlice<'a, u8>>,
+    {
+        let s = storage.into();
+        PacketAssembler {
+            buffer: s,
+            assembler: AssemblerState::NotInit,
+        }
+    }
+
+    /// Start with saving fragments.
+    /// We initialize the assembler with the total size of the final packet.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`Error::PacketAssemblerBufferTooSmall`] when the buffer is too smal for holding all the
+    /// fragments of a packet.
+    pub(crate) fn start(
+        &mut self,
+        total_size: usize,
+        info: Info,
+        start_time: Instant,
+    ) -> Result<()> {
+        match &mut self.buffer {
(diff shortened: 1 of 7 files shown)


src/storage/assembler.rs
```diff
@@ -184,8 +193,9 @@ impl Assembler {
 
     /// Add a new contiguous range to the assembler, and return `Ok(())`,
     /// or return `Err(())` if too many discontiguities are already recorded.
-    pub fn add(&mut self, mut offset: usize, mut size: usize) -> Result<(), TooManyHolesError> {
+    pub fn add(&mut self, mut offset: usize, mut size: usize) -> Result<bool, TooManyHolesError> {
```
aramperes: Doc comment should be updated to reflect signature change (the `return Ok(())` bit)

thvdveld: Updated now, thanks!

thvdveld:
Friendly reminder: could someone review this PR? @Dirbaio @aramperes

src/iface/fragmentation.rs
```diff
@@ -0,0 +1,555 @@
...
+use managed::{ManagedMap, ManagedSlice};
+
+use crate::storage::Assembler;
+use crate::time::Instant;
+use crate::Error;
+use crate::Result;
+
+pub trait PacketAssemblerInfo: PartialEq {
+    /// Calculate a new offset based on some other information.
+    fn calc_offset(&self, offset: usize) -> usize;
```
Dirbaio: Why is this needed? from the 6lowpan code it seems it adds a fixed offset, could the calling code do that on its own instead?

src/iface/fragmentation.rs
```diff
@@ -0,0 +1,555 @@
...
+
+                let len = data.len();
+                self.buffer[offset..][..len].copy_from_slice(data);
+
+                match assembler.add(offset, data.len()) {
+                    Ok(false) => {
+                        *last_updated = now;
+                        self.is_complete()
+                    }
+                    Ok(true) => Err(Error::PacketAssemblerOverlap),
```
Dirbaio: is it required to drop packets with overlap? AFAIK it's not for IPv4 fragmentation. With that, we could avoid making changes to `Assembler`

src/iface/fragmentation.rs
```diff
@@ -0,0 +1,555 @@
...
+                if self.is_complete()? {
+                    let a = &self.buffer[..total_size];
+                    self.assembler = AssemblerState::NotInit;
+                    a
+                } else {
+                    return Err(Error::PacketAssemblerIncomplete);
+                }
+            }
+        };
+        self.assembler = AssemblerState::NotInit;
```
Dirbaio: this line is duplicated

src/iface/fragmentation.rs
```diff
@@ -0,0 +1,555 @@
...
+    ///
+    /// - Returns [`Error::PacketAssemblerSetFull`] when every [`PacketAssembler`] in the buffer is used (only
+    /// when the non allocating version of is used).
+    pub(crate) fn reserve_with_key(&mut self, key: &K) -> Result<&mut PacketAssembler<'a, Info>> {
+        if self.packet_buffer.len() == self.index_buffer.len() {
+            match &mut self.packet_buffer {
+                ManagedSlice::Borrowed(_) => return Err(Error::PacketAssemblerSetFull),
+                #[cfg(any(feature = "std", feature = "alloc"))]
+                ManagedSlice::Owned(b) => {
+                    b.resize_with(self.index_buffer.len() + 1, || {
```
Dirbaio: this could be a .push(), we're just pushing 1 elem.

src/iface/fragmentation.rs
```diff
@@ -0,0 +1,555 @@
...
+                }
+            }
+        }
+
+        let i = self
+            .get_free_packet_assembler()
+            .ok_or(Error::PacketAssemblerSetFull)?;
+
+        // NOTE(thvdveld): this should not fail because we already checked the available space.
+        match self.index_buffer.insert(*key, i as u8) {
```
Dirbaio: This can overflow the u8 when using `alloc` (or if the user gave space for more than 256 slots) and someone spams us with lots of incomplete fragmented packets.

I think we should have a cap on WIP reassembling packets, even when using alloc, otherwise we're vulnerabile to memory exhaustion attacks if someone spams us.

src/iface/fragmentation.rs
thvdveld: Indeed, you are right, I misread the RFC. When there is an overlap AND either the size or the offset is different of the overlapping packet, the packet needs to be dropped in the assembler and a fresh reassembly may start with the new packet. However, I currently don't save the size of the packet in the assembler, so there is no way of knowing that. I guess for now I can remove the dropping.

src/iface/fragmentation.rs
thvdveld: Changed it.

src/iface/fragmentation.rs
thvdveld: I currently set the cap to u8::MAX and change the index to usize instead of u8. This should prevent overflow and cap the WIP reassembling packets.

src/iface/fragmentation.rs
thvdveld: Maybe my solution is not the best. Here is what I tried to solve:

The size in the first fragment header represents the total size after decompression. However, the packets that arrive are compressed and I put the compressed packets in the fragmentation buffers. This means that I need to calculate the total size of the compressed form, which is the total size from the first fragment header, minus the uncompressed size (of all headers that are compressed), plus the compressed size (of all headers that are compressed). Also the offset in the fragment headers are the offset in the uncompressed form. Which also means that I need to recalculate the offset to the compressed format. I have no idea why they made it so difficult (probably to prevent some attacks, though I'm not sure).

I could try to do the following in 6LoWPAN to get rid of `PacketAssemblerInfo`: just decompress the packet before putting it in the packet assembler buffers. I didn't do this in the first place because then I would decompress in two places: first when we there is no fragmentation and second for fragmented packets.

src/iface/fragmentation.rs
thvdveld: Changed it.

src/iface/fragmentation.rs
thvdveld: I removed it for now in this PR and will see later how I will fix it for 6LoWPAN.

Dirbaio (approved):
Thank you!

Sorry for the mega late reply, I've been swamped with work :( 

bors r+

PR #593, opened 2022-03-19
use socklen_t instead of u32 for bind() parameter

trinity-1686a:
Fix compilation on Android 32b, where `socklen_t` is `i32`. This changes nothing on almost all other platforms as their `socklen_t` is  already `u32`.
Solarish and VxWorks are the two other exceptions, they both use `c_uint` as their `socklen_t`. For x86_64-sun-solaris and sparcv9-sun-solaris I was able to confirm `c_uint` is in fact a `u32`, so it worked before, and still works now.
For vxworks, I did not found a toolchain I could install with rustup to confirm whether it worked before, but anyway it should work now.


Dirbaio (approved):
Thank you! 

bors r+

Dirbaio:
bors r+ 

???

PR #600, opened 2022-04-16
Fixed an issue with DHCP client and an assert which conflict

john-sharratt:
The DHCP client will broadcast a packet at the IPv4 address 255.255.255.255 hence when using iface.poll_at it fails with an assert that is checking if the source address is unspecified which of course it is as it defaults to 0.0.0.0.

```
Apr 16 14:32:10.910 DEBUG smoltcp::socket::dhcpv4: DHCP send DISCOVER to 255.255.255.255: Repr { message_type: Discover, transaction_id: 2704349917, client_hardware_address: Address([6, 241, 20, 229, 85, 169]), client_ip: Address([0, 0, 0, 0]), your_ip: Address([0, 0, 0, 0]), server_ip: Address([0, 0, 0, 0]), router: None, subnet_mask: None, relay_agent_ip: Address([0, 0, 0, 0]), broadcast: false, requested_ip: None, client_identifier: Some(Address([6, 241, 20, 229, 85, 169])), server_identifier: None, parameter_request_list: Some([1, 3, 6]), dns_servers: None, max_size: Some(1418), lease_duration: None } 
```

Stack trace:
```
thread 'tcp_simple_dhcp' panicked at 'assertion failed: !ip_repr.src_addr().is_unspecified()', /home/john/.cargo/git/checkouts/smoltcp-3d1f2cb07ccf3010/3e7c21b/src/iface/interface.rs:2420:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:48:5
   3: smoltcp::iface::interface::InterfaceInner::dispatch_ip
             at /home/john/.cargo/git/checkouts/smoltcp-3d1f2cb07ccf3010/3e7c21b/src/iface/interface.rs:2420:9
   4: smoltcp::iface::interface::Interface<DeviceT>::socket_egress::{{closure}}
             at /home/john/.cargo/git/checkouts/smoltcp-3d1f2cb07ccf3010/3e7c21b/src/iface/interface.rs:942:21
   5: smoltcp::socket::dhcpv4::Dhcpv4Socket::dispatch
[...]

PR #602, opened 2022-04-19
(Incomplete) Implement `join_multicast_group()` for IPv6

jgallagher:
This is a preliminary and incomplete start to adding support for IPv6 multicast groups. This PR _only_ adds joining groups via sending an initial MLDv2 report packet; it doesn't support resending the change report, leaving groups, responding to queries, etc. I wanted to open this PR first both to keep the review burden lighter and to get any feedback on the implementation that could affect implementing the remaining features.

On that last point, there are a few things in this PR that I'm not thrilled with and would be happy for any suggestions. I think all of them derive from difficulty crafting a fully-formed `IpPacket` to pass to `dispatch_ip()`. For the MLDv2 report, the specific difficulties and paths I took to deal with them were:

1. We are required to set the router alert option in an ipv6 hop-by-hop header, which requires us to be able to craft a packet with a hop-by-hop header in the first place! I added a new case to `IpPacket` (`IpPacket::HopByHopIcmpv6`), but that doesn't feel quite right.
2. `Ipv6HopByHopRepr`'s `options` field requires a borrowed slice. I couldn't see an easy way to get from a pair of `Ipv6OptionRepr`s to a borrowed slice, so I punted on this and defined `Ipv6HopByHopRepr::MLDV2_ROUTER_ALERT` as a constant.
3. Similarly, `MldRepr::Report` requires a borrowed slice for its data (the address records); I couldn't see an easy way to get from an `MldAddressRecordRepr` to that borrowed data, so I added an `MldRepr::ReportRecordReprs` case which holds a slice of `MldAddressRecordRepr`. This definitely seems wrong. It also might be still have similar problems as the borrowed byte slice when we need to craft packets with multiple address records in them.

In addition to the unit tests present, I've tested this locally on a microcontroller running smoltcp, and I've confirmed I can communicate with it via a multicast address through the switch it's connected to, so I'm reasonably confident that what's here is correct (albeit incomplete and a little hacky, as noted above).

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -2858,7 +2858,7 @@ mod test {
         iface_builder.finalize()
     }
 
-    #[cfg(feature = "proto-igmp")]
+    #[cfg(any(feature = "proto-igmp", feature = "proto-ipv6"))]
     fn recv_all(iface: &mut Interface<'_, Loopback>, timestamp: Instant) -> Vec<Vec<u8>> {
         let mut pkts = Vec::new();
         while let Some((rx, _tx)) = iface.device.receive() {
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -2023,7 +2023,6 @@ impl<'a> InterfaceInner<'a> {
             let opt_repr = result?;
             match opt_repr {
                 Ipv6OptionRepr::Pad1 | Ipv6OptionRepr::PadN(_) | Ipv6OptionRepr::RouterAlert(_) => {
-                    ()
                 }
                 Ipv6OptionRepr::Unknown { type_, .. } => {
                     match Ipv6OptionFailureType::from(type_) {
@@ -2511,7 +2510,6 @@ impl<'a> InterfaceInner<'a> {
 
     fn dispatch_ip<Tx: TxToken>(&mut self, tx_token: Tx, packet: IpPacket) -> Result<()> {
         let ip_repr = packet.ip_repr();
-        assert!(!ip_repr.src_addr().is_unspecified());
         assert!(!ip_repr.dst_addr().is_unspecified());
 
         match self.caps.medium {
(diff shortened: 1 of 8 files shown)


src/iface/interface.rs
```diff
@@ -2417,7 +2510,6 @@ impl<'a> InterfaceInner<'a> {
 
     fn dispatch_ip<Tx: TxToken>(&mut self, tx_token: Tx, packet: IpPacket) -> Result<()> {
         let ip_repr = packet.ip_repr();
-        assert!(!ip_repr.src_addr().is_unspecified());
```
jgallagher, line 2420: Removing this assertion also feels a little dicey, but as noted in the comment when crafting the mldv2 report packet, [RFC 3810](https://datatracker.ietf.org/doc/html/rfc3810#section-5.2.13) instructs that the source address can be set to the unspecified address:

> An MLDv2 Report MUST be sent with a valid IPv6 link-local source
   address, or the unspecified address (::), if the sending interface
   has not acquired a valid link-local address yet.  Sending reports
   with the unspecified address is allowed to support the use of IP
   multicast in the Neighbor Discovery Protocol [[RFC2461](https://datatracker.ietf.org/doc/html/rfc2461)].

Rahix: It seems #600 suggests to remove this assert as well?

Dirbaio: fixed by #616

Dirbaio:
re the difficulties: Yes, `IpPacket` is not "scaling" well, this happens with every new feature added. Your solutions look reasonable. (I've been thinking for a while about refactoring it so that all the packet emitters (sockets, and the IGMP code) gets access to the buffer and write their packets there directly, but that has other downsides...)

From a quick look, the PR looks OK. Could you rebase+squash?

Dirbaio:
friendly ping @jgallagher  :)

thvdveld:
This is implemented and merged in https://github.com/smoltcp-rs/smoltcp/pull/914, which is based on this PR.

PR #603, opened 2022-04-20
tcp socket: keep settings on reset, fix #601

### LINKED ISSUE #601: `socket.listen()` clears previously-set settings such as keep-alive
The following code behaves as expected:
```rust
let mut socket = TcpSocket::new(/* ... */);
socket.listen(dst_addr)?;
socket.set_keep_alive(Some(Duration::from_secs(30)));
dbg!(socket.keep_alive()) // Some(30s)
```

However, if we rearrange the calls to `listen` and `set_keep_alive`, the keep-alive value is forgotten:
```rust
let mut socket = TcpSocket::new(/* ... */);
socket.set_keep_alive(Some(Duration::from_secs(30)));
socket.listen(dst_addr)?;
dbg!(socket.keep_alive()) // None
```

This is unexpected as a similar reordering would work fine on Berkeley sockets. The reason for the behavior is that `socket.listen()` internally calls `socket.reset()`, which among other things clears the keep-alive value. The same problem also affects `ack_delay`, `hop_limit`, `nagle`, and `timeout`. 

I'm happy to submit a patch, but I don't know enough of smoltcp's internals to know what the intended fix should be:

1. Change `reset()` to not modify these values?
2. Do no call `reset()` in `.listen()`? (there already is a check for `is_open()`)?
3. Simply document that `listen()` must be called first?

mhils:
This PR fixes #601 by not overriding the settings on `.reset()`. I've added one test that documents intent, I feel like there is little value in adding a bunch more for the other attributes. Happy to accomodate to whatever you folks prefer of course. 😃

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -639,9 +639,6 @@ impl<'a> TcpSocket<'a> {
         self.tx_buffer.clear();
         self.rx_buffer.clear();
         self.rx_fin_received = false;
-        self.keep_alive = None;
-        self.timeout = None;
-        self.hop_limit = None;
         self.listen_address = IpAddress::default();
         self.local_endpoint = IpEndpoint::default();
         self.remote_endpoint = IpEndpoint::default();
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Looks great, thank you!

bors r+

Dirbaio:
oh, could you rustfmt it? Thanks!

mhils:
done, sorry! :)

PR #605, opened 2022-04-29
fix ipv6 contains_addr function in wire

olinex:
The median should not be used as a unit test case


Dirbaio (approved):
Nice, thank you! :)

PR #606, opened 2022-04-30
Bugfix in benchmark example

AntoonBeres:
fixed race condition bug in benchmark example, client thread could be spawned before configuration was complete, resulting in a "ConnectionRefused" error in around 75% of runs on my multithreaded machine

### DIFF
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -81,8 +81,6 @@ fn main() {
         _ => panic!("invalid mode"),
     };
 
-    thread::spawn(move || client(mode));
-
     let neighbor_cache = NeighborCache::new(BTreeMap::new());
 
     let tcp1_rx_buffer = TcpSocketBuffer::new(vec![0; 65535]);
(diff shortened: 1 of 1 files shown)


Dirbaio:
I don't understand how this can fix an error: neither in the old or new place where the thread is spawned is smoltcp running yet. When it runs (the loop below), the socket is added. There's no way for smoltcp to return a `RST`.

Are you sure your network configuration is OK? Are you setting it up according to the [README](https://github.com/smoltcp-rs/smoltcp#hosted-usage-examples)? In particular, you do NOT have to set the IP `192.168.69.1` to `tap0` on the linux side. If you do that, Linux itself could be replying with the RST.

AntoonBeres:
I am 100% sure this fixes the error. I have checked multiple times, the weird thing was that it oncurred on only 1 of my machines and on that machine, the example would actually run correctly occasionally (like 5% of the times), which already implies some kind of race-codnition bug. 

I think that in the old position the spawned thread starts reading/writing immediatly before the buffers and handles are set up properly, thus leading to the error.

Dirbaio:
The thread opens a Linux TCP socket  that talks to smoltcp through the TAP interface. If smoltcp isn't running yet, all the packets Linux sends to the TAP get lost, which causes the Linux TCP socket to wait around and retry sending the SYN. The only way you're getting ConnectionReset is if smoltcp replies with a RST, which is not possible because it's not running yet.

Can you post trace-level logs of the error ocurring, and the output of `ip addr`?

AntoonBeres:
> 

Sorry haven't found the time for this yet, will soon send the details of the error I get

PR #608, opened 2022-05-07
Return tuntap err instead of unwrap

ngc0202:
Ran into a panic today using the `TunTapInterface`, this line turned out to be the culprit. There doesn't appear to be any reason this code should be unwrapping. The `RawSocket` equivalent of this function returns instead of unwrapping.

### DIFF
--- a/src/phy/sys/tuntap_interface.rs
+++ b/src/phy/sys/tuntap_interface.rs
@@ -103,7 +103,7 @@ impl TunTapInterfaceDesc {
                 buffer.len(),
             );
             if len == -1 {
-                Err(io::Error::last_os_error()).unwrap()
+                return Err(io::Error::last_os_error());
             }
             Ok(len as usize)
         }


whitequark:
What is the panic message?

ngc0202:
It panics with a std::io OS error. In my case, I was getting an EIO from the device.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 5, kind: Uncategorized, message: "Input/output error" }', /home/ngc0202/.cargo/registry/src/github.com-1ecc6299db9ec823/smoltcp-0.8.0/src/phy/sys/tuntap_interface.rs:106:49
```

Dirbaio:
I think all possible panics there are due to user misuse (such as removing the tap0 interface while smoltcp is running), but still, propagating all errors instead of panicking as general policy makes sense.

Thanks for the PR!

PR #610, opened 2022-05-09
Add wire representations for SRv6 header as part of IPv6 Routing header

ngc0202:
This is my first of what is likely to be multiple PRs for working with SRv6 with smoltcp! For now, this is just the wire representations. I believe I implemented all the necessary cases, but worth double checking it. I'm also open to suggestions on more areas that could be upgraded to work better with SRv6, and I'll be happy to work on it.


thvdveld:
@Dirbaio can we merge this? I need this for RPL as well.

@ngc0202 are you still working on this after this is merged? If not, I'll continue your since I require it for RPL routing.

ngc0202:
@thvdveld I have a bunch of local changes, but most wasn't really intended to be pushed up. Stuff like some higher-level methods for working with TLVs and the like, I wasn't really sure what was suitable for the PR. I did have a bug in here that I never pushed the fix for however:
```
     /// Variable-length field containing the segment list
     pub const fn SEGMENTS(last_entry_field: u8) -> Field {
-        let bytes = 16 * (last_entry_field as usize) + 8;
+        let bytes = 16 * (last_entry_field as usize) + 24;
         8..bytes
     }
```
Besides that I think it's good to go as is, with whatever changes you see fit.

thvdveld:
I thought I needed this for RPL, but it's source routing that I need (which is already in there). Maybe if you could add some tests for the parsing of SRv6 packets, that would be nice.

thvdveld:
Could you maybe add tests for parsing/emitting? I think that we can merge this after the tests are added.

PR #611, opened 2022-05-12
fix mtu calculation for raw_socket

thvdveld:
Previously, the Ethernet frame header size was added twice to the MTU of the `RawSocket`.
The first time in `interface_mtu` and a second time after calling `interface_mtu` in the raw socket `new` function.
I removed the one in the `interface_mtu` function.


Dirbaio (approved):
Nice catch, thanks!

PR #612, opened 2022-05-12
Bug POC: icmpv4 source address is missing

bunnie:
In the latest version of smoltcp there is an assert on line 2420 of
interfaec.rs which requires that ip_repr.src_adr().is_unspecified() be
false: https://github.com/smoltcp-rs/smoltcp/blob/007f8dd2e2d0a770609428aa4b0efb63f5dbb708/src/iface/interface.rs#L2420

This contradicts the construction of the icmp v4 packet in the dispatch()
function of icmp.rs, where the src_addr is set to an unspecified default: https://github.com/smoltcp-rs/smoltcp/blob/007f8dd2e2d0a770609428aa4b0efb63f5dbb708/src/socket/icmp.rs#L454

This patch papers over the issue by allowing a calling function
to bind the source ipv4 address.

This patch is almost certainly _not_ the way it should be done, but
it serves as a PoC to confirm a fix, and to help reference the problem in a new issue.

### DIFF
--- a/src/socket/icmp.rs
+++ b/src/socket/icmp.rs
@@ -66,6 +66,8 @@ pub struct IcmpSocket<'a> {
     tx_buffer: IcmpSocketBuffer<'a>,
     /// The endpoint this socket is communicating with
     endpoint: Endpoint,
+    /// The originating IP address (for IPv4)
+    src_addr: Ipv4Address,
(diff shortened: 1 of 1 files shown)


bunnie:
closing this as it's related to another issue and 0.8.1 fixes our problem.

PR #614, opened 2022-05-12
Select correct source addr in the UDP socket

thvdveld:
Fixes issue where UDP source address is `UNSPECIFIED`. I saw that this method was used in the TCP socket.


thvdveld:
Closed because we need a better fix!

Issue #599, opened 2022-04-15
IP version mismatch panic


jgallagher:
Hi! I was experimenting with the `master` branch, and code that was working with 0.8.0 is now panicking with something like

```
IP version mismatch: src=Unspecified dst=Ipv4(Address([192, 168, 69, 100]))
```

This is reproducible by running the `server` example on my machine and sending it data. `git bisect` points to 5989896 as the culprit. AFAICT, the chain of events is:

* A `UdpSocket` is created; `self.endpoint` is initialized to `Unspecified:0`.
* We call `.bind(port)` on the socket; `self.endpoint` is now set to `Unspecified:$PORT`.
* When we try to send on that socket, we hit [this panic introduced in that commit](https://github.com/smoltcp-rs/smoltcp/commit/5989896299d0ee03c829a6e600452ebae3b82588#diff-94a7f934fe66af30b7a65e53345991acde51441d760d711fd2c6812ec58e64f2R565).

This can be fixed on the client side by calling `.bind((specific_address, port))` - is that a requirement with these changes? If so, should `bind()` no longer accept only a port?

Dirbaio:
This is definitely not intended, no.

PR #615, opened 2022-05-13
Resolve unspecified source address in IP dispatching

### LINKED ISSUE #613: ICMPv4 outgoing source address missing
In the latest version of smoltcp there is an assert on line 2420 of
interface.rs which requires that ip_repr.src_adr().is_unspecified() be
false: https://github.com/smoltcp-rs/smoltcp/blob/007f8dd2e2d0a770609428aa4b0efb63f5dbb708/src/iface/interface.rs#L2420

This contradicts the construction of the icmp v4 packet in the dispatch()
function of icmp.rs, where the src_addr is set to an unspecified default: https://github.com/smoltcp-rs/smoltcp/blob/007f8dd2e2d0a770609428aa4b0efb63f5dbb708/src/socket/icmp.rs#L454

I opened a PoC PR of a fix at #612. This basically adds a function to specify an IPv4 source and wedge it into the packet. It's not the "right" fix, but I can confirm that with this in place, I no longer trip the assert and pings work correctly.

thvdveld:
`IpRepr::new` now also accepts unspecified source addresses. `dispatch_ip` now checks if the source address is unspecified and tries to resolve it.

This fixes #613 and #599.

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -2416,7 +2416,15 @@ impl<'a> InterfaceInner<'a> {
     }
 
     fn dispatch_ip<Tx: TxToken>(&mut self, tx_token: Tx, packet: IpPacket) -> Result<()> {
-        let ip_repr = packet.ip_repr();
+        let mut ip_repr = packet.ip_repr();
+
+        if ip_repr.src_addr().is_unspecified() {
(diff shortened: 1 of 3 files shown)


Dirbaio:
Closing in favor of #616

PR #617, opened 2022-05-19
Error refactor, part 1

Dirbaio:
Currently `smoltcp` has a single `Error` enum used everywhere. All methods can return all error variants, but in practice they only return some. For example TCP recv can only return `Finished` or `Illegal`, it will never return any of the other 17 possible variants. 

This makes error handling in user code much harder, because you have to check docs for which errors can actually happen.

Goal is
- Make every method return an error enum with only the errors relevant to that function.
- Remove smoltcp::Error

Also, it's likely this will improve code size a bit.


Dirbaio:
This is now mostly done. The only main use left of `Error` is in `phy::Device`, which I'll deal with later.

PR #618, opened 2022-05-20
dns: fix label-length validation

crawford:
The label length is represented by a single byte with the two most-significant bits reserved, which gives a maximum length of 63.


Dirbaio (approved):
oh, nice! Thanks!

bors r+

PR #620, opened 2022-05-25
Change neighbor discovery timeout from 3s to 1s.

lachlansneff-parallel:
As discussed in the matrix chatroom.

### DIFF
--- a/src/iface/socket_meta.rs
+++ b/src/iface/socket_meta.rs
@@ -45,7 +45,7 @@ impl Meta {
     /// in milliseconds.
     ///
     /// See also `iface::NeighborCache::SILENT_TIME`.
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Thank you!

bors r+

crawford:
For those of us not in the chat room (what’s the room ID, by the way?), can you give a summary? Ideally this would be included in the commit description next time so developers can simply consult the git log when blaming changes.

lachlansneff-parallel:
@crawford For sure! Linux retries ARP requests after 1s, so it made sense to match that.

https://matrix.to/#/!RxaQWWMYJTsVdHEKmL:matrix.org/$iHckbToeYPR82Y8-hkAWwy-UPFo2jAzb9IOMAOBr_vk?via=matrix.org&via=rubdos.be&via=braun-odw.eu

PR #624, opened 2022-06-01
IPv4 packet reassembly

thvdveld:
This adds IPv4 packet reassembly. Replacement for #236.

Depends on the fragmentation work of #580. When #580 gets merged I will rebase on master.

### DIFF
--- a/src/iface/fragmentation.rs
+++ b/src/iface/fragmentation.rs
@@ -3,13 +3,12 @@
 use managed::{ManagedMap, ManagedSlice};
 
 use crate::storage::Assembler;
-use crate::time::Instant;
+use crate::time::{Duration, Instant};
 use crate::Error;
 use crate::Result;
 
 /// Holds different fragments of one packet, used for assembling fragmented packets.
 #[derive(Debug)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct PacketAssembler<'a> {
     buffer: ManagedSlice<'a, u8>,
     assembler: AssemblerState,
@@ -17,14 +16,12 @@ pub struct PacketAssembler<'a> {
 
 /// Holds the state of the assembling of one packet.
 #[derive(Debug, PartialEq)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 enum AssemblerState {
     NotInit,
     Assembling {
         assembler: Assembler,
-        total_size: usize,
-        last_updated: Instant,
-        started_on: Instant,
+        total_size: Option<usize>,
+        expires_at: Instant,
         offset_correction: isize,
     },
 }
@@ -51,30 +48,69 @@ impl<'a> PacketAssembler<'a> {
     /// fragments of a packet.
     pub(crate) fn start(
         &mut self,
-        total_size: usize,
-        start_time: Instant,
+        total_size: Option<usize>,
+        expires_at: Instant,
         offset_correction: isize,
     ) -> Result<()> {
         match &mut self.buffer {
-            ManagedSlice::Borrowed(b) if b.len() < total_size => {
-                return Err(Error::PacketAssemblerBufferTooSmall);
+            ManagedSlice::Borrowed(b) => {
+                if let Some(total_size) = total_size {
+                    if b.len() < total_size {
+                        return Err(Error::PacketAssemblerBufferTooSmall);
+                    }
+                }
             }
-            ManagedSlice::Borrowed(_) => (),
             #[cfg(any(feature = "std", feature = "alloc"))]
             ManagedSlice::Owned(b) => {
-                b.resize(total_size, 0);
+                if let Some(total_size) = total_size {
+                    b.resize(total_size, 0);
+                }
             }
         }
 
         self.assembler = AssemblerState::Assembling {
-            assembler: Assembler::new(total_size),
+            assembler: Assembler::new(if let Some(total_size) = total_size {
+                total_size
+            } else {
+                usize::MAX
+            }),
             total_size,
-            last_updated: start_time,
-            started_on: start_time,
+            expires_at,
             offset_correction,
         };
 
         Ok(())
+    }
+
+    /// Set the total size of the packet assembler.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`Error::PacketAssemblerNotInit`] when the assembler was not initialized (try initializing the
+    /// assembler with [Self::start]).
+    pub(crate) fn set_total_size(&mut self, size: usize) -> Result<()> {
+        match self.assembler {
+            AssemblerState::NotInit => Err(Error::PacketAssemblerNotInit),
+            AssemblerState::Assembling {
+                ref mut total_size, ..
+            } => {
+                *total_size = Some(size);
+                Ok(())
+            }
+        }
+    }
+
+    /// Return the instant when the assembler expires.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`Error::PacketAssemblerNotInit`] when the assembler was not initialized (try initializing the
+    /// assembler with [Self::start]).
+    pub(crate) fn expires_at(&self) -> Result<Instant> {
+        match self.assembler {
+            AssemblerState::NotInit => Err(Error::PacketAssemblerNotInit),
+            AssemblerState::Assembling { expires_at, .. } => Ok(expires_at),
+        }
     }
 
     /// Add a fragment into the packet that is being reassembled.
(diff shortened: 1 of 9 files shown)


src/iface/interface.rs
```diff
@@ -1886,21 +1928,78 @@ impl<'a> InterfaceInner<'a> {
...
+                    }
+                };
+
+                if ipv4_packet.frag_offset() != 0 {
+                    check!(f.extend_at(
+                        ipv4_packet.frag_offset() as usize,
+                        ipv4_packet.total_len() as usize - ipv4_packet.header_len() as usize
+                    ));
+                    check!(f.update_ttl(
+                        check!(f.ttl()).max(Duration::from_secs(ipv4_packet.hop_limit() as u64)),
```
Dirbaio: I believe using the hop limit (TTL) as a timeout in seconds is deprecated. See RFC1122 section 3.2.2:

```
There MUST be a reassembly timeout.  The reassembly timeout
value SHOULD be a fixed value, not set from the remaining TTL.
It is recommended that the value lie between 60 seconds and 120
seconds.  If this timeout expires, the partially-reassembled
datagram MUST be discarded and an ICMP Time Exceeded message
sent to the source host (if fragment zero has been received).

DISCUSSION:
The IP specification says that the reassembly timeout
should be the remaining TTL from the IP header, but this
does not work well because gateways generally treat TTL as
a simple hop count rather than an elapsed time.  If the
reassembly timeout is too small, datagrams will be
discarded unnecessarily, and communication may fail.  The
timeout needs to be at least as large as the typical
maximum delay across the Internet.  A realistic minimum
reassembly timeout would be 60 seconds.
```

src/iface/interface.rs
```diff
@@ -1886,21 +1928,78 @@ impl<'a> InterfaceInner<'a> {
...
+                            check!(fragments.get_assembled_packet(&key))
+                        } else {
+                            return None;
+                        }
+                    }
+                    Ok(false) => {
+                        return None;
+                    }
+                    Err(Error::PacketAssemblerOverlap) => {
+                        todo!();
```
Dirbaio: overlap is not an error, should be ignored (keeping either the old or new data)

src/iface/fragmentation.rs
```diff
@@ -71,12 +71,84 @@ impl<'a> PacketAssembler<'a> {
...
+    /// anticipating the extending of the buffer. For example, the offset in the IPv4 header is the
+    /// offset in the total packet. However, when receiving the first packet, we don't know yet
+    /// what the total size will be, but the offset is correct.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`Error::PacketAssemblerNotInit`] when the assembler was not initialized (try initializing the
+    /// assembler with [Self::start]).
+    /// - Returns [`Error::PacketAssemblerBufferTooSmall`] when the buffer is too small.
+    pub(crate) fn extend_at(&mut self, offset: usize, size: usize) -> Result<()> {
```
Dirbaio: I'm not sure what `extend_at` is supposed to do, but I think it won't behave correctly if yuo receive the same fragment twice, as it'll do `*total_size += size` twice.

I think it's possible to simplify the assembling so that you don't need `extends_at` at all though:

- The assembler has
   - `total_size: Option<usize>`, so that the total size  *might be unknown* (this is the new bit)
   - Packet data
- On fragment receive:
   - Add the data
   - If last fragment (`more_frags=0`) then we learned total_size, so set it.
- Assembler is complete when
   - total_size is known, AND
   - 0..total_size is covered by data.

src/wire/ipv4.rs
```diff
@@ -324,6 +331,8 @@ impl<T: AsRef<[u8]>> Packet<T> {
         } else if self.header_len() as u16 > self.total_len() {
             Err(Error)
         } else if len < self.total_len() as usize {
+            net_debug!("len: {}", len);
```
Dirbaio: we shouldn't have debug logs in `wire` ideally

Dirbaio:
Could you feature-gate this behind a new `proto-ipv4-fragmentation` feature? 

No defrag support still works well in practice depending on the application, I'd like to avoid a code size hit to these.

src/iface/interface.rs
thvdveld: Normally, we should send the ICMP Time Exceeded message to the source host only when we received fragment zero. The payload of that ICMP message should be the first 8 octets of fragment zero as well. However, this is currently not possible with my implementation and therefore I send the payload of the last fragment received when the timeout was detected. I'm not sure if this is really a big issue.

thvdveld:
image

src/iface/fragmentation.rs
```diff
@@ -3,28 +3,28 @@
...
     buffer: ManagedSlice<'a, u8>,
     assembler: AssemblerState,
 }
 
 /// Holds the state of the assembling of one packet.
 #[derive(Debug, PartialEq)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 enum AssemblerState {
     NotInit,
+    Discarded,
```
Dirbaio: Why is the new `Discarded` state needed? The distinction between "NotInit" and "Discarded" seem quite subtle, this adds quite a lot of complexity. I think it should be possible to do ipv4 assembling without it. (?)

src/iface/fragmentation.rs
```diff
@@ -3,28 +3,28 @@
...
     Assembling {
         assembler: Assembler,
-        total_size: usize,
+        total_size: Option<usize>,
         last_updated: Instant,
         started_on: Instant,
+        ttl: Duration,
```
Dirbaio: I think storing just one `expires_at: Instant` would simplify the logic a lot.

src/iface/interface.rs
```diff
@@ -1830,22 +1908,105 @@ impl<'a> InterfaceInner<'a> {
...
+                    ));
+                }
+
+                match f.add(
+                    ipv4_packet.payload(),
+                    ipv4_packet.frag_offset() as usize,
+                    self.now,
+                ) {
+                    Ok(true) => {
+                        if !ipv4_packet.more_frags() {
```
Dirbaio: You shouldn't do this check here. If fragments arrive out of order, the last one to arrive might not be the last one in order, so it might have `more_frags = true`.

src/iface/interface.rs
Dirbaio: The payload has to be the first 8 bytes because that's what contains the TCP/UDP port numbers, so the sender can know which socket it's for. It's somewhat important.

Another issue is: The current code sends the TimeExceeded only in response to an incoming fragment. If one fragment arrives and then no more arrive, it won't send the TimeExceeded.

What about doing an "expired check" in `poll()`, before processing any traffic? 
- For each expired assembler
   - If it has data for offset 0..8, send TimeExceeded.
   - Remove it

It'll probably simplify the code as well, since you no longer have to handle expiration in `process_sixlowpan` and `process_ipv4`.

I'm not sure how important are these TimeExceeded messages in practice, I don't think senders do anything with them, the timeout is so long they'll probably have already retried sending by then (?). So not sending TimeExceeded at all is also an option IMO. So is not sending them for now and adding supprt in a later PR. Up to you!

src/iface/fragmentation.rs
thvdveld: So I was playing with the ICMP Time Exceeded message. And I had to store the first 8 bytes somewhere. That somewhere was in the Discarded state, because the `mark_discarded_when` function checked if the first 8 bytes were already received and stored them in `Discarded`. Previous version had `Option<[u8;8]>`. But then it became a real mess to deal with everything and this is just a leftover of me playing with that. But indeed, if we don't do the ICMP thing, then I can remove it.

src/iface/fragmentation.rs
thvdveld: Yes indeed! Will change it.

src/iface/interface.rs
thvdveld: Indeed, removed it.

src/iface/interface.rs
thvdveld: I think not sending a `TimeExceeded` is totally not a problem. TCP would have noticed it already I think, and when you use UDP, then you know it cannot always arrive at the destination. And I don't think other protocols have big payloads.

I can indeed move the check to `poll()`, but this will have a performance hit when we have a lot of fragments not belonging to the same packet (though it is very unlikely this will happen).

Dirbaio: :+1: Nice!

(I don't think it'll be a perf hit, because lookups in ManagedMap are `O(n)`! so even if it "looks" like you're touching just one assembler, you're still iterating all.)

Dirbaio:
LGTM! Thanks for the great work @thvdveld !

bors r+

PR #625, opened 2022-06-02
Add send_with to udp, raw, and icmp sockets

lachlansneff-parallel:
This PR adds a `send_with` method to raw, UDP, and ICMP sockets. These methods enable reserving a packet buffer with a greater size than you need, and then shrinking the size once you know it.

```rust
pub fn send_with<F>(
        &mut self,
        max_size: usize,
        remote_endpoint: IpEndpoint,
        f: F,
    ) -> Result<usize>
    where
        F: FnOnce(&mut [u8]) -> usize,
    { ... }
```

The closure passed to `send_with` should return the size actually written to the supplied buffer.

Internally, this PR adds a `enqueue_with_infallible` method to `PacketBuffer`. While a fallible `enqueue_with` is noted to be extremely complex in a comment on the `enqueue` method, an infallible one is much less complicated. Still, I'd appreciate some more eyes on it to make sure I didn't screw up the logic.

### DIFF
--- a/src/socket/icmp.rs
+++ b/src/socket/icmp.rs
@@ -320,7 +320,7 @@ impl<'a> Socket<'a> {
         max_size: usize,
         endpoint: IpAddress,
         f: F,
-    ) -> Result<&mut [u8], SendError>
+    ) -> Result<usize, SendError>
     where
         F: FnOnce(&mut [u8]) -> usize,
     {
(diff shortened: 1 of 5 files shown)


src/socket/icmp.rs
```diff
@@ -310,6 +310,36 @@ impl<'a> Socket<'a> {
...
+    /// to the provided closure. The closure then returns the size of the data written
+    /// into the buffer.
+    ///
+    /// Also see [send](#method.send).
+    pub fn send_with<F>(
+        &mut self,
+        max_size: usize,
+        endpoint: IpAddress,
+        f: F,
+    ) -> Result<&mut [u8], SendError>
```
Dirbaio: Unlike with `send`, there's no point in returning `&mut [u8]` to the buffer, because the user is supposed to fill it inside `f`.

Dirbaio: This probably simplifies the code a bit, for example you should be able to pass the `f` closure directly to `enqueue_with_infallible` instead of having to wrap it.

lachlansneff-parallel: What about returning the length instead? Obviously, the user could write it to a variable in their closure, but if they need it, returning it from `send_with` would simpler their code.

lachlansneff-parallel: Since the closure passed to `enqueue_with_infallible` returns `(usize, R)`, this change wouldn't simplify the code significantly.

lachlansneff-parallel: Nvm, it does get simpler.

Dirbaio (approved):
Thank you!

bors r+

PR #627, opened 2022-06-10
WIP: Add support for the RPL routing protocol

thvdveld:
This adds the RPL routing protocol from [RFC6550](https://datatracker.ietf.org/doc/html/rfc6550). This routing protocol is designed for Lossy Low-power Networks.

---

- [x] Add wire representation of the RPL headers and the options ([section 6](https://datatracker.ietf.org/doc/html/rfc6550#section-6)).
    - [x] Add field getters
    - [x] Add field setters
    - [x] Add `parse` functions
    - [x] Add `header_len` functions
    - [x] Add `emit` functions
    - [x] Add tests!
- [x] Update the support of extension headers in 6LoWPAN (currently only for Hop By Hop, which is needed in RPL).
    - Currently, `smoltcp` has support for compressed UDP, uncompressed IPv6 headers, but no support for compressed IPv6 extension headers. These headers are mostly carried unmodified in the bytes immediately following the LOWPAN_NHC octet, with two important exceptions: Length field and Next Header field [RFC6282 § 4.2](https://datatracker.ietf.org/doc/html/rfc6282#section-4.2). I saw Contiki-OS using this for their RPL protocol where they used a compressed Hop-by-Hop extension header.
- [x] Add the Trickle timer mechanism from [RFC6202](https://datatracker.ietf.org/doc/html/rfc6206).
    - The Trickle algorithm allows nodes in low-power and lossy networks to exchange information in a highly robust, energy efficient, simple, and scalable manner. Dynamically adjusting transmission windows allows Trickle to spread new information on the scale of link-layer transmission times while sending only a few messages per hour when information does not change. A simple suppression mechanism and transmission point selection allow Trickle's communication rate to scale logarithmically with density.
- [ ] Add Objective Functions:
    - The Objective Function (OF) defines how RPL nodes select and optimize routes within a RPL Instance. The following OFs are supported by Contiki-OS:
    - [x] Objective Function Zero [RFC6552](https://datatracker.ietf.org/doc/html/rfc6552)
    - [ ] Minimum Rank with Hysteresis Objective Function [RFC6719](https://datatracker.ietf.org/doc/html/rfc6719) (Not sure if this will be added in this PR)
- [x] Support for Mode of Operation 0: no downward routes maintained.
- [ ] Support for Mode of Operation 1: upward routes, without storing mode
- [ ] Support for Mode of Operation 2: upward routes, with storing mode (no multicast)
- [ ] Support for Mode of Operation 3: upward routes, with storing mode (multicast)


thvdveld:
Closing this PR since I'm breaking this down into multiple PRs. I have a tracking issue for RPL #769 .

PR #628, opened 2022-06-10
Device-level packet metadata identifiers

datdenkikniet:
A generic way for the user to obtain an ID that allows them to ask a `Device` for information about the frame/packet with which a specific (UDP) packet was sent.

TODO:
- [x] Finalize API
- [x] Update documentation

[HackMD description](https://hackmd.io/EjbOF_X2RNOB8S_BO08u8A)

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -37,7 +37,7 @@ use crate::config::{
     IFACE_MAX_SIXLOWPAN_ADDRESS_CONTEXT_COUNT,
 };
 use crate::iface::Routes;
-use crate::phy::PacketId;
+use crate::phy::PacketMeta;
 use crate::phy::{ChecksumCapabilities, Device, DeviceCapabilities, Medium, RxToken, TxToken};
 use crate::rand::Rand;
 #[cfg(feature = "socket-dns")]
@@ -321,7 +321,7 @@ pub(crate) enum IpPacket<'a> {
     #[cfg(feature = "socket-raw")]
     Raw((IpRepr, &'a [u8])),
     #[cfg(any(feature = "socket-udp", feature = "socket-dns"))]
-    Udp((IpRepr, UdpRepr, &'a [u8], PacketId)),
+    Udp((IpRepr, UdpRepr, &'a [u8], PacketMeta)),
     #[cfg(feature = "socket-tcp")]
     Tcp((IpRepr, TcpRepr<'a>)),
     #[cfg(feature = "socket-dhcpv4")]
@@ -348,11 +348,11 @@ impl<'a> IpPacket<'a> {
         }
     }
 
-    pub(crate) fn packet_id(&self) -> PacketId {
+    pub(crate) fn meta(&self) -> PacketMeta {
         match self {
             #[cfg(feature = "socket-udp")]
-            IpPacket::Udp((_, _, _, packet_id)) => *packet_id,
-            _ => PacketId::empty(),
+            IpPacket::Udp((_, _, _, meta)) => *meta,
+            _ => PacketMeta::default(),
         }
     }
 
@@ -814,14 +814,14 @@ impl Interface {
         let mut processed_any = false;
 
         while let Some((rx_token, tx_token)) = device.receive(self.inner.now) {
-            let rx_packet_id = rx_token.packet_id();
+            let rx_meta = rx_token.meta();
             rx_token.consume(|frame| {
                 match self.inner.caps.medium {
                     #[cfg(feature = "medium-ethernet")]
                     Medium::Ethernet => {
                         if let Some(packet) = self.inner.process_ethernet(
                             sockets,
-                            rx_packet_id,
+                            rx_meta,
                             &frame,
                             &mut self.fragments,
                         ) {
@@ -834,12 +834,10 @@ impl Interface {
                     }
                     #[cfg(feature = "medium-ip")]
                     Medium::Ip => {
-                        if let Some(packet) = self.inner.process_ip(
-                            sockets,
-                            rx_packet_id,
-                            &frame,
-                            &mut self.fragments,
-                        ) {
+                        if let Some(packet) =
+                            self.inner
+                                .process_ip(sockets, rx_meta, &frame, &mut self.fragments)
+                        {
                             if let Err(err) =
                                 self.inner
                                     .dispatch_ip(tx_token, packet, &mut self.fragmenter)
@@ -852,7 +850,7 @@ impl Interface {
                     Medium::Ieee802154 => {
                         if let Some(packet) = self.inner.process_ieee802154(
                             sockets,
-                            rx_packet_id,
+                            rx_meta,
                             &frame,
                             &mut self.fragments,
                         ) {
@@ -944,7 +942,10 @@ impl Interface {
                 #[cfg(feature = "socket-dns")]
                 Socket::Dns(socket) => {
                     socket.dispatch(&mut self.inner, |inner, (ip, udp, payload)| {
-                        respond(inner, IpPacket::Udp((ip, udp, payload, PacketId::empty())))
+                        respond(
+                            inner,
+                            IpPacket::Udp((ip, udp, payload, PacketMeta::default())),
+                        )
                     })
                 }
             };
@@ -1284,7 +1285,7 @@ impl InterfaceInner {
     fn process_ip<'frame, T: AsRef<[u8]>>(
         &mut self,
         sockets: &mut SocketSet,
-        packet_id: PacketId,
+        meta: PacketMeta,
         ip_payload: &'frame T,
         frag: &'frame mut FragmentsBuffer,
     ) -> Option<IpPacket<'frame>> {
@@ -1293,12 +1294,12 @@ impl InterfaceInner {
             Ok(IpVersion::Ipv4) => {
                 let ipv4_packet = check!(Ipv4Packet::new_checked(ip_payload));
 
-                self.process_ipv4(sockets, packet_id, &ipv4_packet, frag)
+                self.process_ipv4(sockets, meta, &ipv4_packet, frag)
             }
             #[cfg(feature = "proto-ipv6")]
             Ok(IpVersion::Ipv6) => {
                 let ipv6_packet = check!(Ipv6Packet::new_checked(ip_payload));
-                self.process_ipv6(sockets, packet_id, &ipv6_packet)
+                self.process_ipv6(sockets, meta, &ipv6_packet)
             }
             // Drop all other traffic.
             _ => None,
@@ -1367,7 +1368,7 @@ impl InterfaceInner {
     fn process_udp<'frame>(
         &mut self,
         sockets: &mut SocketSet,
-        packet_id: PacketId,
+        meta: PacketMeta,
         ip_repr: IpRepr,
         udp_repr: UdpRepr,
         handled_by_raw_socket: bool,
@@ -1380,7 +1381,7 @@ impl InterfaceInner {
             .filter_map(|i| udp::Socket::downcast_mut(&mut i.socket))
         {
             if udp_socket.accepts(self, &ip_repr, &udp_repr) {
-                udp_socket.process(self, packet_id, &ip_repr, &udp_repr, udp_payload);
+                udp_socket.process(self, meta, &ip_repr, &udp_repr, udp_payload);
                 return None;
             }
         }
@@ -1834,7 +1835,7 @@ impl InterfaceInner {
                         Ok(())
                     }
                 } else {
-                    tx_token.set_packet_id(packet.packet_id());
+                    tx_token.set_meta(packet.meta());
 
                     // No fragmentation is required.
                     tx_token.consume(total_len, |mut tx_buffer| {
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -37,6 +37,7 @@ use crate::config::{
     IFACE_MAX_SIXLOWPAN_ADDRESS_CONTEXT_COUNT,
 };
 use crate::iface::Routes;
+use crate::phy::PacketId;
 use crate::phy::{ChecksumCapabilities, Device, DeviceCapabilities, Medium, RxToken, TxToken};
 use crate::rand::Rand;
 #[cfg(feature = "socket-dns")]
@@ -320,7 +321,7 @@ pub(crate) enum IpPacket<'a> {
     #[cfg(feature = "socket-raw")]
     Raw((IpRepr, &'a [u8])),
     #[cfg(any(feature = "socket-udp", feature = "socket-dns"))]
-    Udp((IpRepr, UdpRepr, &'a [u8])),
+    Udp((IpRepr, UdpRepr, &'a [u8], PacketId)),
     #[cfg(feature = "socket-tcp")]
     Tcp((IpRepr, TcpRepr<'a>)),
     #[cfg(feature = "socket-dhcpv4")]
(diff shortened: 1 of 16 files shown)


examples/multicast.rs
```diff
@@ -63,7 +63,15 @@ fn main() {
     let raw_handle = sockets.add(raw_socket);
 
     // Must fit mDNS payload of at least one packet
-    let udp_rx_buffer = udp::PacketBuffer::new(vec![udp::PacketMetadata::EMPTY; 4], vec![0; 1024]);
```
thvdveld: Is there a reason why you changed this?

src/phy/fault_injector.rs
```diff
@@ -210,36 +212,42 @@ where
...
+                    config,
+                    token: tx_token,
+                    junk: [0; MTU],
+                };
+                (rx, tx)
+            })
     }
 
-    fn transmit(&'a mut self) -> Option<Self::TxToken> {
+    fn transmit(&'a mut self, timestamp_id: PacketId) -> Option<Self::TxToken> {
```
thvdveld: Maybe we should just call it `packet_id` instead of `timestamp_id`?

thvdveld: Well, probably it's just something that you missed changing.

src/phy/mod.rs
```diff
@@ -132,6 +137,33 @@ pub use self::tracer::Tracer;
 ))]
 pub use self::tuntap_interface::TunTapInterface;
 
+/// An ID that can be used to uniquely identify a packet to a [`Device`],
+/// sent or received by that same [`Device`]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Debug, PartialEq)]
+pub struct PacketId {
+    id: usize,
+}
```
thvdveld: Maybe we could write it as:
```rust
pub struct PacketId(usize);
```

src/phy/mod.rs
```diff
@@ -132,6 +137,33 @@ pub use self::tracer::Tracer;
...
+
+impl PacketId {
+    pub const fn none() -> Option<Self> {
+        None
+    }
```
thvdveld: Not sure if this belongs here. If we really need a `None` then we can just write `None` right? Is there a case where you need this maybe?

src/phy/mod.rs
```diff
@@ -132,6 +137,33 @@ pub use self::tracer::Tracer;
...
+
+    pub fn id(&self) -> usize {
+        self.id
+    }
+
+    pub(crate) fn copy(&self) -> Self {
+        Self { id: self.id }
+    }
```
thvdveld: I think we can make `PacketId` `Copy`. Having a unique value for the `PacketId` is up to the user IMO.

src/phy/mod.rs
```diff
@@ -132,6 +137,33 @@ pub use self::tracer::Tracer;
...
+
+    /// CAUTION: don't use too often...
```
thvdveld: What is the reason of the `CAUTION`?

src/phy/mod.rs
```diff
@@ -243,6 +275,8 @@ pub struct DeviceCapabilities {
     /// If the network device is capable of verifying or computing checksums for some protocols,
     /// it can request that the stack not do so in software to improve performance.
     pub checksum: ChecksumCapabilities,
+
+    pub hardware_timestamping: bool,
```
thvdveld: What is the `hardware_timestamping` field used for?

thvdveld (commented):
Just some questions.

examples/multicast.rs
datdenkikniet: Because `PacketMetadata` is (currently) no longer `Copy` because of the contained `PacketId`

src/phy/fault_injector.rs
datdenkikniet: Ah, yes that's a change I forgot to make

src/phy/mod.rs
datdenkikniet: Yes.

Personally, I'm a bigger fan of structs with named fields but it seems much more in line with the rest of the lib (and it's probably more idiomatic?) to use a unit struct here. I'll change it.

src/phy/mod.rs
datdenkikniet: Similar to above, because `PacketMetadata` is no longer `Copy`, the `const fn` makes it "easier" to just call `PacketMetadata::none()`. Not so user-friendly, and `PacketMetadata` will probably become `Copy` again, in which case this can be removed.

src/phy/mod.rs
datdenkikniet: I agree. Initially it made sense to me to have it be non-`Copy`, but given that the only way problems can arise is when the `id` rolls over, it doesn't make any practical sense.

I'll make `PacketId` copy and update all the other issues/changes that occurred because of it,

src/phy/mod.rs
datdenkikniet: From matrix: 

Perhaps a slightly too placeholder-y comment hahaha. Wanted to somehow indicate that obtaining a `PacketId` should generally not be done by calling `PacketId::new`, but rather from an `Interface` or `RawSocket`

src/phy/mod.rs
datdenkikniet: Oops, that's a bit of a leftover from experimenting. If `smoltcp` ever gets "built-in" (through something like a `PtpSocket`) support for timestamping, allowing `Device`s to indicate that they have support for hardware timestamping is probably desirable. Not relevant for this PR though, so I'll remove this field.

datdenkikniet:
Thank you for the comments @thvdveld ! Left some responses/feedback to what you've said. Will make the changes sometimes this week.

korken89:
Hey @Dirbaio / @thvdveld - has there been any progress here or more questions/comments?

src/iface/interface/mod.rs
```diff
@@ -720,7 +724,7 @@ pub(crate) enum IpPacket<'a> {
     #[cfg(feature = "socket-raw")]
     Raw((IpRepr, &'a [u8])),
     #[cfg(any(feature = "socket-udp", feature = "socket-dns"))]
-    Udp((IpRepr, UdpRepr, &'a [u8])),
+    Udp((Option<PacketId>, IpRepr, UdpRepr, &'a [u8])),
```
Dirbaio: I think it makes more sense to pass the ID alongside the `IpPacket`, not within. Any packet could get an ID, independently of whether it's UDP or not.

src/iface/interface/mod.rs
```diff
@@ -1275,9 +1291,14 @@ impl<'a> Interface<'a> {
                     _ => unreachable!(),
                 }),
                 #[cfg(feature = "socket-udp")]
-                Socket::Udp(socket) => socket.dispatch(inner, |inner, response| {
-                    respond(inner, IpPacket::Udp(response))
-                }),
+                Socket::Udp(socket) => {
+                    socket.dispatch(inner, |inner, (rx_packet_id, ip_repr, udp_repr, data)| {
+                        respond(
```
Dirbaio: `respond` always makes up its own ID, AFAICT the `rx_packet_id` that gets stuffed into `IpPacket::Udp` is not used (?)

Dirbaio:
Sorry for the :snail: response!

- I think the ID should be probably a `u32`, so that it doesn't change between 32bit and 64bit platforms.
- Ideally, the feature shouldn't add bloat for those who don't need it. Could you gate it behind a Cargo feature? Perhaps define `PacketId` to be `()`  when not enabled, and `u32` when enabled. I hope this trick works, it'd prevent most of the `cfg` spam while having the same results, because passing around a `()` should generate no code.
- Currently the IDs are all assigned by smoltcp incrementally. I wonder if instead letting the user assign them for TX, and the Device for RX, would be more flexible, or allow more use cases. I think it should still cover PTP fine, but it could allow other interesting things like using some bits flags, or encoding stuff like TX queue priority, or RSSI for wireless stuff. What do you think?

Also, do you have a working impl of PTP on top of this? (not a requirement for merge! I'm just curious, would like to have a look if you do have it)

src/iface/interface/mod.rs
datdenkikniet: Hmm, yes. I seem to have gotten some wires crossed while merging, this thing is entirely unnecessary. I've removed it.

src/iface/interface/mod.rs
datdenkikniet: Yep, not sure what happened here. Same as [above](https://github.com/smoltcp-rs/smoltcp/pull/628#discussion_r1059975699).

datdenkikniet:
* Changing it to a fixed-size int would make sense, yeah. A `u32` is probably big enough? At any rate it'd be relatively easy to change in the future.
* Yes, feature-gating it sounds like a good plan. I've added it as per your suggestion, seems to work pretty well. When adding a relatively simple feature like this, how many tests should I add? I've only included it in the "as aggressive as possible" test for now.
* Hmm, that _would_ be pretty nice. If we could somehow offer that as an alternative that would be quite nice, I think having an easy default (of being just a bog-standard counter) is quite important. Maybe just an additional `send_manually_marked` or something like that would work?

We have a "working" PTP implementation [here](https://github.com/stm32-rs/stm32-eth/pull/46), though I'm not 100% sure that still compiles. "Working" because you can only get out the timestamps so far, we're not actually doing any PTP syncing just yet.

Dirbaio:
> A u32 is probably big enough? At any rate it'd be relatively easy to change in the future.

u32 sounds good to me.

> how many tests should I add? I've only included it in the "as aggressive as possible" test for now.

Don't worry too much about it, since there's no complex logic added, just passing a value around.

> If we could somehow offer that as an alternative that would be quite nice, I think having an easy default (of being just a bog-standard counter) is quite important. Maybe just an additional send_manually_marked or something like that would work?

I think having both is not useful, I think we do need to choose now. If we do both, when the phy gets a packet with some ID, it can't know whether it's the counter value or a user-assigned special value, so it can't act on it on any special way. This defeats the point. If we want setting the packet mark to be useful, all other packets should default to a mark of 0, not to a counter.

Also, assigning counter values to regular packets (ie those not sent with `send_marked`) is not useful, because the user can't obtain these IDs, so they can't use the received IDs on the phy side to correlate with anything. Same for received packets.

Another interesting thing is, if you have lots of traffic in addition to the PTP packets, you're timestamping all packets, so you have to keep a big buffer of id->timestamp mappings or by the time you grab the timestamp it might be already gone. I wonder if this is a problem in practice? For TX, with custom marks you could do something like "0 means don't timestamp, 1 means timestamp" to tell the PHY which ones to timestamp, but this doesn't solve the problem for RX...

I'm not sure what's best :sweat_smile: .

going back on the design discussions: for rx'd packets, simply having the PHY return the timestamp and thread that through all layers up to the socket solves all these issues. It doesn't for tx'd packets though, since you don't know the timestamp after the tx is done, and we don't want `TxToken::consume()` to block until it's done. So we do need some kind of "packet ID" to retrieve it later... I wonder if we're overcomplicating this by sticking to just IDs, instead of some "packet metadata" struct that we can pass around that can contain IDs, timestamps, other stuff...?

datdenkikniet:
> I think having both is not useful, I think we do need to choose now. If we do both, when the phy gets a packet with some ID, it can't know whether it's the counter value or a user-assigned special value, so it can't act on it on any special way. This defeats the point. If we want setting the packet mark to be useful, all other packets should default to a mark of 0, not to a counter.

The initial idea was that we could just have some value to uniquely map packets to some physical frame. This sounds like we'd  be stepping away from that and turning it into a value that can encode some value that only use code and the PHY driver can use properly. If that is what we want to end up with, we should rename it and somehow properly document that.

> Also, assigning counter values to regular packets (ie those not sent with `send_marked`) is not useful, because the user can't obtain these IDs, so they can't use the received IDs on the phy side to correlate with anything. Same for received packets.

While true, it simplifies how handling the IDs (since there will always be one). Maybe too simple, but unless having an ID present somehow induces a bunch of overhead I don't think it makes a large difference.

> Another interesting thing is, if you have lots of traffic in addition to the PTP packets, you're timestamping all packets, so you have to keep a big buffer of id->timestamp mappings or by the time you grab the timestamp it might be already gone. I wonder if this is a problem in practice? For TX, with custom marks you could do something like "0 means don't timestamp, 1 means timestamp" to tell the PHY which ones to timestamp, but this doesn't solve the problem for RX...

In STM32s, the timestamping is a zero-cost additional thing that the MCU can do "for free" (only costing a little bit of memory). The original idea assumes that all additional information/operations you can perform using this ID are also zero-cost in that sense.

In `stm32-eth` we only guarantee that the timestamp can be retrieved within one "interrupt cycle", so the amount of timestamps (= 64 bits) to buffer is limited to the total amount of ethernet frames that the DMA can handle.

> going back on the design discussions: for rx'd packets, simply having the PHY return the timestamp and thread that through all layers up to the socket solves all these issues. It doesn't for tx'd packets though, since you don't know the timestamp after the tx is done, and we don't want `TxToken::consume()` to block until it's done. So we do need some kind of "packet ID" to retrieve it later... I wonder if we're overcomplicating this by sticking to just IDs, instead of some "packet metadata" struct that we can pass around that can contain IDs, timestamps, other stuff...?

You could do this using (an) associated type(s) on `RxToken` and `TxToken` and including that data (probably generated by `Device`, then) instead of the `PacketId`, assuming adding generics to sockets that would want to make use of such associated data is OK.

Otherwise `smoltcp` would have to decide the metadata that can or can't be included, which doesn't sound like a super good idea.

In the end I suppose we'd sooner end up with something closer to

```rust
pub trait TxToken {
    type Metadata;
    
    fn consume<R, F>(self, timestamp: Instant, len: usize, f: F) -> Result<(R, Self::Metadata)> 
}

pub trait RxToken {
    type Metadata;

    fn consume<R, F>(self, timestamp: Instant, f: F) -> Result<(R, Self::Metadata)>
        
}
```

and

```rust
pub struct UdpMetadata<T> {
    endpoint: IpEndpoint,
    packet_meta: T, // T = Device::RxToken::Metadata/Device::TxToken::Metadata
}
```

which honestly would be a much smoother solution. Then potentially letting the user generate some custom metadata value that sets some flags for one thing or the other would fall entirely on the implementor of `Device`. I'd be in favor of a solution like that.

Edit: ah, or some sort of associated type on `Device`, but with the same idea

korken89:
Hi @Dirbaio, do you have any more comments/questions on this?

korken89:
Hi @Dirbaio, I'd like to check in on how this is progressing on your side?

korken89:
Just for completeness, we have a PTP PoC that here using this PR: https://github.com/stm32-rs/stm32-eth/tree/smoltcp-ptp

Checkout `examples/server.rs` and `examples/client.rs`

Dirbaio:
I think the "assign IDs incrementally vs let user/device assign them concern" concern I initially raised is still unsolved. You mention:

> In stm32-eth we only guarantee that the timestamp can be retrieved within one "interrupt cycle", so the amount of timestamps (= 64 bits) to buffer is limited to the total amount of ethernet frames that the DMA can handle.

This does work if both of these are true:
1- The amount of packets processed in a single poll is bounded, and
2- The user is doing manual polling, so they can ensure you do one PTP poll in between each iface/device poll.

`1` might be true in `stm32-eth`, but it's not necessarily true in general. For example in the `embassy-stm32` implementation, if the CPU is slow enough wrt DMA, it is possible for the same descriptor to be tx'd multiple times in the same poll if DMA has sent it and freed it by the time you try to tx again. This makes a single poll able to handle an unbounded amount of packets.

About `2`, there's many ways to use smoltcp that don't involve manual polling, such as with `async`, for example like `embassy-net` does.  I don't think manual polling should be required for correct usage. Smoltcp should *support* manual polling if the user wants to do it, sure, but we shouldn't have an API that can be used correctly *only* with manual polling.

datdenkikniet:
That is a good point, thank you for the feedback. That use case lends itself to too much of a niche (manual polling).

I'll rework it a bit to see if I can find an elegant way of supporting optional packet IDs provided by the device

datdenkikniet:
I'm redoing it now so that we leverage the `Device` much more when it comes to setting up `PacketId`'s. Additionally, it means we don't actually have to break the `Device` trait.

I've reset the tree to `master` before performing these commits (since the approach is so different), but haven't dropped the old just yet.

korken89:
Hi @Dirbaio 

The requested fixed and updates were pushed a few months ago. Do you have any feedback or is it ready for merge?

PR #630, opened 2022-06-13
Fix 6LoWPAN address compression

### LINKED ISSUE #629: 6LowPan ICMPv6 NDISC errors
I'm trying to get two microcontrollers to talk to each other over 805.15.4/6LowPan, both assigned static **site**-local IP addresses derived from the device's hardware address. I'm using site-local instead of link-local because I would like them to be routable through a border gateway. I'm attempting to send raw bytes over a udp socket to the other device. Because the device isn't in the neighbor cache, it appropriately triggers a ICMPv6 NDISC request, but after that things look a little weird to me.

The ICMPv6 type being output is 0, which I don't think is a valid type. This is from the `c4791eb668c421f941c4c489ba1a402162b7be93` commit of upstream.

```
TRACE - udp:fd00:1eaf::a08d:cfd3:fffe:bde1:7:fd00:1eaf::5ad3:fcc:fffe:7391:7: buffer to send 2 octets
TRACE - udp:fd00:1eaf::a08d:cfd3:fffe:bde1:7:fd00:1eaf::5ad3:fcc:fffe:7391:7: sending 2 octets
DEBUG - address fd00:1eaf::5ad3:fcc:fffe:7391 not in neighbor cache, sending Neighbor Solicitation
```

Here is the PCAP file from the sending side, renamed to a txt for github upload. You might need to disable the "Disect only good FCS" in the 802.15.4 protocol preferences since my radio peripheral calculates that after this is dumped.
[gateway.txt](https://github.com/smoltcp-rs/smoltcp/files/8886980/gateway.txt)

thvdveld:
Should fix #629. The `Source Address Mode (SAM)` in the 6LoWPAN `IPHC` header was set to `Compress` (`0b11`), while it should have been `Inline` (`0b00`) when we inline the full IPv6 address.

### DIFF
--- a/src/wire/sixlowpan.rs
+++ b/src/wire/sixlowpan.rs
@@ -910,7 +910,7 @@ pub mod iphc {
         ) -> usize {
             self.set_cid_field(0);
             self.set_sac_field(0);
-            self.set_sam_field(0b11);
+            self.set_sam_field(0b00);
             let src = src_addr.as_bytes();
             if src_addr == ipv6::Address::UNSPECIFIED {
                 self.set_sac_field(1);
@@ -958,6 +958,7 @@ pub mod iphc {
                 }
             } else {
                 // We cannot elide anything.
+                self.set_sam_field(0b00);
                 self.set_field(idx, src);
                 idx += 16;
             }


Dirbaio (approved):
Nice!

bors r+

src/wire/sixlowpan.rs
```diff
@@ -910,7 +910,7 @@ pub mod iphc {
         ) -> usize {
             self.set_cid_field(0);
             self.set_sac_field(0);
-            self.set_sam_field(0b11);
+            self.set_sam_field(0b00);
```
jordens, line 913: This should be removed, right? Every path below sets it.
@thvdveld

thvdveld: Yes indeed, however, I think we can leave it. It's better to first set everything to zero before modifying the fields. And the compiler will optimize that away, probably.

jordens: I wouldn't rely on that. And even then you can remove it in both L917 and L961.
Anyway. Just a minor nit.

thvdveld: Removed it in 75bacb1e46b5e5a0fa0c758ebbc7bbc963fb5bf3 just to be sure :wink:

PR #633, opened 2022-06-16
Modify `poll_at` for 6LoWPAN outgoing fragments

thvdveld:
When there are still 6LoWPAN fragments to be transmitted, `poll_at` should return `Some(Instant::from_millis(0))`.


Dirbaio (approved):
Thank you!

PR #635, opened 2022-06-24
Add additional DHCP configurability

lachlansneff-parallel:
This PR is based on @Luis-Hebendanz's fork of smoltcp that adds some initial support for PXE to DHCP sockets.

From there, I've generalized it, removing most PXE specific code, but allowing users to send and receive custom DHCP options.

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -2,9 +2,10 @@ use crate::{
     iface::Context,
     time::{Duration, Instant},
     wire::{
-        dhcpv4::field as dhcpv4_field, DhcpMessageType, DhcpPacket, DhcpRepr, HardwareAddress,
-        IpAddress, IpProtocol, Ipv4Address, Ipv4Cidr, Ipv4Repr, PxeMachineId, UdpRepr,
-        DHCP_CLIENT_PORT, DHCP_MAX_DNS_SERVER_COUNT, DHCP_SERVER_PORT, UDP_HEADER_LEN,
+        dhcpv4::{field as dhcpv4_field, DhcpOptionsRepr},
+        DhcpMessageType, DhcpOption, DhcpPacket, DhcpRepr, HardwareAddress, IpAddress, IpProtocol,
+        Ipv4Address, Ipv4Cidr, Ipv4Repr, UdpRepr, DHCP_CLIENT_PORT, DHCP_MAX_DNS_SERVER_COUNT,
+        DHCP_SERVER_PORT, UDP_HEADER_LEN,
     },
     Error, Result,
 };
@@ -91,31 +92,6 @@ enum ClientState {
     Renewing(RenewState),
 }
 
-#[derive(Debug)]
-pub struct PxeBuffers {
-    id: [u8; 16],
-    bootfile_name_len: usize,
-    bootfile_name_buf: [u8; 128],
-}
-
-impl PxeBuffers {
-    pub const fn new(id: [u8; 16]) -> Self {
-        Self {
-            id,
-            bootfile_name_len: 0,
-            bootfile_name_buf: [0; 128],
-        }
-    }
-
-    pub fn id(&self) -> &[u8; 16] {
-        &self.id
-    }
-
-    pub fn id_mut(&mut self) -> &mut [u8; 16] {
-        &mut self.id
-    }
-}
-
 /// Timeout and retry configuration.
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
@@ -172,7 +148,10 @@ pub struct Dhcpv4Socket<'a> {
     /// Ignore NAKs.
     ignore_naks: bool,
 
-    pxe_buffers: Option<&'a mut PxeBuffers>,
+    /// A buffer for the bootfile name to be written to.
+    bootfile_name_buffer: Option<(usize, &'a mut [u8])>,
+    outbox_options: Option<DhcpOptionsRepr<&'a [u8]>>,
+    inbox_options: Option<DhcpOptionsRepr<&'a mut [u8]>>,
 }
 
 /// DHCP client socket.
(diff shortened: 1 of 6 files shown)


lachlansneff-parallel:
Before I fix the tests, I would appreciate if someone could take a look at this and tell me whether the maintainers would be willing to merge a PR with this approach.

Dirbaio:
You seem to have some tool going on that automatically reflows (?) text in doc comments. Could you revert that? it makes the diff unnecessarily big.

src/socket/dhcpv4.rs
```diff
@@ -256,6 +323,20 @@ impl Socket {
...
+        {
+            let len = bootfile_name.len().min(bootfile_name_buffer.len());
+            bootfile_name_buffer[..len].copy_from_slice(&bootfile_name.as_bytes()[..len]);
+            *bootfile_name_len = len;
+        }
+
+        if let (Some(inbox_options), Some(received_options)) =
+            (self.inbox_options.as_mut(), dhcp_repr.options)
+        {
+            let _ = inbox_options.emit(received_options.parse());
```
Dirbaio, line 338: this could be a memcpy instead of parse+emit

src/socket/dhcpv4.rs
```diff
@@ -151,7 +185,38 @@ impl Socket {
...
+            outbox_options: None,
+            inbox_options: None,
+            parameter_request_list: None,
+            #[cfg(feature = "async")]
+            waker: WakerRegistration::new(),
+        }
+    }
+
+    /// Create a DHCPv4 socket with retry and DHCP option configuration.
+    pub fn with_config(
```
Dirbaio: instead of an alternate constructor, can you add getters/setters for the configurable things? This is what all other sockets do, see TCP for example.

src/socket/dhcpv4.rs
```diff
@@ -104,32 +99,71 @@ enum ClientState {
...
     transaction_id: u32,
 
-    /// Max lease duration. If set, it sets a maximum cap to the server-provided lease duration.
-    /// Useful to react faster to IP configuration changes and to test whether renews work correctly.
+    /// Max lease duration. If set, it sets a maximum cap to the server-provided
+    /// lease duration. Useful to react faster to IP configuration changes
+    /// and to test whether renews work correctly.
     max_lease_duration: Option<Duration>,
 
+    retry_config: &'a RetryConfig,
```
Dirbaio: this can be owned instead of a reference

src/socket/dhcpv4.rs
```diff
@@ -104,32 +99,71 @@ enum ClientState {
...
+pub struct RetryConfig {
+    pub discover_timeout: Duration,
+    /// The REQUEST timeout doubles every 2 tries.
+    pub initial_request_timeout: Duration,
+    pub request_retries: u16,
+    pub min_renew_timeout: Duration,
+}
+
+impl RetryConfig {
+    pub const DEFAULT: Self = Self {
```
Dirbaio: can you change this so there's only the Default impl? otherwise there's 2 ways of doing  the same thing in user code, which is not nice: `RetryConfig::default()` and `RetryConfig::DEFAULT`

Dirbaio:
Approach LGTM overall, other than the comments.

About incoming bootfile+options: I wonder if it'd be easier to buffer the entire DHCP ACK packet instead of invidivual fields. This way it's possible for user code to parse whatever they need out of the packet without the socket having to add explicit support for it. For example, currently you needed to read the bootfile so you added support for that, which means an extra field and extra code, which is "wasted space" for people who don't need the bootfile. (and no, we can't add Cargo features for each field). If you buffer the whole packet instead, users can parse whatever, so the code size and the socket struct don't grow as we add support for more fields.

Qubasa:
Awesome! Thanks @lachlansneff-parallel for polishing it up  :-)

src/socket/dhcpv4.rs
lachlansneff-parallel: I don't think it could be, since we need to emit as many options that will fit without truncating them.

src/socket/dhcpv4.rs
lachlansneff-parallel: Done.

src/socket/dhcpv4.rs
lachlansneff-parallel: Done.

src/socket/dhcpv4.rs
lachlansneff-parallel: Had this because I needed a const constructor, but that's no longer the case, so removed.

lachlansneff-parallel:
@Dirbaio Storing the entire DHCP packet definitely seems like it could be the right path. The reason why there needed to be a bootfile field is because it's actually part of the DHCP packet, not a DHCP option. Most other things that could be fields are options and won't need explicit fields in the DHCP repr.

src/wire/dhcpv4.rs
```diff
@@ -55,6 +57,58 @@ impl MessageType {
     }
 }
 
+/// A buffer for DHCP options.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct DhcpOptionsRepr<T> {
```
Dirbaio: this is not a "repr". Reprs are Rust structs with everything neatly parsed into fields, this wraps an unparsed buffer. Perhaps name it "DhcpOptionsBuffer"?

lachlansneff-parallel: Agreed. Renamed to `DhcpOptionsBuffer`.

lachlansneff-parallel:
@Dirbaio Thoughts on moving this forward? We've been using the code in this PR in our systems for several weeks and they've been working well.

Dirbaio:
@lachlansneff-parallel 

- You've committed doc comment reflows again
- As discussed previously, please change it to buffer the entire DHCP ACK packet so that users can parse whatever they need out of it:
   - No need for bootp-specific fields and API.
   - No need to parse + reemit options, just do a memcpy.

lachlansneff-parallel:
Apologies for the doc reflows, it's difficult to remember to turn off `cargo fmt` on save based on repo — will fix.

I will go and change it to buffer the DHCP packet. My job ends at the end of this week, so I'll try to finish the change by then, if not I will pass it to a coworker.

lachlansneff-parallel:
I'm going to be closing this PR and reopening it from a different repository to transition the PR to a coworker.

PR #637, opened 2022-07-01
Make `send_with` on UDP, raw, and ICMP sockets fallible.

lachlansneff-parallel:
This PR adds `enqueue_with` to the `PacketBuffer` type that supports allocating a buffer in a fallible manner. If the provided closure returns an `Err`, then the changes to the payload and metadata rings are rolled back.

Also, the `send_with` function available on UDP, raw, and ICMP sockets has been changed slightly to allow for fallible closures.

### DIFF
--- a/src/socket/udp.rs
+++ b/src/socket/udp.rs
@@ -266,7 +266,6 @@ impl<'a> Socket<'a> {
     pub fn send_with<F, E>(
         &mut self,
         at_least: usize,
-        request: usize,
         remote_endpoint: IpEndpoint,
         f: F,
     ) -> Result<Result<usize, E>, SendError>
@@ -285,7 +284,7 @@ impl<'a> Socket<'a> {
 
         let res = self
             .tx_buffer
-            .enqueue_with(at_least, request, remote_endpoint, f)
+            .enqueue_with(at_least, remote_endpoint, f)
             .map_err(|_| SendError::BufferFull)?;
 
         if let Ok(size) = res {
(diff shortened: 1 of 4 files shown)


src/storage/packet_buffer.rs
```diff
@@ -111,6 +108,76 @@ impl<'a, H> PacketBuffer<'a, H> {
         Ok(payload_buf)
     }
 
+    /// Call `f` with a packet from the buffer that's at least `at_least` bytes.
+    ///
+    /// If `f` returns `Ok(size)`, the packet is shrunk to `size` and enqueued.
+    /// If `f` returns `Err`, the internal state of the packet buffer is rolled back and no data is enqueued.
+    pub fn enqueue_with<F, E>(
```
Dirbaio, line 115: `enqueue_with_infallible` is now unused, could you remove it?

src/socket/udp.rs
```diff
@@ -258,19 +258,19 @@ impl<'a> Socket<'a> {
...
+    /// Enqueue a packet to be sent to a given remote endpoint and pass the buffer
+    /// to the provided closure. The closure then returns the either size of the data written
+    /// into the buffer or an error. If an error is returned, no packet is enqueued.
     ///
     /// Also see [send](#method.send).
-    pub fn send_with<F>(
+    pub fn send_with<F, E>(
         &mut self,
-        max_size: usize,
+        at_least: usize,
```
Dirbaio, line 268: I like the old `max_size` name better. The new one sounds like the closure will send "at least X bytes", ie it's a "min size", which is not the case. Could you change it back?

lachlansneff-parallel:
It turns out I don't have a usecase for this anymore.

Issue #640, opened 2022-07-15
6LowPan outgoing fragmentation error


benbrittain:
I have an echo server running on a microcontroller, it successfully reads in fragmented 6LowPan packets and assembles them, but when I turn around to blast them out again, I get what looks like a never ending stream of spurious fragments?

Wireshark shows the actual message contents in the first two packets (#22 & #23), then a repeating write of packets of length 26.

![image](https://user-images.githubusercontent.com/207432/179264640-0039faa2-bbca-4e9a-abd7-5f844ed37981.jpg)

benbrittain:
The phy device is operating as a null sink during the time we are sending the outgoing fragments, always consuming the packet to help isolate the problem. I've attached a wireshark txt export so you can see the actual packets.

The Bad FCS errors can be ignored, on the wire they have the right checksum since it's handled by the radio peripheral I'm using.

[raw-out.txt](https://github.com/smoltcp-rs/smoltcp/files/9122437/out.txt)

Let me know if I can provide any more context @thvdveld !

thvdveld:
Could you maybe share the pcap file? Then I can open it with Wireshark as well.

benbrittain:
Github won't let me upload pcap files, so I've just renamed it as .txt and I hope they don't mangle it. 
[out.txt](https://github.com/smoltcp-rs/smoltcp/files/9122539/out.txt)

thvdveld:
Something is wrong with the calculation of the first fragment packet size and the n fragment packet size (I guess). I'll try to have a look at it this weekend, otherwise it will be in two weeks (holidays).

benbrittain:
Obviously no rush. I think you are right, I'm actually in the middle of printf debugging the `dispatch_ieee802154_out_packet` function and something definitely seems wonky with the fragn size calculation.

benbrittain:
poked at it a little this morning. Something about the fragn calculation seems off to me.

```
let frag_size = (*packet_len - *sent_bytes).min(*fragn_size);
```

this evaluates to 0, after the second packets (as expected!), but that means nothing gets copied into the buffer since we've you know, sent all the relevant data. We still get packets being generated as well.

thvdveld:
I think we can close this now, @benbrittain ? (fixed by #645)

PR #641, opened 2022-07-16
6LowPan fragmentation fix

benbrittain:
This is not yet ready for merging, there is also an issue with datagram offsets in the fragn packets.

Currently I have 
```
-                let datagram_offset = ((40 + *sent_bytes) / 8) as u8;
+                let datagram_offset = ((40 + *sent_bytes) / 8) as u8 - 4u8;
```

Which makes the datagram_offset line up enough that I can reassemble the packets on the other side, I'm fairly certain this is not the right way of doing it, that's an arbitrary offset derived from the difference of the request/response fragments, not from first principles.

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -3206,7 +3206,7 @@ impl<'a> InterfaceInner<'a> {
                 tx_buf = &mut tx_buf[ieee_len..];
 
                 // Add the next fragment header
-                let datagram_offset = ((40 + *sent_bytes) / 8) as u8;
+                let datagram_offset = ((8 + *sent_bytes) / 8) as u8;
                 fragn.set_offset(datagram_offset);
                 let mut frag_packet =
                     SixlowpanFragPacket::new_unchecked(&mut tx_buf[..fragn.buffer_len()]);
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -1188,7 +1188,7 @@ impl<'a> Interface<'a> {
             return Ok(false);
         }
 
-        if *packet_len >= *sent_bytes {
+        if *packet_len > *sent_bytes {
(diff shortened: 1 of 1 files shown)


benbrittain:
@thvdveld just a small upload for what I figured out with regards to #640

Dirbaio:
Marking as draft, please mark it back as ready to review when done.

thvdveld:
It seems that your changes make it work again. However, I don't have the alignment issues you have.

benbrittain:
I haven't had the time to check this since I fixed a different problem, the linux driver for my 802.15.4 card was dropping a few bytes from every near full size message on reception (where I was running wireshark)! There is a chance that this was the cause of the alignment error, so I'll check that now.

benbrittain:
It seems the alignment fix is still needed for me. 

Here are two pcap files, one without the alignment fix, one with. 

[pcap-trace-with-offset-fix.txt](https://github.com/smoltcp-rs/smoltcp/files/9234199/pcap-trace-with-offset-fix.txt)
[pcap-trace-with-no-offset-fix.txt](https://github.com/smoltcp-rs/smoltcp/files/9234198/pcap-trace-with-no-offset-fix.txt)

Linux will happily reassemble the packets of the one with the tweaked offset, otherwise we drop the packet

thvdveld:
Not sure if the offset calculation is correct now. It works in your case when you have one extended IEEE802.15.4 address and a short IEEE802.14.5 address, however, it does not work any more when we have two extended addresses.

benbrittain:
That makes sense and we clearly need to support both! I can't play around with this any more today, but I can take a look at it this week if you don't have time.

thvdveld:
I'm currently checking where it goes wrong. I'll let you know when I find something.

thvdveld:
I created a new PR, with, I think, the necessary fixes for fragmentation.

PR #642, opened 2022-07-25
fix return value of `nagle_enable`

stlankes:
The old implementation returns the wrong value (ACK delay duration) and doesn't describe if the Nagle algorithm is enabled or not.


Dirbaio:
Thank you!

bors r+

PR #645, opened 2022-08-02
Fixes 6LoWPAN fragmentation

thvdveld:
The calculation for the fragment sizes was wrong. This commit seems to
fix the calculation. This replaces #641.

It is a first step of improving the fragment size calculation (still really error-prone) and should be improved in future PRs.


thvdveld:
I think @benbrittain should test this before we should merge it.

benbrittain:
awesome! I can test it later this day. I'll report back soon

benbrittain:
Sorry about the delay, but I can confirm that this code now does work! Very exciting!

Dirbaio (approved):
Thanks guys! :) 

bors r+

PR #646, opened 2022-08-02
Fix `defmt` problems

datdenkikniet:
The missing derive broke a `net_debug!` print when compiled with `DEFMT_LOG=debug`

[This one](https://github.com/smoltcp-rs/smoltcp/blob/a6f6a6142f0040bbe2e3fbccbb869e5c675710ad/src/iface/fragmentation.rs#L181), to be specific, when compiling with (something similar to) `DEFMT_LOG=debug cargo c --features socket-raw,socket-udp,medium-ethernet,proto-ipv4,defmt --no-default-features`

Additionally, `--features defmt,alloc` didn't work. Enabling `defmt/alloc` when `alloc` is enabled fixes that.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.53.0
+          toolchain: 1.60.0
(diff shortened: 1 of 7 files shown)


Dirbaio (approved):
Thank you!

PR #648, opened 2022-08-04
Clarify how to back a PacketAssembler

benbrittain:


src/iface/interface.rs
```diff
@@ -1660,7 +1660,16 @@ impl<'a> InterfaceInner<'a> {
                     // This information is the total size of the packet when it is fully assmbled.
                     // We also pass the header size, since this is needed when other fragments
                     // (other than the first one) are added.
-                    check!(check!(fragments.reserve_with_key(&key)).start(
+                    let frag_slot = match fragments.reserve_with_key(&key) {
+                        Ok(frag) => frag,
+                        Err(Error::PacketAssemblerSetFull) => {
+                            net_debug!("No available packet assembler for fragmented packet");
```
crawford, line 1666: Should this be logged at a higher priority? (I’m not familiar with this functionality)

benbrittain: Debug seemed right to me because you shouldn't generally see this message and it shouldn't be considered fatal. Most smoltcp implementations will have their packet fragmenter backed by a dynamically growing data structure, this should only surface in a statically allocated situation. It also self-heals when the fragment gets completed or evicted after a timeout. The default error message through `check!` was somewhat misleading and made it seem like there was an issue with the underlying fragment, hence the debug message.

crawford: Makes sense, thanks.

Dirbaio (approved):
Thank you! 

bors r+

PR #649, opened 2022-08-04
Make the packet fragmentation cache timeout user configurable

benbrittain:
60 seconds is a really long time to hold onto these fragments. Since I'm working on a memory constrained microcontroller I can only have so fragmented packets in flight at a given point and it's not uncommon to miss a fragment and then be unable to receive any fragments until this evicts.

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -27,7 +27,8 @@ pub(crate) struct FragmentsBuffer<'a> {
     ipv4_fragments: PacketAssemblerSet<'a, Ipv4FragKey>,
     #[cfg(feature = "proto-sixlowpan-fragmentation")]
     sixlowpan_fragments: PacketAssemblerSet<'a, SixlowpanFragKey>,
-
+    #[cfg(feature = "proto-sixlowpan-fragmentation")]
+    sixlowpan_fragments_cache_timeout: Duration,
     #[cfg(not(any(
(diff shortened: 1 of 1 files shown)


thvdveld:
I just made it 60 seconds because that was the maximum timeout in the RFC. But 5 seconds is also fine for me!

datdenkikniet:
Would it perhaps make more sense to make this value configurable somehow?

benbrittain:
I agree with making it configurable. I still think 5 seconds is pretty long!

src/iface/interface.rs
```diff
@@ -393,6 +398,12 @@ let iface = builder.finalize(&mut device);
         self
     }
 
+    #[cfg(feature = "proto-sixlowpan-fragmentation")]
+    pub fn sixlowpan_fragments_cache_timeout(mut self, sec: u64) -> Self {
```
datdenkikniet: `sec: u64` -> `duration: Duration`?

benbrittain: Yeah, that is more ergonomic. 

@thvdveld do you think we should panic if the Duration is over 60s since that's in violation of the RFC?

thvdveld: Probably log it, but I don't think that we should panic.

Dirbaio (approved):
Thank you! 

bors r+

PR #652, opened 2022-08-05
Cleanup CI.

Dirbaio:
The GHA ubuntu images already have rustup, so actions-rs/toolchain is not required anymore.


crawford:
Does Ubuntu update the version of Rust/Clippy within a release number? i.e. Will we see the version of clippy change outside of an explicit change to our workflows? If so, I'm worried that this is going to be a continual surprise whenever the version silently changes.

PR #657, opened 2022-08-08
Pin Clippy Check

crawford:
This is a follow-up to https://github.com/smoltcp-rs/smoltcp/pull/652. I'm concerned that without explicit control over the clippy version, whenever there is a new stable release, work on this repository may be blocked until the latest clippy lints can be addressed. I've also documented the process of bumping the version for future reference.

There's a tradeoff to make here - this pull request reintroduces a slightly cumbersome process, but it does prevent surprise CI failures (notably, it removes the surprise, but technically CI is still broken in between the two pull requests - it just allows the maintainers control over when). Which direction would folks prefer?

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -1,3 +1,11 @@
+# When bumping the toolchain version, note that it must be done in two pull
+# requests - the first one being the version bump itself and the second one the
+# necessary code changes based on the updated clippy lints (if any). This is due
+# to the fact that actions-rs/clippy-check requires a GitHub API token with
+# write permission on the "check" scope. A token of this type is only available
+# when triggering on `pull_request_target`, but this runs the action using the
+# configuration from the base of the pull request rather than the merge commit.
+
 on:
   push:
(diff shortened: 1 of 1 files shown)


Dirbaio:
`actions-rs/toolchain` is unneeded, the GHA images already have rustup. The installed rust can be controlled with the standard rustup ways (RUSTUP_TOOLCHAIN envvar, rust-toolchain.toml file).

To avoid the "two phase" clippy bump, maybe the action can read which version to use from the repo itself? with something like `RUSTUP_TOOLCHAIN=$(cat .github/workflows/clippy-version.txt)`?

crawford:
If you set `RUSTUP_TOOLCHAIN` to a specific version, the job will fail with the following:

    error: toolchain '1.60.0-x86_64-unknown-linux-gnu' is not installed

Same is true if you use `toolchain` on the clippy action directly. The specific toolchain has to be installed before it can be used.

crawford:
Closing in favor of #664.

PR #658, opened 2022-08-08
Limit the permissions of the clippy action

crawford:
actions-rs/clippy-check only requires read/write access to the
"checks" scope. When the "permissions" object is present, all of the
scopes default to "none".


Dirbaio (approved):
Makes a lot of sense, thank you! :)

PR #661, opened 2022-08-10
Asserts packets are sent in TCP tests

### LINKED ISSUE #659: TCP test suites no longer assert when a packet is not received
As of https://github.com/smoltcp-rs/smoltcp/commit/629f0bce7991bfb8e73aeeb3b39982338d0eb97b, it's possible for a socket's dispatch method to return `Ok(())` without ever emitting a packet. However, as the TCP test suite runs its assertions within the emit method, tests silently pass when they should not:

https://github.com/smoltcp-rs/smoltcp/blob/7d224457d2c8f0a42da0c291a74e85705fc439af/src/socket/tcp.rs#L2428-L2442

There's a couple of possible fixes I can see to this - happy to put a PR together if that's useful?
 - Synthesise an `Error::Exhausted` when the callback is not invoked.
 - Assert that the callback is invoked exactly once, and define a separate `recv_nothing` function to use when nothing should have been sent.

SquidDev:
- Assert the emit callback is invoked exactly once.
 - Add a separate macro for checking that no packets were sent.

Fixes #659.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -2425,6 +2425,7 @@ mod test {
     {
         socket.cx.set_now(timestamp);
 
+        let mut sent = 0;
         let result = socket
             .socket
             .dispatch(&mut socket.cx, |_, (ip_repr, tcp_repr)| {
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Thank you!

bors r+

SquidDev:
Thanks!

PR #662, opened 2022-08-11
Only clear retransmit timer when all packets are acked

### LINKED ISSUE #660: TCP packets are not transmitted when bursts are partially acknowleged
I've recently been trying to debug a problem where smoltcp gets stuck sending packets, and will refuse to make progress. My understanding is that when when a burst of packets are sent, but only a subset of those are acknowledged, the remaining packets are never resent.

Here is a minimal test case that I would expect to pass, but fails on 0.8.1

```rust
fn test_data_retransmit_bursts_half_ack() {
    let mut s = socket_established();
    s.remote_mss = 6;
    s.send_slice(b"abcdef012345").unwrap();

    recv!(s, time 0, Ok(TcpRepr {
        control:    TcpControl::None,
        seq_number: LOCAL_SEQ + 1,
        ack_number: Some(REMOTE_SEQ + 1),
        payload:    &b"abcdef"[..],
        ..RECV_TEMPL
    }), exact);
    recv!(s, time 0, Ok(TcpRepr {
        control:    TcpControl::Psh,
        seq_number: LOCAL_SEQ + 1 + 6,
        ack_number: Some(REMOTE_SEQ + 1),
        payload:    &b"012345"[..],
        ..RECV_TEMPL
    }), exact);
    // Acknowledge the first packet
    send!(s, time 5, TcpRepr {
        seq_number: REMOTE_SEQ + 1,
        ack_number: Some(LOCAL_SEQ + 1 + 6),
        window_len: 6,
        ..SEND_TEMPL
    });
    // The second packet should be re-sent.
    recv!(s, time 1500, Ok(TcpRepr { // Fails! Instead we receive Error::Exhausted
        control:    TcpControl::Psh,
        seq_number: LOCAL_SEQ + 1 + 6,
        ack_number: Some(REMOTE_SEQ + 1),
        payload:    &b"012345"[..],
        ..RECV_TEMPL
    }), exact);

    recv!(s, time 1550, Err(Error::Exhausted));
}
```

From what I can tell, when the socket is in the Established state and receives an ack, it _always_ resets the retransmit timer, even if some data is unacknowledged. 

https://github.com/smoltcp-rs/smoltcp/blob/7d224457d2c8f0a42da0c291a74e85705fc439af/src/socket/tcp.rs#L1587-L1591

Patching out the timer reset causes the test to pass, though obviously is not the correct fix here! I assume this check would need to move further down to where `local_seq_no` and `remote_last_seq` are updated?

Apologies if I've misunderstood the expected behaviour or got anything also wrong here! Definitely not an expert on networking.

SquidDev:
Fixes #660.

While I've tested this on several physical devices, and can confirm they behave as expected, I'm definitely not familiar with the intricacies of smoltcp (or TCP in general!), so not sure what potential regressions there might be here. Happy to make any and all changes needed!

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1460,6 +1460,7 @@ impl<'a> Socket<'a> {
         // from the sequence space.
         let mut ack_len = 0;
         let mut ack_of_fin = false;
+        let mut ack_all = false;
         if repr.control != TcpControl::Rst {
             if let Some(ack_number) = repr.ack_number {
                 // Sequence number corresponding to the first byte in `tx_buffer`.
@@ -1477,6 +1478,8 @@ impl<'a> Socket<'a> {
                         tcp_trace!("received ACK of FIN");
                         ack_of_fin = true;
                     }
+
+                    ack_all = self.remote_last_seq == ack_number
                 }
 
                 self.rtte.on_ack(cx.now(), ack_number);
@@ -1585,7 +1588,7 @@ impl<'a> Socket<'a> {
             // ACK packets in ESTABLISHED state reset the retransmit timer,
             // except for duplicate ACK packets which preserve it.
             (State::Established, TcpControl::None) => {
-                if !self.timer.is_retransmit() || ack_len != 0 {
+                if !self.timer.is_retransmit() || ack_all {
                     self.timer.set_for_idle(cx.now(), self.keep_alive);
                 }
             }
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Looks good to me. Nice catch, and thanks for the fix and tests!

bors r+

SquidDev:
Sorry to bump an old issue, was wondering if it would be possible to backport this to the 0.8 branch and cut a new release? We're currently maintaining a local fork of [Artiq](https://github.com/m-labs/artiq/) with this smoltcp patch applied, but we'd quite like to move upstream to use the fixed version of smoltcp too.

Dirbaio:
Sure, we can do that! Can you send a PR against the `v0.8.x` branch with the backport (along with other stuff you want backported if any)?

SquidDev:
Amazing, thank you so much!

PR #664, opened 2022-08-11
Pin clippy to minimum supported Rust version 

crawford:
Alternate to #657. Builds on #663.

The rationale is the same as discussed in 0b0f96e:

> It can be rather surprising when new lints pop up when a new stable
> toolchain is released. Let's pin this check to a specific version to
> avoid those surprises.

In deciding which version of clippy to use, I went with the MSRV since
that's what's been done historically. The other option was to read
from the repo a version number specifically for clippy, but I was
afraid that adding one more version number to juggle would increase
the odds that it would be forgotten and fall out of sync.

Note that this approach uses rustup to install the toolchain
dynamically rather than making use of an action. The advantage of this
method is that it allows a single pull request to contain the version
bump and suggested code changes (this is due to the fact that
actions-rs/clippy-check requires a GitHub API token with write
permission, but a token of this type is only available when triggering
on `pull_request_target` which runs the action using the configuration
from the base of the pull request rather than the merge commit). The
disadvantage of this approach is that the toolchain setup can no
longer be cached by the underlying layering mechanism used by GitHub
actions (unlikely to significantly affect this project).

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -7,8 +7,6 @@ name: Clippy check
 jobs:
   clippy:
     runs-on: ubuntu-latest
-    env:
-      RUSTUP_TOOLCHAIN: stable
     permissions:
       checks: write
     steps:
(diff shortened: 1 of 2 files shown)


crawford:
Also note that CI isn't actually testing these changes. I've done my best to test them in a dummy repository, so I don't _think_ this will break once it's merged. Hopefully this is the last time we'll have to deal with this.

Dirbaio (approved):
This is really nice! I like this much more than the previous approach, this won't require the "two-phase PRs" to bump the MSRV anymore! 

Thank you! :)

bors r+

PR #665, opened 2022-08-16
Send incomplete fin packets even if nagle enabled

mbirtwell:
If we have an incomplete packet to send and the socket has been closed
then send it. Without this change (with nagle enabled) we wait for the
penultimate packet in the stream to be acked before sending the final
packet even if there's plenty of space in the window.

I think this is strictly only an optimisation, but when I wrote the code I thought it was necessary.

If there is no packet loss this change saves you a RTT before sending the final packet in the stream (plus any ack delay). If there is incoming packet loss so that you've missed an ACK, this change might additionally save you from having to do some retransmission. Once the far end has received the FIN/ACK it can drive retransmission if that gets dropped.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1898,7 +1898,12 @@ impl<'a> Socket<'a> {
             _ => false,
         };
 
-        if self.nagle && data_in_flight && !can_send_full {
+        // If we're applying the Nagle algorithm we don't want to send more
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Thank you! Makes a lot of sense, there's no point in delaying the last packet indeed. Nice find!

bors r+

PR #666, opened 2022-08-17
change `respond` closure in `Interface::socket_egress` to propogate errors up

ritikmishra:
On a hardware device, I was running into an issue where the device transmit buffer was filling up, so it was no longer issuing `TxToken`s. However, `smoltcp` was ignoring that error, so we were observing that our packets were not being sent.

Propagating the error up fixed this problem


Dirbaio (approved):
Thank you!

bors r+

PR #667, opened 2022-08-18
Change egress error handling

thvdveld:
Previously, error handling was performed in the closure and after the
closure as well. Now, error handling is performed in one place.

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -1194,32 +1194,24 @@ impl<'a> Interface<'a> {
             let mut neighbor_addr = None;
             let mut respond = |inner: &mut InterfaceInner, response: IpPacket| {
                 neighbor_addr = Some(response.ip_repr().dst_addr());
-                match device.transmit().ok_or(Error::Exhausted) {
-                    Ok(_t) => {
-                        #[cfg(any(
-                            feature = "proto-ipv4-fragmentation",
-                            feature = "proto-sixlowpan-fragmentation"
-                        ))]
-                        if let Err(e) = inner.dispatch_ip(_t, response, Some(_out_packets)) {
-                            net_debug!("failed to dispatch IP: {}", e);
-                            return Err(e);
-                        }
+                let t = device.transmit().ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


src/iface/interface.rs
```diff
@@ -1082,24 +1082,16 @@ impl<'a> Interface<'a> {
...
-                        }
+                        inner.dispatch_ip(t, response, None)?;
+
                         emitted_any = true;
                     }
-                    Err(e) => {
-                        net_debug!("failed to transmit IP: {}", e);
-                        return Err(e);
-                    }
+                    Err(e) => return Err(e),
```
crawford: The entire match expression can be dropped if the error is just being passed through.

src/iface/interface.rs
```diff
@@ -1082,24 +1082,16 @@ impl<'a> Interface<'a> {
...
-                        if let Err(e) = inner.dispatch_ip(_t, response, None) {
-                            net_debug!("failed to dispatch IP: {}", e);
-                            return Err(e);
-                        }
+                        inner.dispatch_ip(t, response, None)?;
+
                         emitted_any = true;
                     }
-                    Err(e) => {
-                        net_debug!("failed to transmit IP: {}", e);
```
crawford, line 1219: On first inspection, it looks like this transmit-specific debug message is being lost. If any of the other code paths in this closure return `Unaddressable`, it won’t be possible to distinguish between them.

src/iface/interface.rs
thvdveld: I removed the match expression now.

src/iface/interface.rs
thvdveld: Totally right! I replaced the `unwrap_or` with an `unwrap_or_else` and log it there that there was a device exhausted error.

Dirbaio (approved):
Thanks! 

One small note: I've reverted the `match result` -> `if let Err(err) = result { match err {` because I think the 1st is more "Rusty" idiomatic, leaving the changes in the `respond` closure only.

bors r+

PR #669, opened 2022-08-22
Adds one-shot mDNS resolution

benbrittain:
RFC 6762 Section 5.1 specifies a one-shot multicast DNS query. This
query has minimal differences from a standard DNS query, mostly just
using a multicast address and a different port (5353 vs 53).

A fully standards compliant mDNS implementation would use UDP source
port 5353 as well to issue queries, however we MUST NOT use that port
and continue using an ephemeral port until features such as service
discovery are implemented.

This change also allows specifying what kind of DNS query we wish to
perform.

https://www.rfc-editor.org/rfc/rfc6762#section-5.1

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -3331,7 +3331,7 @@ impl<'a> InterfaceInner<'a> {
         let mut total_size = 0;
         total_size += iphc_repr.buffer_len();
         let mut _compressed_headers_len = iphc_repr.buffer_len();
-        let mut _uncompressed_headers_len = ip_repr.buffer_len();
+        let mut _uncompressed_headers_len = ip_repr.header_len();
 
         #[allow(unreachable_patterns)]
         match packet {
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -1015,22 +1015,14 @@ impl<'a> Interface<'a> {
         self.inner.now = timestamp;
 
         #[cfg(feature = "proto-ipv4-fragmentation")]
-        if let Err(e) = self
-            .fragments
+        self.fragments
             .ipv4_fragments
-            .remove_when(|frag| Ok(timestamp >= frag.expires_at()?))
-        {
-            return Err(e);
-        }
+            .remove_when(|frag| Ok(timestamp >= frag.expires_at()?))?;
 
         #[cfg(feature = "proto-sixlowpan-fragmentation")]
-        if let Err(e) = self
-            .fragments
+        self.fragments
             .sixlowpan_fragments
-            .remove_when(|frag| Ok(timestamp >= frag.expires_at()?))
-        {
-            return Err(e);
-        }
+            .remove_when(|frag| Ok(timestamp >= frag.expires_at()?))?;
 
         #[cfg(feature = "proto-ipv4-fragmentation")]
         match self.ipv4_egress(device) {
--- a/src/socket/dns.rs
+++ b/src/socket/dns.rs
@@ -90,7 +90,14 @@ struct PendingQuery {
     delay: Duration,
 
     server_idx: usize,
-    mdns: bool,
+    mdns: MulticastDns,
+}
+
+#[derive(Debug)]
+pub enum MulticastDns {
+    Disabled,
+    #[cfg(feature = "socket-mdns")]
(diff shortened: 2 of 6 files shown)


src/socket/dns.rs
```diff
@@ -234,19 +251,33 @@ impl<'a> Socket<'a> {
...
         raw_name: &[u8],
+        query_type: Type,
+        mdns: bool,
     ) -> Result<QueryHandle, StartQueryError> {
         let handle = self.find_free_query().ok_or(StartQueryError::NoFreeSlot)?;
 
+        if mdns {
+            // as per RFC 6762 any DNS query ending in .local. MUST be sent as mdns
+            // so we internally overwrite the servers for any of those queries
+            self.update_servers(&[
```
timokroeger: It looks like this line overrides the dns servers permanently. Which would mean that all following regular DNS queries will fail.

benbrittain: Totally does, fixed.

benbrittain:
@thvdveld Is this the sort of change you review as well? It's relevant in a 6lowpan context.

thvdveld:
I'm not very familiar with DNS, so I don't feel comfortable reviewing this. But I'll take a look at it tomorrow!

Also, a change slipped into master that messed with 6LoWPAN fragmentation, the fix is in #681.

benbrittain:
oh good! I'm glad you told me that because I literally just started a git bisect trying to figure out what broke :)

src/socket/dns.rs
```diff
@@ -200,6 +211,12 @@ impl<'a> Socket<'a> {
 
         let mut raw_name: Vec<u8, MAX_NAME_LEN> = Vec::new();
 
+        let mut mdns = false;
+        if name.split(|&c| c == b'.').last().unwrap() == b"local" {
```
thvdveld, line 223: What would happen if the name is just "local", without a dot. Then `.last()` will return "local". Is this correct behaviour? Like I said, I don't know anything about mDNS, so it could be that this is valid.

src/socket/dns.rs
```diff
@@ -81,6 +90,7 @@ struct PendingQuery {
     delay: Duration,
 
     server_idx: usize,
+    mdns: bool,
```
Dirbaio: This could be a "Protocol" enum, with valus "Dns" and "Mdns".

Dirbaio:
Looks good to me overall. Could you add a Cargo feature for MDNS, so that binaries not using it don't get increased code size?

src/socket/dns.rs
benbrittain: My understanding is that "local" without anything else refers to the top-level domain, in which case the RFC does say under "3.0 Multicast DNS Names" that this has the special semantics.

src/socket/dns.rs
benbrittain: I don't think the API should encourage thinking about mDNS and DNS queries as fundamentally different protocols

benbrittain:
I don't think this will have much meaningful code size, but I created a feature. I think future mDNS service discovery functionality should also go under that feature set.

I left the mdns enum inside the `PendingQuery` struct intentionally. with "socket-mdns" disabled, it should be trivially optimized out and it keeps the cfg flow less confusing below. Happy to make the removal from the PendingQuery struct explicit if you'd prefer though.

src/socket/dns.rs
benbrittain: I did something in this vein when I put this behind a feature flag

src/iface/interface.rs
```diff
@@ -1015,22 +1015,14 @@ impl<'a> Interface<'a> {
         self.inner.now = timestamp;
 
         #[cfg(feature = "proto-ipv4-fragmentation")]
-        if let Err(e) = self
-            .fragments
+        self.fragments
```
Dirbaio, line 1018: seems there are some unrelated changes

Dirbaio: ah it's clippy lints, I see

Dirbaio (approved):
Thank you!

bors r+

(Yes, the enum becoming single-case is fine, it's the clearest way to do this. We do this in other places already, for example for ipv4/v6)

PR #671, opened 2022-08-28
fix socket feature check

M1cha:
it requires you to have at least one socket type enabled but the feature
`socket-dhcp` does not exist because it's name is `socket-dhcpv4`. So if
that's only socket type you want enabled you weren't able to do that due
to this broken check.


Dirbaio (approved):
Thank you! Nice catch.

bors r+

PR #673, opened 2022-08-31
Specified the representation of error codes to make easier passing them through FFI interfaces.

KOLANICH:
This PR is intended to eliminate the code like 

```rust
fn tcpconnect_error_from_connect_error(
	res: core::result::Result<(), tcp::ConnectError>,
) -> TCPConnectError {
	match res {
		Ok(_) => TCPConnectError::OK,
		Err(err) => match err {
			tcp::ConnectError::InvalidState => TCPConnectError::InvalidState,
			tcp::ConnectError::Unaddressable => TCPConnectError::Unaddressable,
		},
	}
}
```

when passing the error enums through FFI by specifying numeric values of them within the main lib.


Dirbaio:
Thanks for the pull request! Unfortunately, I don't think this is something we can merge.

`smoltcp` is a Rust library intended to be used in Rust programs, therefore its public API aims to be as idiomatic Rust as possible. It is not idiomatic to have "Ok" variants in error enums, because it causes duplicate ways of returning "success" from functions returning `Result<(), ResultCode>`: either `Ok(())` or `Err(ResultCode::Ok)`. 

Converting from "idiomatic Rust API" to "idiomatic C API" should be a responsibility of the FFI wrapper crate, not smoltcp itself.

KOLANICH:
>Converting from "idiomatic Rust API" to "idiomatic C API" should be a responsibility of the FFI wrapper crate, not smoltcp itself.

I think it is the most efficient to implement it here. I keep the API idiomatic, but added

1. a grand global enum for enumeratig error codes. All the specific codes are still one byte, and the most of space within 1 byte is unoccupied, so I guess we can affford to assign an unique number for each code, so it should be zero overhead.
2. into this enum I have added 2 additional codes. 1 for success and 1 for providing a buffer that is too small.
3. I have tried to add the `From<Result<...>>` traits within my crate, the compiler didn't allow me to do that.
4. It is a burden to generate these conversion traits separately from the enums. So I have done a different thing:
* the specific enums get the `#[repr(u8)]`, so they are mapped to `uint8_t` when passed through FFI. Otherwise I'd have to just copy these enums definitions into my FFI crate and write nasty boilerplate code as the one I have mentioned, and I'm completely not sure it can be generated automatically without doing additions into the `smoltcp` crate. Doing this way is inefficient, unproductive and is additional burden.
* the specific enums just get the values from the grand enum;
* the names and values they get are the same names and numeric values as in the grad enum, tue assignment is done by the macro, so the code looks idiomatic, it looks even better tuan without the macro;
* for an each specific enum a `From<Result<>>` trait is generated, which can be used to convert tue idiomatic Rust result into the integer code that can be passed through ffi; the integer error code corresponds to the codes in the grand enum, but the typed enums are used in fact, not the grand enum, so accidentially passing the error code from another enum will be statically detected.

KOLANICH:
Also I have done a bit of renamings to reflect the fact that the code may contain not only errors, but also a success.

Dirbaio:
As I said, having an "Ok" variant in Error enums is *not* idiomatic.

Also, having one big Error enum with all errors is not idiomatic either. I've been actively refactoring the code to replace it with function-specific enums, see #617 . The end goal is to remove `Error` completely. This change is a step back w.r.t that goal.

KOLANICH:
>I've been actively refactoring the code to replace it with function-specific enums, see #617 .

Yes, it is good and I have kept them. But to make FFI interface simpler we need a single bijection `number <-> error semantics`. It would allow us to have a single enum on other side of FFI.

If we don't have a single bijection, we would have to define a manual one and maintain it.

>Also, having one big Error enum with all errors is not idiomatic either.  The end goal is to remove Error completely. This change is a step back w.r.t that goal.

It may be idiomatic and pure to liquidate that index, but it will make significantly harder to create and maintain FFI bindings to this lib. If we liquidate it here, then I'll have to hand-assemble it in my lib manually instead of relying on compiler machinery in order to do it.

Dirbaio:
Yes, I agree these things are good to have *for C FFI bindings*. However these things should be done in the *FFI binding crate*, not in smoltcp itself. Smoltcp's goal is to provide a network stack with a Rust API, that's all.

KOLANICH:
But it cannot be automated enough when done in a separate crate, can it?

Dirbaio:
You'll have to write some boilerplate, yes.

PR #674, opened 2022-08-31
Make constants exposed in DNS socket API public

ebarnard:
`MAX_ADDRESS_COUNT` is useful when naming the output type of `dns::Socket::get_query_result`.

`MAX_SERVER_COUNT` is necessary to ensure `dns::Socket::new` and `dns::Socket::update_servers` do not panic if passing a user-provided list of DNS servers.


Dirbaio (approved):
Sounds reasonable, thanks for the PR! :)

PR #675, opened 2022-09-01
Made IPv4 and IPv6 modules a bit DRYer by moving sizes into constants and exposed IPv4-mapping prefix.

KOLANICH:


Dirbaio (approved):
Nice improvements, thank you! :) 

bors r+

PR #676, opened 2022-09-01
Added a sans-io style interface.

KOLANICH:


Dirbaio:
smoltcp can already be used sans-io, by implementing the `Device` trait for your own struct in your own crate That's intended to be the main API for users using smoltcp on top of custom interfaces. 

Unfortunately, I don't think the smoltcp crate should have a "sans IO phy" like this, because in the end it fulfills the same use case as the user implementing Device directly, but with some extra downsides (unbounded memory usage, makes allocations for the `Vec<u8>`s)

PR #677, opened 2022-09-05
Remove cfg-if dep

thvdveld:
We don't actually need the `cfg-if` dependency. It's also annoying to format inside the `cfg_if!` macro. This PR removes it.


thvdveld:
Never mind... (I hate all those features)

PR #678, opened 2022-09-06
DHCPv4: use a Vec to store DNS server addresses instead of an array o…

datdenkikniet:
…f options

This simplifies usage of this field, but does remove the `Copy` derive from that struct.

If the `Copy` derive on `Dhcpv4Repr` is not a requirement, I would like to replace it's internal `Option<[Option<Ipv4Address>; 3]>` with an `Option<Vec<Ipv4Address; 3>>` too.

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -424,7 +424,6 @@ impl<'a> Socket<'a> {
             .dns_servers
             .iter()
             .flatten()
-            .flatten()
             .filter(|s| s.is_unicast())
             .for_each(|a| {
                 // This will never produce an error, as both the arrays and `dns_servers`
@@ -779,8 +778,6 @@ mod test {
     const DNS_IP_1: Ipv4Address = Ipv4Address([1, 1, 1, 1]);
     const DNS_IP_2: Ipv4Address = Ipv4Address([1, 1, 1, 2]);
     const DNS_IP_3: Ipv4Address = Ipv4Address([1, 1, 1, 3]);
-    const DNS_IPS_ARR: [Option<Ipv4Address>; DHCP_MAX_DNS_SERVER_COUNT] =
-        [Some(DNS_IP_1), Some(DNS_IP_2), Some(DNS_IP_3)];
     const DNS_IPS: &[Ipv4Address] = &[DNS_IP_1, DNS_IP_2, DNS_IP_3];
 
     const MASK_24: Ipv4Address = Ipv4Address([255, 255, 255, 0]);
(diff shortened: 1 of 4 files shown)


src/socket/dhcpv4.rs
```diff
@@ -417,17 +418,20 @@ impl<'a> Socket<'a> {
...
-                }
-            }
-        }
+        let mut dns_servers = Vec::new();
+
+        dhcp_repr
+            .dns_servers
+            .iter()
+            .flatten()
+            .flatten()
```
crawford: Are either of the calls to flatten() necessary?

datdenkikniet: The first flattens it from an `Option<[Option<Address>; _]>` to an `Iter<Item = [Option<Address; _]>` and the second to an `Iter<Item = Address>`. The would be to do it with if/let.

I personally prefer it this way, but can change it if that is better.

crawford: Yep, brain fart on my part. I was thinking that it was the heapless::Vec that was being iterated.

Dirbaio: perhaps it's worth changing it to a Vec in DhcpRepr too?

datdenkikniet:
I've swapped the array in `DhcpRepr` too, now. It makes doing const stuff a bit more difficult, but should work fine.

src/wire/dhcpv4.rs
```diff
@@ -773,13 +774,11 @@ impl<'a> Repr<'a> {
...
-                    let chunk_size = 4;
-                    for (server, chunk) in servers.iter_mut().zip(data.chunks(chunk_size)) {
-                        if chunk.len() != chunk_size {
-                            return Err(Error);
-                        }
-                        *server = Some(Ipv4Address::from_bytes(chunk));
+                    let mut servers = Vec::new();
+                    const IP_ADDR_BYTE_LEN: usize = 4;
+                    for chunk in data.chunks(IP_ADDR_BYTE_LEN) {
+                        // If the DCHP offer is standard compliant, this should never fail
```
Dirbaio: the RFCs don't set a max DNS server count, `MAX_DNS_SERVER_COUNT` is a smoltcp restriction only.

datdenkikniet: Oops, bad assumption on my part. Fixed

crawford:
How does this affect the size of the resulting binary? Do the clone()s get optimized away?

datdenkikniet:
Given that the underlying types are `Copy`, it shouldn't make a big difference. I haven't looked at it in detail though, I will try to do so on Thursday.

datdenkikniet:
In my (quite specific) test, it seems like the size is somewhat decreased:

| Branch     | Size  |
|------------|-------|
| `master`   | 29068 |
| `dhcp_vec` | 28736 |

It's a basic project with a smoltcp `Interface`, one TCP socket and one DHCP socket.
This is with `opt-level = "s"`, `lto = true`, (find it [here](https://github.com/datdenkikniet/stm32-eth/tree/smoltcp-size-check), the `rtic-echo` example with `stm32f429`). 

I also included bloat files that seem to indicate that the `buffer_len` calculation has decreased in size a bit, as well as `parse_ack`. I used `DEFMT_LOG=off cargo bloat -n 500 --release --example rtic-echo --features rtic-echo-example,stm32f429 --target thumbv7em-none-eabihf` to generate those.

crawford:
@datdenkikniet, very cool. Thank you for checking.

Dirbaio (approved):
Thank you for checking. This is really cool! Very interesting that size goes *down*, I wonder if we should use `heapless::Vec` more widely then, there's quite a few places around the codebase with `[Option<T>; N]`'s that could ve `Vec`s.

Bors r+

PR #679, opened 2022-09-21
Remove implicit sized bound on device generics

tsoutsman:
The generic is only ever accessed through a reference and so it doesn't have to be sized.

Signed-off-by: Klim Tsoutsman <klim@tsoutsman.com>


Dirbaio (approved):
Thank you! It makes sense, can't see a reason to not allow this. 

What're you doing that is making your Device unsized though? Just curious, It seems odd!

bors r+

tsoutsman:
I want to do something like so:
```rust
impl smoltcp::phy::Device for dyn Nic {}
```
where `Nic` is my own trait.

PR #681, opened 2022-09-22
Refactor 6LoWPAN function

thvdveld:
Now, handling UDP packets is done using `process_udp`. Also removed `cfg_if!` in `interface` since that messed with formatting of the code.

It also contains a fix for the fragmentation.

### DIFF
--- a/src/iface/interface.rs
+++ b/src/iface/interface.rs
@@ -1950,59 +1950,17 @@ impl<'a> InterfaceInner<'a> {
                         let udp_repr = check!(SixlowpanUdpNhcRepr::parse(
                             &udp_packet,
                             &iphc_repr.src_addr,
-                            &iphc_repr.dst_addr,
+                            &iphc_repr.dst_addr
                         ));
 
-                        // Look for UDP sockets that will accept the UDP packet.
-                        // If it does not accept the packet, then send an ICMP message.
-                        //
-                        // NOTE(thvdveld): this is currently the same code as in self.process_udp.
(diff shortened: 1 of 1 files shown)


benbrittain:
Can confirm this fixes the 6LoWPAN fragmentation errors.

src/iface/interface.rs
```diff
@@ -778,6 +778,33 @@ impl<'a> IpPacket<'a> {
     }
 }
 
+/// Struct only used internally. We want to be able to use `process_udp` for every medium we
```
Dirbaio: This sounds very very complicated. Why not pass the parsed UdpRepr (src+dst ports) and the UDP payload to `process_udp` instead? This way each caller can do the parsing accordingly. (if there's multiple callers to `process_udp` uncompressed, perhaps having a `process_ip_udp` fn could help avoid duplicating parsing code?).

thvdveld: I changed it. But I'm still not happy about it. The ICMP reply for when the destination is unreachable makes everything more difficult.

thvdveld: I guess it always has been incorrect for 6LoWPAN.

thvdveld:
Going to merge this, but making an issue for the ICMP unreachable payload that is incorrect for 6LoWPAN (which has always been the case since for now).

bors r+

thvdveld:
Apparently ICMPv6 unreachable error messages are created correctly, and I just thought they were wrong.

PR #683, opened 2022-09-23
Use renewal time from DHCP server ACK, if given

JarredAllen:
# Description

Per RFC 2132 section 9.11 the server can manually specify a renewal (T1) time different from the default value (half the lease time) through option code 58. This PR updates the behavior of the dhcp client to use that value, if provided, and only if not provided does it default to half of the lease duration.

Since smoltcp seems to ignore the REBINDING state, I also made it look for a provided rebinding (T2) time provided by the server (dhcp option 59) and made it use that value as the renewal time if no renewal time was provided and the rebinding time is less than the default. This behavior seems sensible to me, given that we're not following the REBINDING part of the spec, but I can change it to ignore option code 59, or any other handling, if that is preferred.

# Verification

I realized that this functionality was missing when I changed my configuration to set a 10 second renew time on a lease which lasts for a very long time, and observed that my devices (which use this library) weren't attempting to renew. To verify that this PR works, I ran it in my existing setup and confirmed that my devices now renew their leases at approximately 10 second intervals. I think this, alongside the tests in CI, should be good enough.

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -439,8 +439,23 @@ impl<'a> Socket<'a> {
             packet: None,
         };
 
-        // RFC 2131 indicates clients should renew a lease halfway through its expiration.
-        let renew_at = now + lease_duration / 2;
+        // Set renew time as per RFC 2131:
+        // The renew time (T1) can be specified by the server using option 58:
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
Looks great. Thank you!

bors r+

PR #689, opened 2022-10-08
fix boundary case of assembler::remove_contig_at

hikaricai:
Found a bug when learning source code of smoltcp.


Dirbaio (approved):
Very nice catch, thank you! Tricky bug :) 

bors r+

PR #693, opened 2022-10-12
Constification of the wire module

ngc0202:
A huge portion of the `wire` module is constifyable, as shown in this commit. Making these `const fn`s, especially the functions set aside in the `field` modules of wire, would be greatly beneficial to me as a consumer of this crate, and I imagine many others. Many of the structs in my own crate's "wire" module for a custom protocol are more constrained versions of the packets and reprs provided by smoltcp. Due to being more constrained, many fields which are variable in the generic reprs are now constant in mine, which then lend themselves to being constants in my crate, however these functions and methods not being marked `const` prevents me from using smoltcp's tools to calculate these values. **Most** of the functions which I marked `const` in this commit are by their very nature/semantics guaranteed to be constifiable anyway, however a handful aren't necessarily, therefore:

**Important note:** While the majority of the functions are marked `const` are very safe to do so, not all of them are guaranteed to be so, and the reviewer of this PR should take caution and feel free to unmark something as const if they're not comfortable making that guarantee in the public API.

Another note: There are various other functions which can't be made `const fn` as written, but could be with minor tweaks. I did not include these tweaks but I think it should be considered as further work for this PR or a potential follow-up PR. Furthermore, plenty outside of `wire` is constifyable as well but I did not include those in the scope of this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
         # Failure is permitted on nightly.
         rust:
           - stable
-          - 1.60.0
+          - 1.61.0
           - nightly
 
         features:
@@ -65,7 +65,7 @@ jobs:
         # Failure is permitted on nightly.
         rust:
           - stable
-          - 1.60.0
+          - 1.61.0
           - nightly
 
         features:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "smoltcp"
 version = "0.8.1"
 edition = "2018"
-rust-version = "1.60"
+rust-version = "1.61"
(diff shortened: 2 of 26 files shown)


ngc0202:
In response to the test failure: The issue is the `new_unchecked` functions which I marked `const fn`. This was stabilized in Rust 1.61, one version after the MSRV. Should I remove those for now or is bumping up the MSRV by one viable for this?

Dirbaio (approved):
Thanks! Constifying makes sense, there's no downside really :) 

LGTM to bumping the MSRV, we're going to bump it to 1.65 in #572  anyway :+1: 

bors r+

PR #695, opened 2022-10-15
SocketSet: managed slice and socket storage needs seperate lifetimes

gopakumarce:
Consider a code snippet like below, its compiled with default-features = false and #![no_std]

```

fn foo<'a>(some_socket: tcp::Socket<'a>) -> Socket<'a> {
    let mut sock_arr: [SocketStorage; 1] = [SocketStorage::EMPTY; 1];
    let onesock: &mut [SocketStorage] = &mut sock_arr;
    let mut sockets = SocketSet::new(onesock);
    let handle = sockets.add(some_socket);
    return sockets.remove(handle);
}
```

The above will fail complaining that &mut sock_arr does not live long enough. This is because the lifetime of the socket and the slice where the socket is stored are both the same right now and hence the above cant be done.

```
   |
19 |     let onesock: &mut [SocketStorage] = &mut sock_arr;
   |                                         ------------- `sock_arr` is borrowed here
...
22 |     return sockets.remove(handle);
   |            ^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

Adding an additional lifetime to make sure that this works fine. And obviously doing this meant touching all the APIs where SocketSet is used - most of the APIs in interface.rs. Note that in APIs like poll(), the lifetime 'data is different from lifetime 'a of the Interface itself (obviously?) because both represent differententities and we dont want callers of iface.poll(...,socket_set) to have the complaint that socket_set should live as long as iface

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -16,7 +16,9 @@ jobs:
           ref: refs/pull/${{ github.event.number }}/head
       - uses: actions/checkout@v2
         if: github.event_name != 'pull_request_target'
-      - run: sed -n 's,^rust-version = "\(.*\)"$,RUSTUP_TOOLCHAIN=\1,p' Cargo.toml >> $GITHUB_ENV
+      # TODO update when stable Rust 1.65 is out.
+      #- run: sed -n 's,^rust-version = "\(.*\)"$,RUSTUP_TOOLCHAIN=\1,p' Cargo.toml >> $GITHUB_ENV
+      - run: echo RUSTUP_TOOLCHAIN=beta >> $GITHUB_ENV
       - run: rustup toolchain install $RUSTUP_TOOLCHAIN
       - run: rustup component add clippy
       - uses: actions-rs/clippy-check@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,8 +20,9 @@ jobs:
         # Test on stable, MSRV, and nightly.
         # Failure is permitted on nightly.
         rust:
-          - stable
-          - 1.60.0
+          # TODO update when stable Rust 1.65 is out.
+          #- stable
+          - beta
           - nightly
 
         features:
@@ -64,8 +65,9 @@ jobs:
         # Test on stable, MSRV, and nightly.
         # Failure is permitted on nightly.
         rust:
-          - stable
-          - 1.60.0
+          # TODO update when stable Rust 1.65 is out.
+          #- stable
+          - beta
           - nightly
 
         features:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove IpAddress::Unspecified
 - When sending packets with a raw socket, the source IP address is sent unmodified (it was previously replaced with the interface's address if it was unspecified).
 - Fix enable `defmt/alloc` if `alloc` or `std` is enabled.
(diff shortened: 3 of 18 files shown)


gopakumarce:
@Dirbaio I have this pull request into your branch gat-device, I was using your branch (with very useful device GAT changes!) and ran into this, hence raising this into your branch hoping it will get into master sometime soon

Dirbaio:
Why do you need to create a socketset with sockets that outlive it? usually the socketset lives for the entire program's duration, which prevents this issue.

gopakumarce:
> Why do you need to create a socketset with sockets that outlive it? usually the socketset lives for the entire program's duration, which prevents this issue.

@Dirbaio the socket-set as a parameter passed to iface.poll() is a nice new addition, and the user of the iface "knowing" that only certain sockets need to be polled is also quite reasonable IMO, in which case a socketset that lives for the entire programs duration becomes restrictive. Also I think its semantically its a lil wierd to have a "set" into which one can add elements but cant take out of it, and the changes are harmless I believe so why not :)

Dirbaio:
> the user of the iface "knowing" that only certain sockets need to be polled is also quite reasonable IMO

You can't do that as an optimization, you'll break things. Any socket that exists might need polling at any time. Eitner because it receives packets, or because it has internal timers that fire.

It's not a generic "socket set", it's THE socket set of your TCP stack. You wouldn't create the Interface from scratch every time you want to send some packets, right? Same for the SocketSet.

> the changes are harmless I believe so why not :)

The cost is to have 2 lifetimes, which increases complexity substantially IMO. We have to explain the user which is which, and the user has to decide whether to "unify" them (`SocketSet<'a, 'a>`) or keep them separate in their structs which is not trivial.

Also note that this still doesn't solve all possible lifetime problems. Before the socket and slice lifetimes had to be the same. This only allows the slice lifetime to be different, but it's still forcing all socket lifetimes to be the same, so you can still cook up an example where you want to put a short-lived socket in a SocketSet mixed with other long-lived sockets and you can't.

gopakumarce:
> > the user of the iface "knowing" that only certain sockets need to be polled is also quite reasonable IMO
> 
> You can't do that as an optimization, you'll break things. Any socket that exists might need polling at any time. Eitner because it receives packets, or because it has internal timers that fire.
> 
> It's not a generic "socket set", it's THE socket set of your TCP stack. You wouldn't create the Interface from scratch every time you want to send some packets, right? Same for the SocketSet.
> 
> > the changes are harmless I believe so why not :)
> 
> The cost is to have 2 lifetimes, which increases complexity substantially IMO. We have to explain the user which is which, and the user has to decide whether to "unify" them (`SocketSet<'a, 'a>`) or keep them separate in their structs which is not trivial.
> 
> Also note that this still doesn't solve all possible lifetime problems. Before the socket and slice lifetimes had to be the same. This only allows the slice lifetime to be different, but it's still forcing all socket lifetimes to be the same, so you can still cook up an example where you want to put a short-lived socket in a SocketSet mixed with other long-lived sockets and you can't.

@Dirbaio on second thoughts and more experimentation, I agree. What I figured is that the use of smoltcp with nostd is very tight / constrained - not because of anything smoltcp is doing, just because needing an &mut u8 in a ManagedSlice that is kept alive introduces all kinds of problems with Rust, at that point basically each Socket can just be an individual object, it cant be added to even a fixed size "array of sockets"  because that would need mutably borrowing that array multple times. I was "hoping" that I can have a no-std version where I can have a pre-defined array of sockets etc.. and hence was trying to optimize for that. I will close this PR since I realize thats not possible.

But the medium-ethernet being needed to compile medium-ip is a problem, ill send an MR for that seperately, I hope you can pick just that

PR #697, opened 2022-10-22
[DO NOT MERGE] Allow a type implementing `phy::Device` to contain borrowed data

ebarnard:
Previously, a limitation of Rust's type system required a device to
implement `phy::Device<'a>` for all lifetimes, including `'static`, to
be able to call methods on `iface::Interface`. This prevented types
implementing `phy::Device<'a>` from containing any borrowed data.

Now that Rust supports GATs (generic associated types), it is possible
to lift this restriction and allow implementations of `phy::Device<'a>`
which contain borrowed data.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v2
         if: github.event_name != 'pull_request_target'
       - run: sed -n 's,^rust-version = "\(.*\)"$,RUSTUP_TOOLCHAIN=\1,p' Cargo.toml >> $GITHUB_ENV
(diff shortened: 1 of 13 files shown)


ebarnard:
This is not yet ready to merge. It requires GATs in stable which is set to be released in Rust 1.65 on 3rd Nov.

adamgreig:
I haven't checked the code diffs, but this sounds similar to https://github.com/smoltcp-rs/smoltcp/pull/572 ?

ebarnard:
It's exactly the same. I even looked at the list of open PRs before starting this. Oh well.

Closing in favour of the existing #572.

PR #698, opened 2022-10-23
Split interface into multiple files

thvdveld:
Currently, `src/iface/interface.rs` is 5178 lines. I started splitting the interface because I think it was becoming to big when working on RPL. I think splitting it the way I did is currently most logical, however, I'm sure I missed some things. What do you all think?


Dirbaio (approved):
LGTM. 

I too think that splitting by protocol makes the most sense, even if it isn't a 10)% "clean" split (for example some IP features have to "care" about Ethernet stuff...). I don't think there's much we can do about it, that's the reality of how the protocols work... :shrug: 


bors r+

PR #706, opened 2022-11-06
Fix medium-ip not compiling complaining of needing EthernetAddress

Dirbaio:
#703 rebased

cc @gopakumarce


gopakumarce:
Thx @Dirbaio

PR #712, opened 2022-11-29
add more tests in CI

thvdveld:
IPv4 fragmentation was not covered in CI, neither was 6LoWPAN without fragmentation.


Dirbaio (approved):
Thanks!

bors r+

PR #714, opened 2022-12-07
Fix PacketBuffer contig_window check when empty

### LINKED ISSUE #713: PacketBuffer can incorrectly report Exhausted with certain access patterns
If the `RingBuffer` inside a `PacketBuffer` is manipulated so that its `read_at` pointer is near the end of the buffer, `PacketBuffer::enqueue` will return `Exhausted` despite the ringbuf being **completely empty**.

This can happen during normal usage, especially when sending packets that are near the buffer size.

Here's an example:
```rust
use smoltcp::storage::{PacketBuffer, PacketMetadata};

fn main() {
    let mut data_buf = [0u8; 16];
    let mut meta_buf: [PacketMetadata<char>; 1] = [PacketMetadata::EMPTY; 1];

    const DUMMY_DATA: [u8; 15] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];

    let mut packet_buf = PacketBuffer::new(meta_buf.as_mut_slice(), data_buf.as_mut_slice());
    packet_buf
        .enqueue(15, 'a')
        .unwrap()
        .copy_from_slice(&DUMMY_DATA);

    // Remove that packet and check it
    let (h, d) = packet_buf.dequeue().unwrap();
    assert_eq!(h, 'a');
    assert_eq!(d, DUMMY_DATA);

    // We are now empty
    assert!(packet_buf.is_empty());

    // Insert a new packet - this should work, because we've got plenty of space
    packet_buf.enqueue(15, 'a').unwrap(); // surprise!
}
```

This fails on the final unwrap (using `smoltcp 0.8.2`):
```
$ cargo run
   Compiling smoltcp-test v0.1.0 (/Users/mjk/oxide/smoltcp-test)
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/smoltcp-test`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Exhausted', src/main.rs:24:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is because `PacketBuffer::enqueue` [requires a contiguous window large enough to fit the packet](https://github.com/smoltcp-rs/smoltcp/blob/b7a7c4b1c56e8d4c2524c1e3a056c745a13cc09f/src/storage/packet_buffer.rs#L91-L96).

However, we've got plenty of space in the buffer (it's literally empty).

In fact, `Ringbuffer::enqueue_many_with` has [a special case to handle this](https://github.com/smoltcp-rs/smoltcp/blob/b7a7c4b1c56e8d4c2524c1e3a056c745a13cc09f/src/storage/ring_buffer.rs#L185-L189), but it's not called until _after_ the `contig_window` check, so it never gets a chance to execute.

We should probably add a similar special-case to `PacketBuffer::enqueue`, e.g. adding
```rust
if self.payload_ring.is_empty() {
    self.payload_ring.clear(); // resets read_at to maximize contiguous space
}
```

mkeeter:
Fixes #713 (which also includes a more detailed explanation)

### DIFF
--- a/src/storage/packet_buffer.rs
+++ b/src/storage/packet_buffer.rs
@@ -82,6 +82,12 @@ impl<'a, H> PacketBuffer<'a, H> {
             return Err(Full);
         }
 
+        // Ring is currently empty.  Clear it (resetting `read_at`) to maximize
+        // for contiguous space.
+        if self.payload_ring.is_empty() {
+            self.payload_ring.clear();
+        }
+
         let window = self.payload_ring.window();
         let contig_window = self.payload_ring.contiguous_window();
 
@@ -371,6 +377,14 @@ mod test {
         assert!(buffer.dequeue().is_ok());
         assert_eq!(buffer.enqueue(8, ()), Err(Full));
         assert_eq!(buffer.metadata_ring.len(), 1);
+    }
+
+    #[test]
+    fn test_contiguous_window_wrap() {
+        let mut buffer = buffer();
+        assert!(buffer.enqueue(15, ()).is_ok());
+        assert!(buffer.dequeue().is_ok());
+        assert!(buffer.enqueue(16, ()).is_ok());
     }
 
     #[test]


Dirbaio (approved):
LGTM. Nice find, thank you!

bors r+

PR #717, opened 2022-12-13
fix reversed udp server example in README

overheat:
I try this example and it respond with reversed chunks of the input indefinitely instead of respond "hello" to any incoming packet.


Dirbaio:
THanks!

Issue #458, opened 2021-04-01
Bogus ARP requests when no IP configured


Dirbaio:
When DHCP isn't up, the interface has an address of 0.0.0.0/0 which is actually a CIDR that matches all IPs out there. If you try to connect to a WAN IP in that state it actually sends ARP requests trying to resolve it. It seems we're missing a check that treats 0.0.0.0/0 specially? Alternatively Interface could have a `[Option<Cidr>]` so we can represent "don't have an IP" as an actual None.

![screenshot-2021-04-01_19-02-15](https://user-images.githubusercontent.com/1247578/113329356-bed69280-931d-11eb-8c9f-482fd26655a1.png)

jordens:
0.0.0.0/**32** might be less likely to be caught up like that. Or `None` as you say. In any case it seems that if the address is unspecified (be that 0.0.0.0 or `None`), most egress should be explicitly prevented.

Dirbaio:
Fixed in #719. 

Users are no longer supposed to set `0.0.0.0/0` as an IP address, instead they should set no IP addr at all. #719 makes it possible in nostd.

Issue #466, opened 2021-04-13
TCP sockets can transmit packets with wrong source IP


Dirbaio:
If there's a TCP socket open, and then the smoltcp changes IP address (due to eg being plugged into a different Ethernet network), the socket stays established and still transmits packets with the old source IP address.

No matter what, a host should *never* transmit packets with a source IP that's not its own :)

In the packet capture below you can see the DHCP server gives the address `10.42.0.59` to the smoltcp host, but it continues transmitting packets for the old `192.168.1.37` address. Eventually the socket times out and the application establishes a new one, which then gets the right source IP and communication returns.

![screenshot-2021-04-13_21-33-31](https://user-images.githubusercontent.com/1247578/114610339-11aa3580-9ca0-11eb-844e-64e7506c8a77.png)

I think the same bug can be triggered by explicitly calling `tcp_socket.connect` with a wrong source address. Maybe with UDP and other sockets too.

Not sure what the best solution is.

- Simply check in `Interface` that we have the source address, and fail emitting if not? The socket would still have to timeout, but at least we don't emit garbage into the network.
- Check in TcpSocket that emitting failed due to wrong source addr and change to Closed? Seems better because it allows the application to immediately notice and reconnect, but i'm not sure if there are any disadvantages.

This is somewhat related with #458

MathiasKoch:
@Dirbaio 

I am currently running into needing to fix this issue for my application.

My approach would absolutely be to implement 
> Check in TcpSocket that emitting failed due to wrong source addr and change to Closed? Seems better because it allows the application to immediately notice and reconnect, but i'm not sure if there are any disadvantages

I am quite new to the `smoltcp` codebase. Where would be the most correct place to compare `src_addr` for the interface vs packets?

MathiasKoch:
I think I've managed to check for a mismatched src addr in `Interface`, in order to not emit garbage, and instead return an error from `emit` (`respond` closure in interface). 

Next up is catching this in `TcpSocket` to call `set_state(State::Closed)`.. But how to go about that? The error type on `emit` is generic, so I can't match on the result of `emit` in `TcpSocket`, and `set_state()` is private, so I can't call that in `Interface`.. 

Which way would be the most correct to go about this?

MathiasKoch:
This diff essentially does what I want it to functionally, but I would suspect this is not the proper way to solve it? @Dirbaio ?

```rust
[...]

PR #719, opened 2022-12-20
Interface `ip_addrs` field from slice to `heapless::Vec`

davidedellagiustina:
As stated in the title.
One cargo test is still failing, needs to be looked into.

Moreover, for now the `heapless::Vec` used is fixed-size.


MabezDev:
Is there any particular reason you need/want this? What's wrong with the `ManagedSlice` method?

Dirbaio:
It's to fix #458 and some cases of #466.

The problem with `ManagedSlice` is that in the borrowed case it is fixed length, so you can't dynamically add/remove IPs. If you want to have "no IP" (for example for when DHCP is not up yet) you have to change it to `0.0.0.0` instead. This causes smoltcp to transmit garbage packets with source IP `0.0.0.0`, because it treats it like any other IP in the egress code.

The alternative would be to change egress to treat `0.0.0.0` as "not a real IP", but that's an ungly workaround. The root issue is that the IP list should be variable length even in nostd.

(This probably applies to other ManagedSlice's in Interface as well, these'd be better off as `Vec`s. For example the routes and neighbor cache. We'll change them in future PRs)

davidedellagiustina:
Exactly what @Dirbaio said. I'm still trying to understand how to fix the failing test though.

MabezDev:
Thanks! I'm pretty sure smoltcp & managed predate heapless so It'll be nice to eventually use heapless instead and phase out managed. I'm sure it would help for https://github.com/smoltcp-rs/smoltcp/issues/655.

davidedellagiustina:
Before forgetting, there is also an additional point that I briefly discussed with @Dirbaio in private: for now, I used a hardcoded constant as the maximum size of the `ip_addrs` vector, but this might restrict some use cases. We agreed on the fact that using a generic parameter would pollute the API in an undesirable way, and probably the best way would be adding some cargo features to allow for different (though predefined) sizes for such vector. But any further idea is welcome.

Dirbaio (approved):
LGTM, thank you!

Yes, we can work on making the max count customizable in the future.

PR #721, opened 2022-12-22
Make `AnySocket` object safe

tsoutsman:
Apologies, in #718, removing the sized bound in `downcast` and `downcast_mut` implicitly made `AnySocket` object unsafe as they don't have a `self` method. I had a bit of a brain fart.

Signed-off-by: Klim Tsoutsman <klim@tsoutsman.com>

### DIFF
--- a/src/socket/mod.rs
+++ b/src/socket/mod.rs
@@ -93,8 +93,12 @@ impl<'a> Socket<'a> {
 /// A conversion trait for network sockets.
 pub trait AnySocket<'a> {
     fn upcast(self) -> Socket<'a>;
-    fn downcast<'c>(socket: &'c Socket<'a>) -> Option<&'c Self>;
-    fn downcast_mut<'c>(socket: &'c mut Socket<'a>) -> Option<&'c mut Self>;
+    fn downcast<'c>(socket: &'c Socket<'a>) -> Option<&'c Self>
+    where
+        Self: Sized;
+    fn downcast_mut<'c>(socket: &'c mut Socket<'a>) -> Option<&'c mut Self>
+    where
+        Self: Sized;
 }
 
 macro_rules! from_socket {


Dirbaio:
Oh, whoops! So that's what `Sized` was for,

Hmm, but adding `where Self: Sized` to the methods won't let you call them on a `dyn AnySocket`. Doesn't this kinda defeat the point of using `dyn AnySocket`? 

Why do you want to use `dyn AnySocket` instead of just `Socket`? The `Socket` enum is already sort of a "dyn socket" which can be any actual socket type.

tsoutsman:
> Hmm, but adding `where Self: Sized` to the methods won't let you call them on a `dyn AnySocket`. Doesn't this kinda defeat the point of using `dyn AnySocket`?

In my use case I don't use the functions on a trait object:

```rust
static SOCKETS: Mutex<Vec<MySocket<dyn AnySocket>>> = Mutex::new(Vec::new());

#[repr(transparent)]
pub struct MySocket<T>
where
    T: AnySocket<'static> + ?Sized,
{
    pub(crate) inner: SocketSet<'static>,
    phantom_data: PhantomData<T>,
}

impl<T> Deref for MySocket<T>
where
    T: AnySocket<'static>,
{
    type Target = T;

    fn deref(&self) -> &Self::Target {
        AnySocket::downcast(self.inner.iter().next().expect("no socket in socket set").1)
            .expect("incorrect socket type")
    }
}

impl<T> DerefMut for MySocket<T>
where
    T: AnySocket<'static>,
{
    fn deref_mut(&mut self) -> &mut Self::Target {
        AnySocket::downcast_mut(
            self.inner
                .iter_mut()
                .next()
                .expect("no socket in socket set")
                .1,
        )
        .expect("incorrect socket type")
    }
}

pub fn add_socket<T>(&self, socket: T) -> Arc<Mutex<Socket<T>>>
where
    T: AnySocket<'static> + Send,
{
    let mut socket_set = SocketSet::new([iface::SocketStorage::default(); 1]);
    socket_set.add(socket);
    let socket_arc = Arc::new(Mutex::new(Socket::<T>::new(socket_set)));
    SOCKETS
        .lock()
        // SAFETY: Socket has a transparent representation, so the memory layout is identical
        // regardless of T. The Send bound on T ensures that transmuting to a type that
        // implements Send is sound.
        // NOTE: We are transmuting an arc, not the underlying type. That doesn't change the
        // safety requirements though.
        .push(unsafe { core::mem::transmute(socket_arc.clone()) });
    socket_arc
}
```

The `dyn AnySocket` is to satisfy the type system, the code never actually constructs a trait object. `SOCKETS` could be a vector of `MySocket<UdpSocket>`; it doesn't matter as the memory is still the same. A `dyn AnySocket` just better expresses the intentions of the code.

The `T` in the `Deref` and `DerefMut` implementations has an implicit `Sized` bound, so we can call `downcast` and `downcast_mut`. However, this means we can't call `deref` on a `MySocket<dyn AnySocket>` which makes sense as we don't know what to downcast it to.

Dirbaio:
Hmm... That transmute looks very cursed :sweat_smile:  Probably sound, but still...

One thing I noticed from your snippet and looking at https://github.com/theseus-os/Theseus/pull/742 is that you're creating one SocketSet for each socket, and doing one `poll()` call on the interface for each socket. This will NOT work. smoltcp assumes `poll()` is called with a SocketSet containing ALL sockets. If you don't, you will run into issues like: if you have two TCP sockets A and B, when polling with a SocketSet containing only A, any received packets for B will be replied to with a RST.

One way to do it is to make your Socket struct a "wrapper" that only contains a handle into the socket inside the SocketSet and borrow it out each time you want to do something with it. This is how [embassy-net does it](https://github.com/embassy-rs/embassy/blob/master/embassy-net/src/tcp.rs#L206-L208) for example. The drawback is you have to put the entire network stack within one giant mutex which is not great if you want to support multithreading/multicore. 

Another way would be to put each socket behind its own Mutex, and then on iface poll lock them all, collect them into a single SocketSet, then poll that. Would be quite annyoing though...

Back to the PR: the consequence is that `dyn AnySocket` is only usable for `upcast`, and can't be used with `downcast`. This defeats much of the point of `AnySocket` for `dyn`. I don't think there's a downside for code not using `dyn` though, so can't see why not merge.

Thanks!

bors r+

PR #724, opened 2022-12-22
Make doc-test `no_run` instead of `ignore`

### LINKED ISSUE #723: Test has ignore instead of no_run
Shouldn't the following doc-test have the `no_run` attribute instead of the `ignore`?
Because I think the doc test is just explaining what the endpoints should look like.
The signature in the doc-test is not correct either (does not mention the `Context` struct).

I'm asking because running `cargo test -- --ignored` fails because of this doc-test.

https://github.com/smoltcp-rs/smoltcp/blob/3455acb1fb22769b557d08b615a456e4cf66c821/src/socket/tcp.rs#L729-L748

thvdveld:
This fixes #723.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -731,8 +731,43 @@ impl<'a> Socket<'a> {
     /// The local port must be provided explicitly. Assuming `fn get_ephemeral_port() -> u16`
     /// allocates a port between 49152 and 65535, a connection may be established as follows:
     ///
-    /// ```rust,ignore
-    /// socket.connect((IpAddress::v4(10, 0, 0, 1), 80), get_ephemeral_port())
+    /// ```rust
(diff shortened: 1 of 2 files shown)


thvdveld:
Ah, I only tested the fix with `cargo test -- --ignored` (which does compile and run the tests fine). But now `cargo test` complains about the test not compiling.

Dirbaio:
Why is `compile_fail,no_run` better than `ignore`? 

IMO the right fix should be to make it compile, using "hidden" lines (prefixed with `#`) if necessary, then use just `no_run`.

thvdveld:
Yes indeed, that would have been the correct thing to do! I was just in a hurry then. Fixed it now. Also fixed the pretty print test.

Dirbaio (approved):
Thank you!

bors r+

PR #726, opened 2022-12-24
Assembler refactors

Dirbaio:
Some refactors trying to simplify the data structures and API for the assemblers. See individual commit messages for details.

@thvdveld since you wrote the original code: Could you take a look? Is there anything I might have missed?

One thing that's wrong for now is the offset correction. Previously, the PacketAssembler was initialized when receiving the first fragment, with the right offset correction, and non-first fragments were discarded if they arrived first. The new API has only `.get()` which creates the assembler if it doesn't exist, so now non-first fragments will still be processed, but with the wrong offset correction. This is a problem, because only the first packet has enough info to learn the offset correction, right? Perhaps what we're supposed to do is decompress the headers before writing them to the assembler?

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -1,15 +1,8 @@
 mod utils;
 
 use log::debug;
-use std::collections::BTreeMap;
 use std::os::unix::io::AsRawFd;
 use std::str::{self, FromStr};
-
-#[cfg(any(
-    feature = "proto-sixlowpan-fragmentation",
-    feature = "proto-ipv4-fragmentation"
-))]
-use smoltcp::iface::ReassemblyBuffer;
 
 use smoltcp::iface::{InterfaceBuilder, NeighborCache, Routes, SocketSet};
 use smoltcp::phy::{wait as phy_wait, Device, Medium};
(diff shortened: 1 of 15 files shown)


thvdveld:
> Perhaps what we're supposed to do is decompress the headers before writing them to the assembler?

That is actually the correct thing to do. I didn't do this in the beginning because then I would have lost information that it came from a 6LoWPAN message. However, I did it like that because it was all very new to me as well and didn't know better.

So, yes, we need to decompress it before adding it to the buffer. This will remove the complicated offset calculation I think.

thvdveld:
The thing with decompress first is that the UDP emit function needs to change. Because the emit function needs all the UDP data etc.

Dirbaio:
I believe this is ready to go. Offset correction is gone :rocket: 

@thvdveld what do you think?

thvdveld:
I think `emit_header` needs some work in `src/wire/udp.rs`. It needs to be able to correctly calculate the checksum (and needs the data for that). Other than that I think this is ready indeed.

Dirbaio:
Hmmm, if it's just for decompressing the packet for further internal use we should be able skip calculating+reverifying the checksum I think. I'll take a look.

Dirbaio:
It sets the checksum to 0, then later when processing the decompressed UDP packet, checksum 0 is [treated as correct due to not present](https://github.com/smoltcp-rs/smoltcp/blob/master/src/wire/udp.rs#L122).. This is why the tests and everything was already passing.

I think it's the behavior we want: not generate the checksum, not check it. 

The only weird thing is that if you *do* tell `emit_header` to calculate the checksum, it won't do it right because the UDP payload is not written yet. I've simply changed it to not be able to, and clarified in docs that it's for internal-packet-use only.

PR #727, opened 2022-12-27
Refactor reassembly for 6LoWPAN

thvdveld:
Changes made for 6LoWPAN for the assembler refactor. @Dirbaio, let me know what you think about it. Still needs some error handling, but this should work.

### DIFF
--- a/src/iface/fragmentation.rs
+++ b/src/iface/fragmentation.rs
@@ -89,6 +89,42 @@ impl<K> PacketAssembler<K> {
         self.expires_at
     }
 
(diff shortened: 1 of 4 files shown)


Dirbaio:
Looks good to me! I like the idea of handling all 6lowpan by decompressing and passing to `process_ipv6`.

I'm not convinced about using an assembler slot for non-fragmented packets though. This means that, for example, if all slots are in use, then no non-fragmented packets can get processed. I'll try to take a stab at it, perhaps decompressing to a stack buffer instead.

PR #732, opened 2023-01-04
Make IpEndpoints parse unspecified IPv6 IPs

fluxchief:
The previous solution only works for `Ipv6Address` or `Ipv6Cidr`, but not for `Ipv6Endpoint`.
It's not clear to me whether allowing `::` is something that was intentionally left out for `Ipv6Endpoint`s and only implemented for CIDRs by using the `is_cidr` parameter or if there were other thoughts behind that.

All existing tests pass and I added one to cover the new behavior.


Dirbaio (approved):
Thakns!

bors r+

PR #278, opened 2019-02-27
Implement peek functions for UDP sockets

pothos:
This needed a peek function for the packet buffer.

PR #734, opened 2023-01-13
Add peek and peek_slice functions to RawSocket

ngc0202:
This adds the `peek` and `peek_slice` functions to raw sockets, same as was added to UDP sockets in #278


Dirbaio:
LGTM. Could you add tests? Copying the ones from UDP should work.

Dirbaio (approved):
Thank you! 

bors r+

PR #738, opened 2023-01-16
Allow specifying different server/client DHCP ports

czocher:
The default linux dhclient utility allows to use DHCP with non-standard ports. This change adds support for those usecases.

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -149,6 +149,12 @@ pub struct Socket<'a> {
     /// Ignore NAKs.
     ignore_naks: bool,
 
+    /// Server port config
+    pub(crate) server_port: u16,
+
+    /// Client port config
+    pub(crate) client_port: u16,
+
     /// A buffer contains options additional to be added to outgoing DHCP
     /// packets.
     outgoing_options: &'a [DhcpOption<'a>],
@@ -186,6 +192,8 @@ impl<'a> Socket<'a> {
             receive_packet_buffer: None,
             #[cfg(feature = "async")]
             waker: WakerRegistration::new(),
+            server_port: DHCP_SERVER_PORT,
+            client_port: DHCP_CLIENT_PORT,
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/socket/dhcpv4.rs
```diff
@@ -247,6 +251,18 @@ impl<'a> Socket<'a> {
         self.ignore_naks = ignore_naks;
     }
 
+    /// Set the server/client port
+    ///
+    /// Allows you to specify the ports used by DHCP.
+    /// This is meant to support esoteric usecases allowed by the dhclient program.
+    pub fn set_ports(&mut self, ports: (u16, u16)) {
```
czocher: I'm not sure if this should be done as a single method or two separate ones `(set_server_port`, `set_client_port`). We could also provide a structure with meaningful field names instead of a tuple. What do you think?

src/socket/dhcpv4.rs
```diff
@@ -247,6 +251,18 @@ impl<'a> Socket<'a> {
...
+        self.ports = ports;
+    }
+
+    pub fn get_ports(&self) -> (u16, u16) {
```
Czocher, line 262: Required to perform the test in `ipv4.rs`. I can also make the field `pub` if a getter is not what we'd like.

src/socket/dhcpv4.rs
```diff
@@ -1011,6 +1050,41 @@ mod test {
         }
     }
 
+    #[test]
+    fn test_bind_different_ports() {
```
czocher, line 1055: Any other tests you'd like to see?

src/socket/dhcpv4.rs
Dirbaio: Single method is fine. 

I'd do separate args instead of a tuple though, so the argument names tells which is which: `fn set_ports(&mut self, server_port, client_port)`. Same for the `ports` field in the struct.

src/socket/dhcpv4.rs
Dirbaio: no, this one's OK!

src/socket/dhcpv4.rs
czocher: Done!

Dirbaio (approved):
Thank you!

bors r+

PR #744, opened 2023-02-06
Add support for rebinding to DHCP implementation

JarredAllen:
The current DHCP implementation does not include the REBINDING state as detailed in the spec. I propose modifying the REBINDING state to send request packets until the REBINDING time begins, and then broadcast request packets once the REBINDING time begins.

I added assertions for broadcast request packets to the existing unit tests, which pass. I also validated on my setup that it sent a broadcast request at the T2 time my DHCP server sends. I think this should be sufficient validation, but I can provide more if requested.

There were a couple edge cases of the DHCP protocol pertaining to this change for which I couldn't find a specification in the standards. I documented the decisions I made in those cases, but I'm not attached to what I did (my setup hits none of them) so I can change it to something else if you prefer.

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -81,8 +81,20 @@ struct RenewState {
 
     /// Renew timer. When reached, we will start attempting
     /// to renew this lease with the DHCP server.
-    /// Must be less or equal than `expires_at`.
+    ///
+    /// Must be less or equal than `rebind_at`.
     renew_at: Instant,
+
+    /// Rebind timer. When reached, we will start broadcasting to renew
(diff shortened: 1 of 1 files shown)


JarredAllen:
Hey, are there any updates on this? Thanks for this library, it's great to use and really helpful! But I've been hitting issues caused by this part of the DHCP spec being missing, so I'd love it if we could get this PR in soon. Thanks!

JarredAllen:
(force push done to rebase onto current changes)

thvdveld:
@Dirbaio this looks good to me.

Dirbaio (approved):
Thank you!

bors r+

PR #745, opened 2023-02-08
Modify `hardware_addr` and `neighbor_cache` to be not `Option`

thvdveld:
Since `neighbor_cache` is now using `heapless::LinearMap` it doesn't need to be an `Option` any more. 

It's also possible to just make the `hardware_address` not `Option`.

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -240,9 +240,8 @@ pub struct InterfaceInner {
     rand: Rand,
 
     #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
-    neighbor_cache: Option<NeighborCache>,
-    #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
-    hardware_addr: Option<HardwareAddress>,
+    neighbor_cache: NeighborCache,
+    hardware_addr: HardwareAddress,
     #[cfg(feature = "medium-ieee802154")]
     sequence_no: u8,
     #[cfg(feature = "medium-ieee802154")]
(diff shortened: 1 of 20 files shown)


src/iface/interface/mod.rs
```diff
@@ -474,25 +474,16 @@ impl Interface {
...
-                config
-                    .hardware_addr
-                    .expect("hardware_addr required option was not set"),
-            ),
+            Medium::Ieee802154 => config
+                .hardware_addr
+                .expect("hardware_addr required option was not set"),
+
+            #[cfg(feature = "medium-ip")]
+            Medium::Ip => unreachable!(),
```
Dirbaio: This is not unreachable. 

If the user enables both `medium-ethernet` and `medium-ip` features, we want the user to be able to create both Ethernet and IP medium interfaces, choosing at runtime. (an example use case is if you have a device with both Ethernet and mobile connectivity).

This is the reason the `hardware_addr` was an `Option`. 

I agree the current solution is weird though. Perhaps it makes sense to make `HardwareAddress` be an Enum with `Ethernet`, `Ieee802154` and `Ip` cases, where `Ip` is empty? (And move it from `wire` to `iface`, because that makes it very iface-specific)

thvdveld: Indeed, sounds like the best solution at the moment.

Dirbaio:
Hmm I think now that HardwareAddress is valid in all cases, we could remove the `#[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]` everywhere? If you only enable IP medium you'll have a dummy zero-sized HardwareAddress which should still compile down to nothing.

Passing a `HardwareAddress` in `new` should be mandatory (no Option), and `new` should check the medium matches.

src/iface/interface/mod.rs
thvdveld: I think this is now resolved in ddbb967

thvdveld:
@Dirbaio any update on this? :smile:

Dirbaio (approved):
LGTM! Thank you and sorry for the delay.

bors r+

Dirbaio:
rebased!

bors r+

PR #746, opened 2023-02-08
Fix parsing of link layer address for ndiscoptions

thvdveld:
Since c9f3a1c6ffa0b1044a3296cfe150bb757efcb0b3, the ndisc options are parsed instead of just calling `link_layer_addr()` to get the link layer address. This is correct, however, the parse function only allowed for link layer addresses of 6 bytes instead of allowing other lengths.

A test was also added to check for this in the future.

### DIFF
--- a/src/wire/ndiscoption.rs
+++ b/src/wire/ndiscoption.rs
@@ -433,14 +433,14 @@ impl<'a> Repr<'a> {
     {
         match opt.option_type() {
             Type::SourceLinkLayerAddr => {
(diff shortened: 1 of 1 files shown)


thvdveld:
Version 0.9.0 does not respond with Neighbour Advertisements for `medium-ieee802154`, so we should release 0.9.1 with this fix I think.

Dirbaio (approved):
Thanks!

bors r+

Dirbaio:
????

bors r+

PR #752, opened 2023-03-14
Change IpPacket to contain more IP headers

thvdveld:
`IpPacket` currently only allows to have an IP base header together with only just one other header. However, for my work on RPL, an IP packet can have the base IP header, a hop-by-hop header and an ICMPv6 header (or a UDP header).

I modified the `IpPacket` struct such that it has the base IP header and then a `heapless::Vec` for other headers. I'm not sure if this is a good solution, because the type size goes from around 180 to 592 bytes ... (which is why clippy fails)


thvdveld:
I think this was not the best solution...

PR #753, opened 2023-03-17
Refactor ip packet

thvdveld:
I refactored the `IpPacket` struct, such that we can add more extensions headers to it for IPv6

Previously it was not possible to create packets that had hop-by-hop options (for emitting). Reading hop-by-hop options was no problem though. However, I need to be able to emit hop-by-hop for RPL, which with these changes would be easier to add.

The only _downside_ of this, is that we need `alloc`, or we could change it such that `Ipv6HopByHop` has a heapless `Vec` instead. This would then increase the size of the `IpPacket` struct.

I think this approach is better than #752, since now the structure of the IP packets is better represented than in #752.


thvdveld:
I think I have figured out a better way of doing this. Instead of adding the hop-by-hop data to the IpPacket (in the processing stage of packets), the hop-by-hop data should just be added when in the dispatching stage of the packets.

PR #754, opened 2023-03-20
Better defmt for Instant, Duration and Ipv6Address

thvdveld:
This changes the defmt formatting of Instant, Duration and Ipv6Address. They now have the same display as fmt::Display.


Dirbaio (approved):
thanks!

bors r+

PR #758, opened 2023-03-29
Better impls defmt for some structs

thvdveld:
Sometimes, the defmt implementation was not readable when debugging. I changed it such that the defmt implementation matches the Display implementation (which was more readable).


Dirbaio (approved):
Thanks!

bors r+

PR #760, opened 2023-03-29
Bump minor version of heapless

GrantM11235:
Smoltcp uses `heapless::Vec::remove`, which was added in [version 0.7.15](https://github.com/japaric/heapless/blob/HEAD/CHANGELOG.md#v0715---2022-07-05)

There is also a newer version, [0.7.16](https://github.com/japaric/heapless/blob/HEAD/CHANGELOG.md#v0716---2022-08-09), but I don't know if smoltcp requires it


Dirbaio (approved):
thanks!

bors r+

PR #763, opened 2023-04-03
Add opcode and flag enums to dns export

wmcleish:
Added opcode and flag enums to dns export. As it currently stands, user is not able to access the enums which are parameters to DnsPacket's "set_opcode" and "set_flags" functions. This change will allow the user to import and use those enums to set the packet contents. If there is a more elegant way to include them in DnsPacket instead of being their own imports, please suggest what changes can be made. Refer to [issue 762](https://github.com/smoltcp-rs/smoltcp/issues/762)

Edit: Sorry for commit spam, rustfmt not working locally for some reason.

### DIFF
--- a/src/wire/mod.rs
+++ b/src/wire/mod.rs
@@ -251,7 +251,7 @@ pub use self::dhcpv4::{
 };
 
 #[cfg(feature = "proto-dns")]
-pub use self::dns::{Packet as DnsPacket, Repr as DnsRepr, Type as DnsQueryType};
+pub use self::dns::{Packet as DnsPacket, Repr as DnsRepr, Type as DnsQueryType, Opcode as DnsOpCode, Flags as DnsFlags};
 
 /// Parsing a packet failed.
 ///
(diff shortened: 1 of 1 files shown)


src/wire/mod.rs
```diff
@@ -251,7 +251,10 @@ pub use self::dhcpv4::{
 };
 
 #[cfg(feature = "proto-dns")]
-pub use self::dns::{Packet as DnsPacket, Repr as DnsRepr, Type as DnsQueryType};
+pub use self::dns::{
+    Flags as DnsFlags, Opcode as DnsOpCode, Packet as DnsPacket, Repr as DnsRepr,
```
Dirbaio:
```suggestion
    Flags as DnsFlags, Opcode as DnsOpcode, Packet as DnsPacket, Repr as DnsRepr,
```

Dirbaio (approved):
thanks! 

bors r+

wmcleish:
@Dirbaio is there any chance you could generate a draft release or something with this change so that we can link to that in our code while we wait for the next release? Our use case is relatively blocked until this change is made.

PR #765, opened 2023-04-07
6LoWPAN Hop-by-Hop Header compression

thvdveld:
This changes the wire representation for IPv6 Hop-by-hop Headers, such that they can be used together with 6LoWPAN (and later with RPL).

6LoWPAN has the same Extension Headers as IPv6, however, the next header field can sometimes be elided, and the length field does not have the same semantics as the definition in the IPv6 standard.

More information can be found here: https://datatracker.ietf.org/doc/html/rfc6282#section-4.2

I added an enum `PacketFormat` that is added to the `Header` struct. When the `proto-sixlowpan` feature flag is not used, then there is only one variant of the `PacketFormat`, so I think the compiler should be able to optimize this away.

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -268,7 +268,7 @@ impl InterfaceInner {
         self.process_nxt_hdr(
             sockets,
             ipv6_repr,
-            hbh_repr.next_header,
+            hbh_repr.next_header.unwrap(),
             handled_by_raw_socket,
             &ip_payload[hbh_repr.buffer_len()..],
         )
--- a/src/wire/ipv6hopbyhop.rs
+++ b/src/wire/ipv6hopbyhop.rs
@@ -1,6 +1,8 @@
 use super::{Error, Result};
 use core::fmt;
 
+use super::PacketFormat;
+
 pub use super::IpProtocol as Protocol;
 use crate::wire::ipv6option::Ipv6OptionsIterator;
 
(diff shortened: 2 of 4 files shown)


Dirbaio:
Hmm, this approach requires explicitly adding NHC support to every single extension header, which makes me a bit scared. It adds quite a bit of complexity to the wire code.

The compressed NHC format is the same for all extension headers, and the non-compressed is *almost* the same (the fragmentation header has "Reserved" in the length byte, lol? WHY?). This makes me wonder whether this should be handled at a "higher" layer the same for all extension headers, instead of adding support to each one.

How were you planning to integrate this into `iface`? It seems to me that for RX `decompress_sixlowpan` could decompress the NHCs so that the regular ipv6 code doesn't need to handle NHC. For tx we don't have a `compress`, so i'm not sure...

Another idea we could play with is to make the `wire` code for the extension headers deal only with the "payload", which is all bytes except the first 2 (next header, and length). The "payload" is the same in the compressed and uncompressed versions, this would leave up to the higher layer code to deal with the "linked list" (either uncompressed or NHC).

:S

thvdveld:
I agree that it changes the complexity of the extension headers. At first, I didn't want to change a lot of the wire representations. I think I'll change the extension headers in general like you said:

> Another idea we could play with is to make the wire code for the extension headers deal only with the "payload", which is all bytes except the first 2 (next header, and length). The "payload" is the same in the compressed and uncompressed versions, this would leave up to the higher layer code to deal with the "linked list" (either uncompressed or NHC).

This will make the handling in the `iface` a lot easier and transparent with regard to compression or not. Thanks a lot for the feedback!

thvdveld:
Opened #781, which addresses the comment.

PR #768, opened 2023-04-13
Add a justfile

thvdveld:
The justfile contains the same cargo checks and tests that are performed by the github actions. Many times opening a pull request will result in failed tests or checks because of all the feature flags. Therefore, I created this file which can be used with https://github.com/casey/just .

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,71 @@
+
+# The MSRV that smoltcp compiles with.
+MSRV := "1.65.0"
+
+# `cargo check` and `cargo test`, usually used to check if a PR will succesfully check everything.
+check-all: check test clippy check-nightly test-nightly
+
+# `cargo check`
+check:
+    # These feature sets cannot run tests, so we only check they build.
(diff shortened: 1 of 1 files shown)


crawford:
This seems like it will take a bit of work to keep in sync with the actions. Is there any way to operate from a single source of truth instead?

thvdveld:
I'm totally not familiar with GitHub Actions, so I'm not sure how I would do that. Probably we'll have to read out the feature combinations from a file and converting that to a matrix. But I'm not sure how to set that up.

crawford:
What about using [`act`](https://github.com/nektos/act)?

thvdveld:
I tried using that in the past, but I get the following: `/var/run/act/workflow/1: line 2: cargo: command not found`, linking to https://github.com/nektos/act/issues/107 . Not sure why that is failing.

Dirbaio:
+1 to having some way to run CI locally.

In other repos I've just placed a `ci.sh` in the root, and made GHA run that. So the same file is used for both GHA and local testing. 

Super low-tech, works fine. I'm not sure if it's worth making people install `just`, instead of using a plain old shell script.

This means we lose parallelism in the GHA runs because it'll all be 1 job instead of 45. However, we're already hitting the max limit of 20 concurrent jobs, so what GHA is doing is start 20 jobs, wait for them to finish, start the next 20, wait, start the last 5. Starting a job has quite a bit of latency, so in practice it's quite likely that a single job will be faster.

thvdveld:
Closing this and created a PR with `ci.sh` for GHA.

PR #770, opened 2023-04-13
6LoWPAN Routing Header compression

thvdveld:
Based on changes I made in #765.

This adds the compressed format of the IPv6 Routing header used for 6LoWPAN.


thvdveld:
Better solution in #781 .

PR #771, opened 2023-04-14
Use bash script in CI

thvdveld:
Use a bash script in CI. The advantage of this is that we can run this script locally as well, and thus check if a PR would pass the checks or not locally.


ci.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
```
crawford: Not all distributions install bash to this location.

```suggestion
#!/usr/bin/env bash
```

thvdveld: Changed it. Thanks!

Dirbaio (approved):
Thank you!

bors r+

thvdveld:
bors retry

thvdveld:
I didn't know this was important, but I think CI still needs for bors:
```diff
diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
index c87d092..eaa2c3d 100644
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,13 @@ on:
 name: Test
 
 jobs:
+  tests:
+    runs-on: ubuntu-22.04
+    needs: [check-msrv, test-msrv, test-stable, clippy]
+    steps:
+      - name: Done
+        run: exit 0
+
   check-msrv:
     runs-on: ubuntu-22.04
     steps:
```

PR #776, opened 2023-04-17
Add file descriptor support for tuntap devices

blechschmidt:
Android exposes a file descriptor to a TUN interface through the [VpnService](https://developer.android.com/reference/android/net/VpnService) interface. This commit adds support for operating on such a file descriptor directly.

### DIFF
--- a/src/phy/tuntap_interface.rs
+++ b/src/phy/tuntap_interface.rs
@@ -28,9 +28,21 @@ impl TunTapInterface {
     /// no special privileges are needed. Otherwise, this requires superuser privileges
     /// or a corresponding capability set on the executable.
     pub fn new(name: &str, medium: Medium) -> io::Result<TunTapInterface> {
-        let mut lower = sys::TunTapInterfaceDesc::new(name, medium)?;
-        lower.attach_interface()?;
+        let lower = sys::TunTapInterfaceDesc::new(name, medium)?;
         let mtu = lower.interface_mtu()?;
+        Ok(TunTapInterface {
+            lower: Rc::new(RefCell::new(lower)),
+            mtu,
+            medium,
+        })
+    }
+
+    /// Attaches to a TUN/TAP interface specified by file descriptor `fd`.
+    ///
+    /// On platforms like Android, a file descriptor to a tun interface is exposed.
+    /// On these platforms, a TunTapInterface cannot be instantiated with a name.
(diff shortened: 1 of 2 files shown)


src/phy/tuntap_interface.rs
```diff
@@ -37,6 +37,19 @@ impl TunTapInterface {
             medium,
         })
     }
+
+    /// Attaches to a TUN/TAP interface specified by file descriptor `fd`.
+    ///
+    /// On platforms like Android, a file descriptor to a tun interface is exposed.
+    /// On these platforms, a TunTapInterface cannot be instantiated with a name.
+    pub fn from_fd(fd: RawFd, medium: Medium, mtu: usize) -> io::Result<TunTapInterface> {
```
Dirbaio, line 44: why set the MTU manually? `SIOCGIFMTU` doesn't work on Android?

blechschmidt: Because the device name is not known when using the [VpnService builder](https://developer.android.com/reference/android/net/VpnService.Builder), so I wouldn't know the device name to supply as part of the `ifreq` struct to obtain the MTU through `SIOCGIFMTU`.

Dirbaio: aw, that's unfortunate. Oh well.

Dirbaio:
Adding support for this sounds good to me. However I'm not very comortable with all the options in `TunTapInterfaceDesc`. I suggest the following: 

- TunTapInterfaceDesc stores `mtu` (no Option), and no `ifreq`.
- move all the code from `attach_interface` and `interface_mtu` into `TunTapInterfaceDesc::new()` , since that only makes sense when using an ifname. It gets the MTU and stores it in `self.`
- `interface_mtu` simply returns it from `self`.

blechschmidt:
Thank you for the feedback. What is your opinion on d1d1d2368f90871f8c6cf2508bf5cc9af4f4d5d6? It basically implements what you suggested with minor differences:

- Instead of moving the code, I have changed the signatures of `attach_interface` and the original `interface_mtu` function. They are called in `TunTapInterfaceDesc::new` and are no longer public. If you prefer, I can also remove them entirely and simply copy their code to the `new` function.
- `medium` is no longer needed in `self`.

Dirbaio (approved):
looks good! Thank you!

bors r+

PR #781, opened 2023-04-24
IPv6 Extension Headers rewrite

thvdveld:
I rewrote the IPv6 Extension Headers. All those headers have the `Next Header` and `Header Length` fields in common. These are now accessible using `Ipv6ExtHeader`. I removed the IPv6HopByHopOption, because this header only contains options, which can be accessed using the `Ipv6OptionsIterator`.

I also added tests for 6LoWPAN extension headers, routing header and hop-by-hop header), replacing #765 and #770 .

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -244,9 +244,11 @@ impl InterfaceInner {
         handled_by_raw_socket: bool,
         ip_payload: &'frame [u8],
     ) -> Option<IpPacket<'frame>> {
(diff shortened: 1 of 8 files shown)


thvdveld:
I think this is an elegant solution compared to #765 and #770. I already use it this way with the 6LoWPAN extension headers and RPL. Also, I think this is now ready to merge.

thvdveld:
bors r+

(I hope you don't mind @Dirbaio)

PR #787, opened 2023-05-12
IPv4: Don't discard from unspecified src addresses

bjoernQ:
IPv4 packets from `0.0.0.0` get discarded early. This makes it impossible to create a DHCP server.

Not sure if this is the right location or even the right approach for this fix.

Maybe there is also a way to make this work without modifying the code?

### DIFF
--- a/src/iface/interface/ipv4.rs
+++ b/src/iface/interface/ipv4.rs
@@ -18,9 +18,9 @@ impl InterfaceInner {
         frag: &'a mut FragmentsBuffer,
     ) -> Option<IpPacket<'a>> {
         let ipv4_repr = check!(Ipv4Repr::parse(ipv4_packet, &self.caps.checksum));
-        if !self.is_unicast_v4(ipv4_repr.src_addr) {
-            // Discard packets with non-unicast source addresses.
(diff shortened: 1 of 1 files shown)


thvdveld:
I think this is fine. We just need to make sure that we don't transmit to the unspecified address. Which we check here:

https://github.com/smoltcp-rs/smoltcp/blob/master/src/iface/interface/mod.rs#L1647

bjoernQ:
Thanks! That's good news
Any pointer at where to add tests to make codecov happy? (Sorry, first contribution here)

thvdveld:
I think it's fine. I think the codecov setup is not yet complete. It should not fail CI in this case. I made some changes to CI in #789. Maybe if you can rebase on master, then we can check if CI still fails or not.

bjoernQ:
> I think it's fine. I think the codecov setup is not yet complete. It should not fail CI in this case. I made some changes to CI in #789. Maybe if you can rebase on master, then we can check if CI still fails or not.

Awesome, that worked! Thanks a lot

Dirbaio (approved):
thanks!

bors r+

PR #788, opened 2023-05-24
tcp: Perform a reset() after an abort()

mkj:
It is useful to wake wakers once the RST packet has been emitted, so async callers can wait before dropping the socket. Another option would be to just wake wakers explicitly, but a .reset() is simpler.

This is to enable https://github.com/embassy-rs/embassy/pull/1471 "embassy-net: Make TcpSocket::abort() async"

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -2231,6 +2231,11 @@ impl<'a> Socket<'a> {
         if self.state == State::Closed {
             // When aborting a connection, forget about it after sending a single RST packet.
             self.tuple = None;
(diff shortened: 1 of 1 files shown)


Dirbaio:
This changes observable behavior: it is currently possible to receive some data, call `abort()`, and read that data out from the socket buffer. `reset()` throws the data away. Could you change it to just trigger the waker instead?

Dirbaio (approved):
thanks!

bors r+

Dirbaio:
no you're not wtf

bors cancel
bors r+

PR #796, opened 2023-06-09
Improve tests using `rstest`

thvdveld:
`rstest` provides useful macros for making tests more generic. A place where this is useful is in the interface. Here, we need to test for different mediums: IP, Ethernet and IEEE802.15.4. Previously, we had a variable `MEDIUM` which was the medium that was used by the interface, and depending on the feature flags that were set at compile time a different medium was used. This was not ideal. I think `rstest` is a better fit for achieving the same goal. I think this is a first step into making the writing of tests simpler, such that we can increase the coverage by adding more tests.


thvdveld:
~For some reason, the test directory does not count anymore for coverage. This file had a coverage of 100%. This is now gone. Overall, coverage increased.~

https://github.com/taiki-e/cargo-llvm-cov/pull/251 does not count test files for the coverage.

Overall, the coverage has increased for IPv6/IPv4/IEEE8021.5.4/6LoWPAN.

PR #799, opened 2023-06-12
Pass the now time when creating the iface

thvdveld:
I think that the now time should be passed to the interface when creating it. For RPL for example, some timers need to be created (Trickle timer for DIO's or a timer for DIS). Passing the now time makes this very easy. Having an interface with `Instant::ZERO` is also not really _correct_.

For the tests, the now time is ZERO, because we don't have a source when not using `std`.

### DIFF
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -97,7 +97,7 @@ fn main() {
     };
     config.random_seed = rand::random();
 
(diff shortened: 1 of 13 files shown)


Dirbaio:
Makes a lot of sense, Thank you! :)

PR #803, opened 2023-06-22
iface: fix outdated docs on `Interface::new`

elenaf9:
The methods linked in the docs don't exist anymore.
Also I believe with #745 `neighbor_cache` and `hardware_address` are not optional anymore and thus a panic can not happen?

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -451,16 +451,6 @@ enum IgmpReportState {
 
 impl Interface {
     /// Create a network interface using the previously provided configuration.
-    ///
-    /// # Panics
-    /// If a required option is not provided, this function will panic. Required
-    /// options are:
-    ///
-    /// - [ethernet_addr]
-    /// - [neighbor_cache]
-    ///
-    /// [ethernet_addr]: #method.ethernet_addr
-    /// [neighbor_cache]: #method.neighbor_cache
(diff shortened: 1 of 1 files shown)


thvdveld:
Thank you, you are right! However, there can still be a panic when the medium of the hardware address does not match the address of the device.

https://github.com/smoltcp-rs/smoltcp/blob/81fbe9194743c62fbe98ca1de401c15c322e3e40/src/iface/interface/mod.rs#L470-L474

Could you maybe add this information to the documentation?

src/iface/interface/mod.rs
```diff
@@ -283,7 +283,7 @@ pub struct Config {
 
     /// Set the IEEE802.15.4 PAN ID the interface will use.
     ///
-    /// **NOTE**: we use the same PAN ID for destination and source.
+    /// **NOTE**:kwe use the same PAN ID for destination and source.
```
thvdveld: A typo sneaked in I think.

elenaf9:
> Thank you, you are right! However, there can still be a panic when the medium of the hardware address does not match the address of the device.
> 
> https://github.com/smoltcp-rs/smoltcp/blob/81fbe9194743c62fbe98ca1de401c15c322e3e40/src/iface/interface/mod.rs#L470-L474
> 
> Could you maybe add this information to the documentation?

:+1: Done. Feel free to nitpick my wording.

thvdveld:
LGTM, thanks!

PR #807, opened 2023-06-26
icmpv6: truncate the packet to MTU

### LINKED ISSUE #638: ICMPv6 packets that don't transmit the entire payload can't get parsed
When ICMPv6 packets get parsed, the IPv6 packet field gets parsed as if it were a regular IPv6 packet. However, since ICMPv6 packets are usually sent with length of 1280 bytes to ensure it gets delivered, and also because of the ICMP protocol overhead, quite a bit of the original packet payload can get truncated, but the original payload length is sent anyway. Nonetheless, smoltcp attempts to parse and validate the packet, and fails, since the transmitted payload length is smaller than its advertised length.

This especially prevents it being used for e.g. path mtu discovery.

chayleaf:
This follows RFC4443 section 2.4. Closes #638


thvdveld:
Thank you!

chayleaf:
@thvdveld please see #808

PR #808, opened 2023-06-28
icmpv6: fix repr parsing panic with packets 40-43 bytes long

chayleaf:
this is a followup to #807, I read the code again and it turned out to be wrong


Dirbaio:
Could you add a previously-failing test?

chayleaf:
ok

Dirbaio (approved):
Thanks!

Issue #811, opened 2023-07-05
Expose the packet parsing and crafting infrastructure as a separate crate


HKalbasi:
Rust ecosystem lacks a mature packet parsing and crafting library similar to [libtins](http://libtins.github.io/) in C++, the best I found is the [packet crate](https://github.com/meh/rust-packet) which lacks some basic functionality like ARP packets (if a good library existed, I guess this crate would probably use it).

This crate has a great in house implementation of such functionality, which is zero copy, handles corner cases, is well maintained, ... . Would you accept a PR to exposing it as a separate crate, e.g. `smolpackets`? Currently most (all?) of that functionality is private, which is reasonable since it is not the intended job of the `smoltcp` crate. The new crate can still live in this repository, and I would expect no stability guarantees on it (it can even be published with the version of the `smoltcp` crate or in some auto release fashion) but it will still probably increase the maintenance cost a little.

thvdveld:
This idea came up more than once (matrix room of smoltcp). So I was working on this as well! I have a branch that is kind of ready. I'll create a pull request this afternoon.

thvdveld:
I discussed this a bit in the matrix room. The thing is that it would be difficult to maintain it. People might want to merge exotic protocols, however, these protocols might not fit the use case of smoltcp and there is not really someone that wants to commit to maintain it.

Then there is a second problem: smoltcp has a lot of feature flags. With these feature flags, we can keep the memory footprint low, since I think most of the people use it for embedded. If we have two crates, then smolwire could enable `proto-ipv6` and `proto-ipv4`, and smoltcp might only enable `proto-ipv6`, which will cause compile errors.

What is your use case? Couldn't you just use `smoltcp::wire` directly?

HKalbasi:
So it seems what I need is basically the `smoltcp::wire`. I thought I need `smoltcp::iface::ip_packet::IpPayload` and friends but it looks like those are thin internal wrappers over `smoltcp::wire`.

[...]

PR #812, opened 2023-07-06
split smoltcp in smolutils, smolwire and smoltcp

thvdveld:
This splits smoltcp into 3 crates.

## smolutils

This contains logic that is shared between smolwire and smoltcp. The logic that is shared is `time` and some parts of `phy`, such as `ChecksumCapabilities` and `Medium`. I don't think this crate should be released or something (I'm not sure since I never actually released big crates).

## smolwire

This crate contains logic for parsing and emitting IP packets. This crate should be released, since this contains useful parsing structures, not only for smoltcp but also for other use cases.

I named it smolwire since the structures were in the wire module. Maybe `smolpackets` is better, as suggested in #811 ?

## smoltcp

This is the TCP/IP library of smoltcp.

## TODO

- [ ] Check if imports and exports are correct
- [ ] Update READMEs and documentation
- [ ] (Something else?)


thvdveld:
Going to close this as this would add more maintenance overhead (for the smoltcp team) compared to a single crate.

PR #813, opened 2023-07-08
Add comment about using static lifetime for SocketSets with owned storage

paulmenage:
It wasn't obvious to me that `'static` can/should be the lifetime to use with a `SocketSet` if you're not borrowing storage; adding a comment will hopefully save others a little time fighting with lifetime-creep and interactions with thread constraints (thread closures can only borrow `'static` data).


Dirbaio (approved):
Thanks!

PR #814, opened 2023-07-13
RFC: listen on all ports

### LINKED ISSUE #809: Support listening on multiple ports with a single socket
For VPN/proxy-like programs, it's useful to listen to all traffic on all ports. For example, see [this library](https://github.com/eycorsican/netstack-lwip). I'm currently using it, but it would be nice to be able to use smoltcp for this use case as well.

How this would be implemented:

- For TCP sockets, very little has to change - port 0 should listen on all ports instead of being forbidden (the address tuple will be changed after accepting a packet so this is fine), the state machine is obviously socket-specific so the user will be responsible for creating additional sockets as necessary.
- For UDP sockets, it's somewhat harder as the implementation is very tied to being available only for a single port right now. Nonetheless, even just implementing sockets that listen for any traffic, then bind to the port where traffic arrived would probably be helpful. Alternatively, new code for wildcard UDP sockets with `recv_to` and `send_from` functions can be added, as UDP is a relatively simple protocol.
- For other sockets, nothing needs to be done as they are either port-specific or port-agnostic

chayleaf:
This would close #809. It can also allow rebinding the UDP socket port, but that isn't implemented (granted, it's a single-line change)

I've still to create examples, tests and documentation, I'm publishing it in this state to get some feedback

### DIFF
--- a/examples/multicast.rs
+++ b/examples/multicast.rs
@@ -111,7 +111,7 @@ fn main() {
         }
 
         let socket = sockets.get_mut::<udp::Socket>(udp_handle);
-        if !socket.is_open() {
+        if !socket.is_bound() {
             socket.bind(MDNS_PORT).unwrap()
         }
 
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -96,7 +96,7 @@ fn main() {
 
         // udp:6969: respond "hello"
         let socket = sockets.get_mut::<udp::Socket>(udp_handle);
-        if !socket.is_open() {
+        if !socket.is_bound() {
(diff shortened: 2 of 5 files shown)


raftario:
Just chiming in to show interest in the feature, not sure how to feel about the UDP implementation and method names though. Also I feel like it might make more sense to update `ListenEndpoint` to take an `Option<u16>` to reflect the existing behaviour for addresses and since binding to port 0 going from being an error to matching any port would be a massive footgun for people upgrading versions as the change wouldn't cause any compile errors.

Issue #816, opened 2023-07-18
IPSec Support


thegreathir:
Is there any plan for supporting `IPSec` protocol?
`IPSec` protocol is contained in a full implemented network stack like Linux and BSDs. I know that the key exchange part is delegated to user space programs by OSes, but without implementing the main part of the `IPSec` by network stack implementations, users are not able to use `IPSec` protocol.
So it would be nice if `smoltcp` support the implementation of this protocol.

PR #821, opened 2023-07-25
Add basic IPSec support to `wire`

thegreathir:
As mentioned #816 here, supporting the `IPSec` protocol kernel aspects would be nice.
The first step for supporting `IPSec` will be implementing `AH` and `ESP` related `Packet` and `Repr` in the `wire`.
I don't know if it is OK to merge the wire representations and then add complete support to other layers or not.


whitequark:
> I don't know if it is OK to merge the wire representations and then add complete support to other layers or not.

This is completely normal for smoltcp; the wire module is intended to be usable without the rest anyways.

thegreathir:
@Dirbaio What is your opinion?
Do you think this PR (and IPSec in general!) is valid and aligned with the purposes of the project?

src/wire/ip.rs
```diff
@@ -71,6 +73,8 @@ impl fmt::Display for Protocol {
             Protocol::Udp => write!(f, "UDP"),
             Protocol::Ipv6Route => write!(f, "IPv6-Route"),
             Protocol::Ipv6Frag => write!(f, "IPv6-Frag"),
+            Protocol::Esp => write!(f, "ESP"),
+            Protocol::Ah => write!(f, "AH"),
```
thvdveld: Could you change these to `IPsec-ESP` and `IPsec-AH`? I think that's clearer.

src/wire/ipsec_ah.rs
```diff
@@ -0,0 +1,279 @@
...
+        assert_eq!(
+            packet.integrity_check_value(),
+            &[0x27, 0xcf, 0xc0, 0xa5, 0xe4, 0x3d, 0x69, 0xb3, 0x72, 0x8e, 0xc5, 0xb0]
+        );
+    }
+
+    #[test]
+    fn test_construct() {
+        let mut bytes = vec![0xa5; 24];
+        let mut packet: Packet<&mut Vec<u8>> = Packet::new_unchecked(&mut bytes);
```
thvdveld: Why do you need to specify the type of `packet`?

src/wire/ipsec_ah.rs
```diff
@@ -0,0 +1,279 @@
...
+        packet.set_next_header(IpProtocol::Esp);
+        packet.set_payload_len(4);
+        packet.clear_reserved();
+        packet.set_security_parameters_index(0xba8bd060);
+        packet.set_sequence_number(1);
+        const ICV: [u8; 12] = [
+            0xaf, 0xd2, 0xe7, 0xa1, 0x73, 0xd3, 0x29, 0x0b, 0xfe, 0x6b, 0x63, 0x73,
+        ];
+        packet.integrity_check_value_mut().copy_from_slice(&ICV);
+        assert_eq!(&*packet.into_inner(), &PACKET_BYTES2[..]);
```
thvdveld: I think you could just compare with bytes here.

src/wire/ipsec_ah.rs
```diff
@@ -0,0 +1,279 @@
...
+        assert_eq!(Repr::parse(&packet).unwrap(), packet_repr());
+    }
+
+    #[test]
+    fn test_emit() {
+        let mut bytes = vec![0x17; 24];
+        let mut packet = Packet::new_unchecked(&mut bytes);
+        packet_repr().emit(&mut packet);
+        assert_eq!(&*packet.into_inner(), &PACKET_BYTES2[..]);
+    }
```
thvdveld: Could you add an assert for the `buffer_len` function?

src/wire/ipsec_esp.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn test_deconstruct() {
+        let packet = Packet::new_unchecked(&PACKET_BYTES[..]);
+        assert_eq!(packet.security_parameters_index(), 0xfb5128a6);
+        assert_eq!(packet.sequence_number(), 2);
+    }
+
+    #[test]
+    fn test_construct() {
+        let mut bytes = vec![0xa5; 8];
+        let mut packet: Packet<&mut Vec<u8>> = Packet::new_unchecked(&mut bytes);
```
thvdveld: Same as previous remark about the type.

src/wire/ipsec_esp.rs
```diff
@@ -0,0 +1,170 @@
...
+        packet.set_security_parameters_index(0xfb5128a6);
+        packet.set_sequence_number(2);
+        assert_eq!(&*packet.into_inner(), &PACKET_BYTES[..8]);
```
thvdveld: I think you can compare with bytes here.

src/wire/ipsec_esp.rs
```diff
@@ -0,0 +1,170 @@
...
+        assert_eq!(Repr::parse(&packet).unwrap(), packet_repr());
+    }
+
+    #[test]
+    fn test_emit() {
+        let mut bytes = vec![0x17; 8];
+        let mut packet = Packet::new_unchecked(&mut bytes);
+        packet_repr().emit(&mut packet);
+        assert_eq!(&*packet.into_inner(), &PACKET_BYTES[..8]);
+    }
```
thvdveld: Can you add an assert for the `buffer_len` function?

src/wire/mod.rs
```diff
@@ -272,6 +278,12 @@ pub use self::dns::{
     Repr as DnsRepr, Type as DnsQueryType,
 };
 
+#[cfg(feature = "proto-ipsec-ah")]
+pub use self::ipsec_ah::{Packet as IPSecAuthHeaderPacket, Repr as IPSecAuthHeaderRepr};
+
+#[cfg(feature = "proto-ipsec-esp")]
+pub use self::ipsec_esp::{Packet as IPSecEspPacket, Repr as IPSecEspRepr};
```
thvdveld: Maybe the letter `P` should be lowercase, then it matches the other types that have IP in their name.

Cargo.toml
thvdveld, line 1: Maybe you could add a `proto-ipsec` feature flag that enables `proto-ipsec-esp` and `proto-ipsec-ah`. Then use that one in the default feature set.

thvdveld (commented):
Thank you for the PR! Some nitpicks.

Cargo.toml
thvdveld: Or maybe don't add the feature to the default feature set since it is not used anywhere other than the wire module.

src/wire/ip.rs
hkalbasi: I changed the display output. Should the enum variant also be changed?

Cargo.toml
hkalbasi: I added the `proto-ipsec` feature. It is added in the default features to make CI run its tests, otherwise there is no need for it to be in the defaults as it is not actually implemented yet.

src/wire/ip.rs
thvdveld: Yes, maybe it should indeed!

Cargo.toml
thvdveld: You can add a line in the `ci.sh` file with the `proto-ipsec` feature.

src/wire/ip.rs
thegreathir: It's done

Cargo.toml
thegreathir: `proto-ipsec` has been added to `FEATURES_TEST` and `FEATURES_CHECK`

thegreathir: I think all of your comments are fixed now

thvdveld:
Thank you very much!

PR #827, opened 2023-08-07
refactor: move IPv6 HBH in own module

thvdveld:
The Ipv6HopByHopRepr was previously the same as an Ipv6ExtHeader. However, the RFC says that hop-by-hop options might update en route, which is not possible with the current implementation.

I added a representation for the Hop-by-Hop header, which has a heapless Vec of parsed options. This way, we can modify the options when needed.

The function of the Ipv6ExtHeader struct is now purely for parsing the ext header type and the length. It also returns a pointer to the data it holds, which must be parsed by the correct extension header representations.


whitequark:
Hm, why are we using both `heapless` and `managed`?

whitequark:
It looks like `heapless` was added in https://github.com/smoltcp-rs/smoltcp/commit/da1a2b2df0eafebb7fb92c00e56e88d533daa446, which is fair enough (I didn't investigate why the DNS server needs it; I assume it does) but in the core of the stack I'd rather only see `managed`.

thvdveld:
I think that'll be hard to pass the managed buffer to the parser no? The parse function should get a managed buffer from the interface, so then the interface inner should have a managed buffer, so then we need to implement a function that gives a managed buffer to the interface, which we then have to check if it was given by the user or not, and so on. I think it's simpler using the heapless::Vec in the wire module. I agree that using managed buffers for fragmentation and socket buffers is a must.

whitequark:
Yeah, on second thought you're right, go ahead.

PR #830, opened 2023-08-13
RFC: support for TCP Segmentation Offloading (TSO)

stlankes:
I tried to extend a driver to support TCP segmentation offloading (TSO). In this case, smoltcp has to create one header (Ethernert+IP+TCP) followed by multiple payloads. To support TSO, I extended the `DeviceCapabilities`  with `TsoCapabilities`. `TsoCapabilities` can describe, if TSO is supported for IPv4, IPv6 or both. In case TSO is supported, smoltcp emits packets with the maximum size of the remote window to the device driver.

Maybe my solution is a little bit naive. I am not an TCP expert. My test driver is a revised driver for [Virtio](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-1940001). Consquently, my VM uses virtio to forward messages to a tap device. Afterwards, the host forwards the packets over a 10 Gbit/s device to a server, which sends a response back. Both devices (tap and the ethernet device) supports TSO. I monitored the communication with Wireshark. It seems to work and I don't see any strange behavior.

I would appreciate comments to improve my work.

### DIFF
--- a/src/phy/mod.rs
+++ b/src/phy/mod.rs
@@ -304,7 +304,7 @@ pub struct DeviceCapabilities {
     /// TCP segmentation offload (TSO) support
     ///
     /// If the network device is able to support TCP segmentation offloading (TSO),
-    /// the stack forward the segementation to the harware to improve qthe performance.
+    /// the stack forward the segementation to the harware to improve the performance.
     pub tso: TsoCapabilities,
 }
 
--- a/src/phy/mod.rs
+++ b/src/phy/mod.rs
@@ -224,6 +224,39 @@ impl ChecksumCapabilities {
     }
 }
 
+/// Describes if the device supports TCP segment offloading (TSO).
+#[derive(Debug, Copy, Clone, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum TsoCapabilities {
+    /// Supports TSO for IPv4 and IPv6
+    Both,
+    /// Supports TSO for IPv4
+    Tso4,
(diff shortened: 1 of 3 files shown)


src/phy/mod.rs
```diff
@@ -267,6 +300,12 @@ pub struct DeviceCapabilities {
     /// If the network device is capable of verifying or computing checksums for some protocols,
     /// it can request that the stack not do so in software to improve performance.
     pub checksum: ChecksumCapabilities,
+
+    /// TCP segmentation offload (TSO) support
+    ///
+    /// If the network device is able to support TCP segmentation offloading (TSO),
+    /// the stack forward the segementation to the harware to improve qthe performance.
```
whitequark:
```suggestion
    /// the stack forward the segementation to the harware to improve the performance.
```

src/phy/mod.rs
```diff
@@ -224,6 +224,39 @@ impl ChecksumCapabilities {
     }
 }
 
+/// Describes if the device supports TCP segment offloading (TSO).
```
whitequark:
```suggestion
/// Describes if the device supports TCP segmentation offloading (TSO).
```

whitequark (commented):
The functionality looks fine to me. I'll leave style comments to @Dirbaio.

stlankes:
Ok, I found some issues. I have to think about it.

stlankes:
I close it. I have to think about it.

PR #832, opened 2023-08-22
Parse full IPv6 packet before processing

thvdveld:
With these changes, an IPv6 packet is fully parsed before being processed by other functions. I made these changes, because for RPL, IPv6 hop-by-hop headers and routing headers need to be modified after they are parsed such that packets can be forwarded/processed correctly. For example, if an hop-by-hop is present followed by a routing header, first the hop-by-hop needs to be processed before the routing header. With the implementation currently in main, it is not clear if the processing in of the headers is done in this order. It is also not easy to keep these headers around until they need to be forwarded.

With the changes, an IPv6 packet is parsed into an `Ipv6Packet` (not the one from wire). This packet contains a parsed IPv6 header, Hop-by-Hop header, Routing header, Fragmentation header and a payload, which may be TCP, UDP, ICMPv6, etc. With the changes, when a packet needs to be forwarded, the parsed IPv6 packet can now easily be modified and returned by the `process_ipv6` function, such that the packet is forwarded with the correct headers.

At first, I thought that the modifications I made will impact the binary sizes and make it way bigger, however, I was surprised to see that it's even a tiny bit smaller now.

Example server (before):
```
   text	  data	   bss	   dec	   hex	filename
1482212	 81272	   536	1564020	17dd74	server
```

Example server (after):
```
   text	  data	   bss	   dec	   hex	filename
1482060	 80800	   536	1563396	17db04	server
```

Example sixlowpan (before):
```
   text	  data	   bss	   dec	   hex	filename
1469581	 80256	   536	1550373	17a825	sixlowpan
```

Example sixlowpan (after):
```
   text	  data	   bss	   dec	   hex	filename
1469429	 79784	   536	1549749	17a5b5	sixlowpan
```

### DIFF
--- a/src/iface/interface/ipv4.rs
+++ b/src/iface/interface/ipv4.rs
@@ -122,7 +122,7 @@ impl InterfaceInner {
         }
 
         match ipv4_repr.next_header {
-            IpProtocol::Icmp => self.process_icmpv4(sockets, ip_repr, ip_payload),
+            IpProtocol::Icmp => self.process_icmpv4(sockets, &ipv4_repr, ip_payload),
 
             #[cfg(feature = "proto-igmp")]
             IpProtocol::Igmp => self.process_igmp(ipv4_repr, ip_payload),
@@ -140,8 +140,8 @@ impl InterfaceInner {
                 self.process_udp(
                     sockets,
                     meta,
-                    ip_repr,
-                    udp_repr,
+                    &ip_repr,
+                    &udp_repr,
                     handled_by_raw_socket,
                     udp_packet.payload(),
                     ip_payload,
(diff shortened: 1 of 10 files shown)


thvdveld:
I think that with these changes, we can remove a buffer for 6LoWPAN, where the decompression of incoming packets create an Ipv6Packet representation (not from wire), instead of writing bytes to a buffer and then process the bytes as IPv6.

Edit: still need to find a workaround for raw sockets, because then we need the raw data instead of the parsed data.

thvdveld:
> Therefore, extension headers must be processed strictly in the order they appear in the packet.

This PR breaks this, since we can't assume that the order is the same as the recommended order. Also, when the checksum of an UDP packet would be wrong, that check would fail and thus it would drop the packet, even though the packet might need to be forwarded instead.

I'm therefore closing this PR.

PR #834, opened 2023-08-22
Match the behavior of peek_slice to recv_slice

hunhoffe:
Previously, ```peek_slice()``` did not match the behavior of ```recv_slice()``` when there is a buffer wrapparound. I believe that by using the ```read_allocated()``` method, this will be fixed. I created an issue for this here: https://github.com/smoltcp-rs/smoltcp/issues/833

This is my first time trying to contribute to smoltcp so I've only tested locally a few ways. One is running my program that uses ```peek_slice()``` that previously failed (it now succeeds). I also ran:
```
cargo +stable test --no-default-features --features alloc,socket-tcp,proto-ipv4,medium-ethernet
``` 
And the tests completed succesfully. Please let me know if I need to do anything else.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1865,7 +1865,10 @@ impl<'a> Socket<'a> {
         let assembler_was_empty = self.assembler.is_empty();
 
         // Try adding payload octets to the assembler.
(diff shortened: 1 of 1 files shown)


Dirbaio:
Looks good! Can you add a test asserting the new behavior? ie test that it does return data past the wraparound.

Dirbaio (approved):
thanks!

PR #835, opened 2023-08-23
DHCP: Add an upper bound to the renew/rebind timeout in `RetryConfig`

JarredAllen:
# Description

This lets the user set an upper bound on the duration between times smoltcp sends a renew or rebind by setting a `max_renew_timeout` field to DHCP `RetryConfig` struct, which defaults to not imposing a cap unless the user explicitly sets one.

### Design questions

Use of this feature doesn't follow the official DHCP spec, but the `RetryConfig` struct already has a `min_renew_timeout` field which breaks the same spec by allowing the user to change the lower-bound imposed, so I think this is reasonable to add. Feel free to close this PR if you disagree.

I also added `#[non_exhaustive]` to the `RetryConfig` struct because this addition will necessarily be backwards incompatible and ineligible for inclusion until `v0.11.0`, but marking it as `#[non_exhaustive]` will allow anyone else in the future who wants more configuration options to add them without breaking backwards-compatibility. I can remove this attribute and let any future additions also require a major change instead.

I also didn't put any checks for the user putting `max_renew_duration` smaller than the existing `min_renew_duration`. If they do, then they'll silently get `max_renew_duration` as the chosen value every time. I think this is a reasonable response to an unlikely scenario, but I can do something else instead if you prefer.

# Validation

The existing unit tests cover logic in the case where no upper bound is set. I added a new unit test that covers the user setting a custom upper- and lower- bound to ensure that this behavior acts correctly.

I also have a firmware setup which uses this DHCP client that I switched to using this branch and set the maximum, and observed it capping at the maximum like expected.

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -114,12 +114,17 @@ enum ClientState {
 /// Timeout and retry configuration.
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
 pub struct RetryConfig {
     pub discover_timeout: Duration,
     /// The REQUEST timeout doubles every 2 tries.
     pub initial_request_timeout: Duration,
     pub request_retries: u16,
     pub min_renew_timeout: Duration,
+    /// An upper bound on how long to wait between retrying a renew or rebind.
+    ///
+    /// Set this to [`Duration::MAX`] if you don't want to impose an upper bound.
+    pub max_renew_timeout: Duration,
 }
 
 impl Default for RetryConfig {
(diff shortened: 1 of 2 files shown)


src/socket/dhcpv4.rs
```diff
@@ -1358,6 +1371,39 @@ mod test {
...
+    #[rstest]
+    #[case::ip(Medium::Ethernet)]
+    #[cfg(feature = "medium-ethernet")]
+    fn test_min_max_renew_timeout(#[case] medium: Medium) {
+        let mut s = socket_bound(medium);
+        // Set a minimum of 45s and a maximum of 120s
+        let config = RetryConfig {
+            max_renew_timeout: Duration::from_secs(120),
+            min_renew_timeout: Duration::from_secs(45),
+            ..s.get_retry_config()
```
thvdveld, line 1383: Since Default is implemented, could we not use that to complete the struct initialization? Then we don't need to add the `get_retry_config` function, unless you think that function is helpful to keep it.

JarredAllen: For this test code, you're right that `Default::default()` would work equally well to `get_retry_config`.

For my use-case, it is convenient to be able to set some parameters of the config while leaving others unchanged, which the `get_retry_config` method enables.

Dirbaio (approved):
thank you!

PR #837, opened 2023-08-26
reformatting project

ssrlive:
When I modified the file `Cargo.toml`, it caused many parts of this file to be modified automatically. I used vscode and configured the "automatically format when saving the file" option.

For the convenience of other developers, I reformatted the entire project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+.vscode/
+tmp/
+build/
 /target
 Cargo.lock
 *.pcap
--- a/src/wire/sixlowpan/iphc.rs
+++ b/src/wire/sixlowpan/iphc.rs
@@ -176,7 +176,7 @@ impl<T: AsRef<[u8]>> Packet<T> {
     /// Return the ECN field (when it is inlined).
     pub fn ecn_field(&self) -> Option<u8> {
         match self.tf_field() {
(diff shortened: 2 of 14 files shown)


Dirbaio:
We don't autoformat TOML files, the autoformatter's output is much less readable.

You can configure vscode to not format TOML like this:

```json
    "[toml]": {
        "editor.formatOnSave": false
    },
```

src/wire/sixlowpan/iphc.rs
```diff
@@ -176,7 +176,7 @@ impl<T: AsRef<[u8]>> Packet<T> {
     /// Return the ECN field (when it is inlined).
     pub fn ecn_field(&self) -> Option<u8> {
         match self.tf_field() {
-            0b00 | 0b01 | 0b10 => {
```
whitequark, line 179: This isn't a good reformatting. (Neither is the one below.)

.gitignore
```diff
@@ -1,3 +1,6 @@
+.vscode/
+tmp/
+build/
```
whitequark, line 3: Please don't add these, smoltcp's `.gitignore` is not the place to collect the configuration files of every editor known to man. Please use a global gitignore on your system.

PR #841, opened 2023-09-16
Don't error when calling `listen` again on the same local endpoint.

MabezDev:
Provided the local endpoint we are listening to is the same, it is safe to call `listen` multiple times. Fixes the doc comment for the function too.

This isn't much use for this in smoltcp, in embassy-net it's crucial when writing an acceptor loop. Without this, if you try and accept on a socket that's already listening you'll immediately get an error. The only way around this is to first `abort` the socket, but this means that incoming connections can actually get aborted between the `abort()` and the next `accept()`, as I have sadly experienced myself today :smiling_face_with_tear:.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -744,7 +744,7 @@ impl<'a> Socket<'a> {
 
     /// Start listening on the given endpoint.
     ///
-    /// This function returns `Err(Error::Illegal)` if the socket was already open
+    /// This function returns `Err(Error::InvalidState)` if the socket was already open
     /// (see [is_open](#method.is_open)), and `Err(Error::Unaddressable)`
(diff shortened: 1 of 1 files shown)


src/socket/tcp.rs
```diff
@@ -757,7 +757,12 @@ impl<'a> Socket<'a> {
         }
 
         if self.is_open() {
-            return Err(ListenError::InvalidState);
+            // If we were already listening, to same endpoint, there is nothing to do.
```
whitequark: Maybe add an explanation of why this particular behavior is desirable, based on the PR description?

MabezDev: Done :)

whitequark (approved):

thvdveld:
Thank you!

PR #844, opened 2023-09-25
Add function to explicitly set the flag given the u16 value.

i7solar:
Hi, 

I have a current use case to explicitly utilize the u16 values of the flag types of a TCP header rather than using the Boolean setter as this will speed up the current implementation of the logic I am writing.

This PR adds explicit functionality to allowing a flag to be set given a u16 value.


Dirbaio:
have you actually benchmarked that this makes a difference for your use case when compiling with --release? The compiler is very good at optimizing "useless" load/stores. It's likely that, with optimizations, setting flags individually is as fast as  `set_flags`.

Dirbaio:
I'm going to close this for now, feel free to reopen if you find this does speed up your code when building with --release.

PR #854, opened 2023-11-11
feat: add the retry time interval for sending syn messages

putyWang:


putyWang:
Due to the lack of a retry count and retry time interval for sending syn messages that establish links in the current TCP synSend state, an excessive number of syn link establishment messages will be sent during the link establishment process, and if there is no reply, it will cause messages to be sent continuously. Therefore, I have added the retry time interval (1, 2, 4...) for sending syn messages and the default maximum number of retransmissions (6 times) to the TCP socket timer

PR #855, opened 2023-11-13
Make `config` constants public

nilfit:
This allows dependent crates to use the constants directly instead of duplicating the system of features and conversion into constants.


Dirbaio (approved):
thank you!

PR #859, opened 2023-11-27
return RecvError::Truncated when calling recv_slice with a small buffer

### LINKED ISSUE #330: recv_slice for RawSocket, IcmpSocket and UdpSocket may lose data
`recv_slice` for [RawSocket](https://docs.rs/smoltcp/0.6.0/smoltcp/socket/struct.RawSocket.html#method.recv_slice), [IcmpSocket](https://docs.rs/smoltcp/0.6.0/smoltcp/socket/struct.IcmpSocket.html#method.recv_slice) and [UdpSocket](https://docs.rs/smoltcp/0.6.0/smoltcp/socket/struct.UdpSocket.html#method.recv_slice) may lose data if the provided buffer is smaller than the packet to dequeue.

I don't know if it should not dequeue and return an error, as it might not be possible to get a bigger slice when lacking an allocator, or stay with current behavior. However I think the documentation should be clearer as it seems to say nothing about that.

thvdveld:
When calling `recv_slice`/`peek_slice` in the udp, icmp and raw sockets, data is lost when the provided buffer is smaller than the payload to be copied (not the case for `peek_slice` because it is not dequeued).

With this commit, an `RecvError::Truncated` error is returned. In case of UDP, the endpoint is also returned in the error. In case of ICMP, the IP address is also returned in the error.

I implemented it the way @whitequark proposes it in #330. Data is still lost, but at least the caller knows that the data was truncated to the size of the provided buffer. As @whitequark says, it is preferred to call `recv` instead of `recv_slice` as there would be no truncation. If this is expected behaviour for `recv_slice`, we could close #330.

### DIFF
--- a/src/socket/icmp.rs
+++ b/src/socket/icmp.rs
@@ -61,12 +61,14 @@ impl std::error::Error for SendError {}
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum RecvError {
     Exhausted,
+    Truncated,
 }
 
 impl core::fmt::Display for RecvError {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
         match self {
             RecvError::Exhausted => write!(f, "exhausted"),
+            RecvError::Truncated => write!(f, "truncated"),
         }
     }
 }
@@ -130,8 +132,8 @@ impl<'a> Socket<'a> {
     /// Create an ICMP socket with the given buffers.
     pub fn new(rx_buffer: PacketBuffer<'a>, tx_buffer: PacketBuffer<'a>) -> Socket<'a> {
         Socket {
-            rx_buffer: rx_buffer,
-            tx_buffer: tx_buffer,
+            rx_buffer,
+            tx_buffer,
             endpoint: Default::default(),
             hop_limit: None,
             #[cfg(feature = "async")]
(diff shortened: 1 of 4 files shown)


whitequark (changes requested):
I'm not sure I like the design where there are two different ways of receiving metadata from the socket. This certainly takes "convenience" out of "convenience method".

I am not entirely sure what to do about this, but I would personally be happy to have `Error::Truncated` mean "there was a packet in the socket that was too big, it is lost, and you can move on to the next packet if you are *this* sure you only care about smaller ones". Anyone else should be using `recv`, probably.

thvdveld:
I was not sure about that part either. I'll remove the endpoint from the `Error::Truncated`.

thvdveld:
> here was a packet in the socket that was too big, it is lost

Do you mean that we also don't really need to copy part of the packet in the provided buffer and we should drop the full packet?

whitequark:
> Do you mean that we also don't really need to copy part of the packet in the provided buffer and we should drop the full packet?

Yep. These are convenience methods for the case where you're pretty certain you do not care at all about the bigger packets. If you sometimes want to process only the first part, or you want to execute some complex error handling code, they're not for you.

A note to that end in the docs may be useful.

thvdveld:
I made the changes and added a note in the documentation of the functions. 

Note that now the behaviour of these functions is different: no data is actually copied when the buffer is too small, whereas previously, data was still copied into the smaller buffer.

whitequark (approved):

whitequark:
> Note that now the behaviour of these functions is different: no data is actually copied when the buffer is too small, whereas previously, data was still copied into the smaller buffer.

This is a breaking change regardless of how much data is copied though, isn't it?

thvdveld:
I think only because of the addition of `Truncated` to the enums.

thvdveld:
Do you think we can close #330 now?

whitequark:
> I think only because of the addition of `Truncated` to the enums.

I mean that returning a different error, with or without payload, in case an overlong packet is detected, is a breaking change. Whether or not the buffer is filled out with data, at that point, doesn't make it any more breaking.

whitequark:
> Do you think we can close #330 now?

Yeah.

PR #861, opened 2023-11-28
Check IPv6 address after processing HBH

thvdveld:
Now, the destination address of the IP packet is checked if it is part of the interface addresses or the multicast addresses, only after processing the hop-by-hop header, as this header must be processed by every node. 

1. I modified the `has_multicast_group` function to also check IPv6 multicast addresses, such as the ALL_NODES, ALL_RPL_NODES (only when using RPL) and the IPv6 solicited address.
2. For some HBH options, an ICMPv6 parameter problem must be transmitted [RFC2460 Section 4.2](https://datatracker.ietf.org/doc/html/rfc2460#section-4.2). In some cases even when the destination address is a multicast address. Therefore also did point (3).
3. ICMPv6 messages were not transmitted when the destination address was a multicast address. I am not sure why we did it like that. When the destination address is multicast, I use the first IPv6 address from the interface. The selection of the correct IPv6 source address is something that still needs work in smoltcp.

[RFC8200 Section 4](https://datatracker.ietf.org/doc/html/rfc8200#section-4) says the following about hop-by-hop headers:

>  Extension headers (except for the Hop-by-Hop Options header) are not
   processed, inserted, or deleted by any node along a packet's delivery
   path, until the packet reaches the node (or each of the set of nodes,
   in the case of multicast) identified in the Destination Address field
   of the IPv6 header.

Therefore, when an hop-by-hop extension is present, we should process it. Then we check if the packet reached its destination. If not, we discard the packet if we are not using a routing protocol. In case of RPL, we route the packet to the correct next node.

### DIFF
--- a/src/iface/interface/igmp.rs
+++ b/src/iface/interface/igmp.rs
@@ -208,21 +208,6 @@ impl Interface {
 }
 
 impl InterfaceInner {
-    /// Check whether the interface listens to given destination multicast IP address.
-    ///
-    /// If built without feature `proto-igmp` this function will
-    /// always return `false`.
-    pub fn has_multicast_group<T: Into<IpAddress>>(&self, addr: T) -> bool {
-        match addr.into() {
-            IpAddress::Ipv4(key) => {
-                key == Ipv4Address::MULTICAST_ALL_SYSTEMS
-                    || self.ipv4_multicast_groups.get(&key).is_some()
-            }
-            #[allow(unreachable_patterns)]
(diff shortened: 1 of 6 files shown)


whitequark:
@thvdveld Unfortunately this PR deals squarely with the bits of IPv6 I do not have a good understanding of.

thvdveld:
@whitequark no problem!

Dirbaio (approved):

PR #863, opened 2023-11-30
Allow withholding the SYN|ACK packet by user code

eeeeeta:
In certain use cases, it's desirable to not send a SYN|ACK packet immediately after receiving a SYN -- for example, a TCP proxy that doesn't want to do so until it's connected to the end destination, because the outgoing connection might get refused. (Currently, you have to send a SYN|ACK and then reset the connection afterwards in this case, which is what we do in [onionmasq](https://gitlab.torproject.org/tpo/core/onionmasq).)

To fix this, add a simple `synack_paused` flag, controllable by user code, that withholds SYN|ACK packets in `SynReceived` state until it is unset.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -461,6 +461,9 @@ pub struct Socket<'a> {
     /// Used for rate-limiting: No more challenge ACKs will be sent until this instant.
     challenge_ack_timer: Instant,
 
+    /// If this is set, we will not send a SYN|ACK until this is unset.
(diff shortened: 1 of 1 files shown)


whitequark (commented):
That seems a rather niche feature and I'm not sure it's pulling its weight in e.g. the size of the already big `Socket` structure. Is this even available in any other network stack?

eeeeeta:
Fair enough!

PR #867, opened 2023-12-05
fix(tcp): don't send TCP RST with unsp. src addr

thvdveld:
When the destination address of the TCP packet was unspecified and the packet was not handled by any socket, smoltcp would transmit a TCP RST packet back. We should never transmit a packet with a source address unspecified.

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -1337,8 +1337,12 @@ impl InterfaceInner {
             }
         }
 
-        if tcp_repr.control == TcpControl::Rst {
-            // Never reply to a TCP RST packet with another TCP RST packet.
+        if tcp_repr.control == TcpControl::Rst
+            || ip_repr.dst_addr().is_unspecified()
+            || ip_repr.src_addr().is_unspecified()
+        {
+            // Never reply to a TCP RST packet with another TCP RST packet. We also never want to
+            // send a TCP RST packet with unspecified addresses.
             None
         } else {
             // The packet wasn't handled by a socket, send a TCP RST packet.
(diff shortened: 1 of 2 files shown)


src/iface/interface/mod.rs
```diff
@@ -1337,8 +1337,9 @@ impl InterfaceInner {
             }
         }
 
-        if tcp_repr.control == TcpControl::Rst {
+        if tcp_repr.control == TcpControl::Rst || !ip_repr.dst_addr().is_unspecified() {
```
Dirbaio: I think the condition is reversed? if dst addr is unspecified you don't want to reply.

```suggestion
        if tcp_repr.control == TcpControl::Rst || ip_repr.dst_addr().is_unspecified() {
```

thvdveld: Maybe we should just check for both src and dst. I don't think either should be unspecified.

thvdveld: Yes indeed, it was inverted. Weird that CI did not catch it. I'll see if I can add a test for that.

Dirbaio (approved):

thvdveld:
Wait, this was merged when CI failed, how could the merge queue do that. I thought this was not possible as it said merge when ready with ready being all checks successful.

whitequark:
> Wait, this was merged when CI failed, how could the merge queue do that. I thought this was not possible as it said merge when ready with ready being all checks successful.

This can happen if the merge queue is not set up correctly. I see `tests` has been skipped, is that the required check? Skipped checks are counterintuitively considered successful for merge queue purposes.

thvdveld:
I have no idea why that one was skipped. I don't see it in the workflow files. And I think all CI jobs are required: https://github.com/smoltcp-rs/smoltcp/pull/872#issuecomment-1849762059

PR #879, opened 2023-12-13
Remove generic in parse and emit functions

thvdveld:
I was using `cargo bloat` and I saw duplicate functions in wire. After this PR, the duplicates were gone. Removing the generics in the parse and emit functions prevents monomorphization of these functions. Preventing monomorphization reduces the binary size of the library.

`cargo bloat --release --example server --filter smoltcp`:
On main:        216.3KiB
On this branch: 202.5KiB

`cargo bloat --release --example sixlowpan --filter smoltcp`:
On main:        213.9KiB
On this branch: 199.0KiB

I know that this is for an example of `std`. However, the impact of code size will be smaller for embedded I suppose. The negative impact on user experience seems low since I didn't had to change much in the examples (only the ping example) and in the tests.

### DIFF
--- a/examples/ping.rs
+++ b/examples/ping.rs
@@ -208,7 +208,7 @@ fn main() {
 
             match remote_addr {
                 IpAddress::Ipv4(_) => {
-                    let icmp_packet = Icmpv4Packet::new_checked(&payload).unwrap();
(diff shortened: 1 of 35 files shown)


thvdveld:
Oh, will fix the commits tomorrow.

thvdveld:
I'll close this. I don't think this is actually a good idea any more.

PR #880, opened 2023-12-15
Use any_ip in process_arp

erthalion:
Currently, process_arp will not process a request if the ip_addrs doesn't contain its target. This could lead to weird situations, when any_ip address is used, but arp has to be handled somehow otherwise. An example use-case is a benchmarking tool that opens bunch of connections to the target using different addresses for each one.


thvdveld:
Thank you! I added a test for the feature.

erthalion:
Great, thanks!

PR #882, opened 2023-12-27
Ipv6RoutingHeader::clear_reserved: fix offsets for Type 2 routing headers.

luqmana:
Noticed this off-by-2, bytes 6 & 7 are part of the `Home Address` field not the Reserved portion.


thvdveld:
Thank you! I updated the emit test to make sure that this is not missed when changes are made to the logic.

luqmana:
> Thank you! I updated the emit test to make sure that this is not missed when changes are made to the logic.

Thanks!

PR #891, opened 2024-01-08
Add missing exports in wire for DNS

### LINKED ISSUE #890: dns Question type not exported
This makes it impossible to construct a `smoltcp::wire::DnsRepr` outside of smoltcp.

thvdveld:
Exporting more things from DNS in wire.

Fixes #890


bjorn3:
That was quick! Thanks!

PR #892, opened 2024-01-09
Prevent panic in update_servers for Domain Name System Socket

nick-we:
I wanted to connect my microcontroller with the wifi, but this library caused panics, because it found more than just one DNS Server. So I needed to change one line in the src/socket/dns.rs file with the update_servers method.

Instead of causing panic if the length of the servers argument exceeds DNS_MAX_SERVER_COUNT, it should take a slice of servers for the first DNS_MAX_SERVER_COUNT elements.

### DIFF
--- a/src/socket/dns.rs
+++ b/src/socket/dns.rs
@@ -167,7 +167,7 @@ impl<'a> Socket<'a> {
     ///
     /// Panics if `servers.len() > MAX_SERVER_COUNT`
     pub fn update_servers(&mut self, servers: &[IpAddress]) {
-        self.servers = Vec::from_slice(servers).unwrap();
+        self.servers = Vec::from_slice(&servers[..DNS_MAX_SERVER_COUNT]).unwrap();
     }
 
     /// Return the time-to-live (IPv4) or hop limit (IPv6) value used in outgoing packets.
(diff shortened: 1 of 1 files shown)


thvdveld:
Indeed, the documentation states that if the list of servers bigger than `DNS_MAX_SERVER_COUNT`, it will panic.

There are two options:
1. We prevent the panic and add logic for this.
2. We don't prevent the panic since it is written in the documentation that the list of servers cannot exceed `DNS_MAX_SERVER_COUNT`.

I'll leave the decision to @Dirbaio.

---

Your code might still panic. For example, it will panic when the length of the slice is smaller than `DNS_MAX_SERVER_COUNT`.

The update should be something like:
```rust
self.servers = Vec::from_slice(&servers[..servers.len().min(DNS_MAX_SERVER_COUNT)]).unwrap();
```

nick-we:
Ahh yes thanks for the correction!

thvdveld:
I think it's better that the there is an actual panic when the provided list of servers is bigger than `DNS_MAX_SERVER_COUNT`. With your change, the user of the API would never know if all servers where added to the list of DNS servers. The function is also clearly documented that a panic might occur.

PR #900, opened 2024-01-22
fix: accept any ip for IPv6

### LINKED ISSUE #899: How to make IPv6 handle any ip
https://github.com/smoltcp-rs/smoltcp/blob/9bd836ba6555c813b5b132f24316b02fa56510b8/src/iface/interface/ipv6.rs#L192-L194

Due to the addition of filtering, but no support similar to any_ip in IPv4, all unicast destination addresses that are not the current interface are now lost.

I don’t know if there is any way to solve it?

thvdveld:
Fixes #899 .

However, I'm not sure if this is the intended behaviour of Any IP. Should we check if the packet is routed locally, or should we just accept any packet?

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -193,8 +193,17 @@ impl InterfaceInner {
             && !self.has_multicast_group(ipv6_repr.dst_addr)
             && !ipv6_repr.dst_addr.is_loopback()
         {
-            net_trace!("packet IP address not for this interface");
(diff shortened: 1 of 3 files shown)


mhils:
@thvdveld: Thank you so much! 🍰 I was just looking into this issue when upgrading smoltcp in https://github.com/mitmproxy/mitmproxy_rs, what a fantastic timing. 🤩  Thank you for your work! ❤️

PR #902, opened 2024-02-01
Add VLAN support

DerFetzer:
This PR adds basic VLAN support.

There are probably some things missing since I am not yet that familiar with the codebase. I hope that you can point me in the right direction.

If you add the VLAN config option to the loopback example it works and you can see the tags in the output as well as in the recorded pcap file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,7 @@ defmt = ["dep:defmt", "heapless/defmt-03"]
 "proto-ipsec" = ["proto-ipsec-ah", "proto-ipsec-esp"]
 "proto-ipsec-ah" = []
 "proto-ipsec-esp" = []
+"proto-vlan" = ["medium-ethernet"]
 
 "socket" = []
 "socket-raw" = ["socket"]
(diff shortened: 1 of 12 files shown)


DerFetzer:
Ok, so first thing is to figure out how to integrate VLAN into existing tests for coverage.

thvdveld:
Since `src/wire/vlan/mod.rs` is not that big, could you move that file to `src/wire/vlan.rs`? Could you also add some documentation in that file where we can find information about the frame formats? I'm not so familiar with VLAN. Thank you!

DerFetzer:
@thvdveld I implemented your remarks a while back but forgot to write a new comment here.
What do you think about the current state of this PR? Thanks!

thvdveld:
I'm not really comfortable merging this as I'm not familiar with the VLAN/Ethernet implementation. Maybe @Dirbaio or @whitequark could take a look?

whitequark:
This is a large and complex PR and I won't be reviewing it as-is; splitting it into multiple parts, starting with the `wire` bits, may help.

Dominaezzz:
@DerFetzer are you still planning to update this?

stevefan1999-personal:
Yep, this would be pretty useful for me too. I need it as a foundation for building VXLAN in software stack. @Dominaezzz @DerFetzer mind if you guys could give it a fast track?

DerFetzer:
@Dominaezzz @stevefan1999-personal Feel free to take it from here since I do not have the use case anymore.
We chose to stick with lwIP in the end.

vic1707:
Would love to see this land, would love to help but completely ignorant over all this

PR #904, opened 2024-02-02
UDP: Store local and use local address in metadata

### LINKED ISSUE #903: UDP: Read and set local address on unbound sockets (IPV6_PKTINFO style functionality)
When UDP sockets are created and bound to a port but no IP address (the POSIX equivalent being being bound to `[::]:${PORT}`), applications often need to know which local address the packet was sent to, so that responses can be sent from the same address. This is requirement for implementing CoAP and (AIU) also QUIC -- and part of the embedded-nal-async (where [more extensive rationale](https://github.com/rust-embedded-community/embedded-nal/pull/73#issuecomment-1262041739) is given).

I can put in work on smoltcp to make this possible, but being new to the project I'll need a bit of guidance as to where to do that. My naive approach would be to add a field `localendpoint` to PacketMetadata, possibly behind a `pktinfo`ish flag. Odd thing is that this is really not phy-level information -- it'd just as well fit in `UdpMetadata`, but that is not `#[non_exhaustive]` (should it be?).

For actually making it functional on the incoming side, I think I'd just need to update socket::udp::Socket::process to propagate data that is available in the ip_repr down into the metadata. On the send side, this would be in its `dispatch()`, where we'd have to decide whether we just take the application's word that the source address is good, or whether we check if it really matches self.endpoint.addr / is a valid source address of cx.

### LINKED ISSUE #2516: net/UDP: Implement embedded-nal-async traits for UDP
Currently, embassy-net's UDP is providing its own custom set of methods and doesn't implement the embedded-nal-async traits.

I plan on providing a patch within the next weeks; initially, this is blocked by https://github.com/smoltcp-rs/smoltcp/issues/903.

chrysn:
Closes: https://github.com/smoltcp-rs/smoltcp/issues/903

As discussed there, this is a breaking change, and no feature flag is set for it right now.

I'm surprised that things were this easy. So far, my only tests were whether this builds, thus leaving this as a draft PR. I expect to test this with code that would resolve https://github.com/embassy-rs/embassy/issues/2516, at which point I'll mark it as ready.

### DIFF
--- a/src/socket/udp.rs
+++ b/src/socket/udp.rs
@@ -647,6 +647,13 @@ mod test {
         addr: REMOTE_ADDR.into_address(),
         port: REMOTE_PORT,
     };
+    fn remote_metadata_with_local() -> UdpMetadata {
+        // Would be great as a const once we have const `.into()`.
+        UdpMetadata {
+            local_address: Some(LOCAL_ADDR.into()),
+            ..REMOTE_END.into()
+        }
+    }
 
     pub const LOCAL_IP_REPR: IpRepr = IpReprIpvX(IpvXRepr {
         src_addr: LOCAL_ADDR,
@@ -823,7 +830,10 @@ mod test {
             PAYLOAD,
         );
 
-        assert_eq!(socket.recv(), Ok((&b"abcdef"[..], REMOTE_END.into())));
+        assert_eq!(
+            socket.recv(),
+            Ok((&b"abcdef"[..], remote_metadata_with_local()))
+        );
         assert!(!socket.can_recv());
     }
 
@@ -851,8 +861,14 @@ mod test {
             &REMOTE_UDP_REPR,
             PAYLOAD,
         );
-        assert_eq!(socket.peek(), Ok((&b"abcdef"[..], &REMOTE_END.into(),)));
-        assert_eq!(socket.recv(), Ok((&b"abcdef"[..], REMOTE_END.into(),)));
+        assert_eq!(
+            socket.peek(),
+            Ok((&b"abcdef"[..], &remote_metadata_with_local(),))
+        );
+        assert_eq!(
+            socket.recv(),
+            Ok((&b"abcdef"[..], remote_metadata_with_local(),))
+        );
         assert_eq!(socket.peek(), Err(RecvError::Exhausted));
     }
 
@@ -1025,7 +1041,7 @@ mod test {
             dst_port: LOCAL_PORT,
         };
         socket.process(cx, PacketMeta::default(), &REMOTE_IP_REPR, &repr, &[]);
-        assert_eq!(socket.recv(), Ok((&[][..], REMOTE_END.into())));
+        assert_eq!(socket.recv(), Ok((&[][..], remote_metadata_with_local())));
     }
 
     #[test]
--- a/src/socket/udp.rs
+++ b/src/socket/udp.rs
@@ -743,6 +743,17 @@ mod test {
         assert_eq!(socket.send_slice(b"abcdef", REMOTE_END), Ok(()));
     }
 
+    #[test]
+    fn test_send_with_source() {
+        let mut socket = socket(buffer(0), buffer(1));
+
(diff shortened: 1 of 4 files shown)


src/socket/udp.rs
```diff
@@ -8,20 +8,27 @@ use crate::socket::PollAt;
 #[cfg(feature = "async")]
 use crate::socket::WakerRegistration;
 use crate::storage::Empty;
-use crate::wire::{IpEndpoint, IpListenEndpoint, IpProtocol, IpRepr, UdpRepr};
+use crate::wire::{IpEndpoint, ip::Address, IpListenEndpoint, IpProtocol, IpRepr, UdpRepr};
```
thvdveld: You could just import `IpAddress` instead of `ip::Address`.

chrysn: Changed, thanks.

chrysn:
I'm having trouble finding the tests that need updating b/c they cover Debug output -- where do the test cases live? I failed to find anything matching `case_3_ieee802154` in the sources.

thvdveld:
We test the same test function for different cases by using a proc-macro. The name of the test is before ::case_something.

thvdveld:
Tests are failing because the `local_addr` in the `Meta` is `None` when calling `.into()` on it, which is probably not what you want in the test cases.

chrysn:
Tests are now updated and should pass.

chrysn:
The coverage report rightfully complains lines in my change that are not covered. I've added a test that does set an explicit correct source address -- so the coverage should be better now.

We *could* have a second test on the topic of unadressability, but that first needs a decision: what if an application sends an invalid (eg. the unspecified) address as a source address? If we just let it pass, there are some protocol violating outcomes of this (like the unspecified address), and some questionable outcomes (if an IP address's lease time expired between accepting a request and sending a response, can I still send from that address?), but in practice that'd need either deliberate crafting of such a message, or keeping an address for overly long (and yes the application can't know what "overly long" is, but usually it will be fine).

Thus, my **recommendation is that we accept whatever the application sets as a source address** -- after all, smoltcp AIU doesn't limit applications' privileges, and the application could just send odd data through raw sockets. The stack may still start checking at a later point, and then consider the datagrams unaddressable. For tests, that means that while we could add a test like

```patch
diff --git a/src/socket/udp.rs b/src/socket/udp.rs
index c1313aa..d64ff18 100644
--- a/src/socket/udp.rs
+++ b/src/socket/udp.rs
@@ -740,6 +740,16 @@ mod test {
             ),
             Err(SendError::Unaddressable)
         );
+        assert_eq!(
+            socket.send_slice(
+                b"abcdef",
+                UdpMetadata {
+                    local_address: Some(IpvXAddress::UNSPECIFIED.into()),
+                    ..REMOTE_END.into()
+                }
+            ),
+            Ok(())
+        );
         assert_eq!(socket.send_slice(b"abcdef", REMOTE_END), Ok(()));
     }
 ```

it would just test the current behavior that is not guaranteed, and thus doesn't really add value.

[edit: Sorry for the force pushes, didn't have cargo fmt checks in my pre-commit hooks; thankfully `git absorb` came in handy cleaning that up in all the individual commits]

chrysn:
I've now run this with some tests all the way to an embedded-nal-async application through https://github.com/embassy-rs/embassy/pull/2519. Can't claim I've covered every edge case (that will come as I extend those applications), but I'm confident enough with the current state to mark it as ready for review.

ivmarkov:
@Dirbaio @chrysn Is there anything outstanding in this PR prohibiting merging?

chrysn:
From my PoV this is ready for final review -- especially if my recommendation of https://github.com/smoltcp-rs/smoltcp/pull/904#issuecomment-1923604593 is what we go with, but also if not (because even then the signatures would not change, and merely behavior of applications whose input violates the UDP/IP protocol may change between sending an erroneous packet vs. reporting an error).

thvdveld:
Let's merge this!

PR #905, opened 2024-02-04
Update DHCP example

### LINKED ISSUE #783: examples/dhcp_client.rs is misleading/wrong
As of 0.9.1 and current main 8b0c521e2, the "dhcp_client.rs" example, when it gets a Configured event, attempts to set the interface's IPv4 address as follows:

```rust
fn set_ipv4_addr(iface: &mut Interface, cidr: Ipv4Cidr) {
    iface.update_ip_addrs(|addrs| {
        let dest = addrs.iter_mut().next().unwrap();
        *dest = IpCidr::Ipv4(cidr);
    });
}
```

This doesn't actually work in Smoltcp 0.9.1, because iter_mut() on a heapless::Vec iterates only over the _existing_ items, and is incapable of adding one if the Vec is currently empty. It always panics unwrapping the error from next(). This code worked in 0.8.x because iter_mut() on a mutable slice (which is what ManagedSlice derefs to) _does_ iterate over all elements (which were set to UNSPECIFIED if not currently in use).

For smoltcp 0.9.x, set_ipv4_addr should probably do something like:
```rust
                iface.update_ip_addrs(|addrs| {
                    addrs.push(IpCidr::Ipv4(config.address)).unwrap();
                });
```
and likewise on a Deconfigured event it should do addrs.clear().

jlogan03:
Fixes #783 

* Initialize interface with an unspecified (but populated) IP address so that there is something in place to mutate later
* Update set_ipv4_addr to clear existing addresses and push a new one, and add a small docstring

### DIFF
--- a/examples/dhcp_client.rs
+++ b/examples/dhcp_client.rs
@@ -7,7 +7,7 @@ use std::os::unix::io::AsRawFd;
 use smoltcp::iface::{Config, Interface, SocketSet};
 use smoltcp::socket::dhcpv4;
 use smoltcp::time::Instant;
-use smoltcp::wire::{EthernetAddress, IpCidr, Ipv4Address, Ipv4Cidr};
+use smoltcp::wire::{EthernetAddress, IpCidr, Ipv4Cidr};
 use smoltcp::{
     phy::{wait as phy_wait, Device, Medium},
     time::Duration,
--- a/examples/dhcp_client.rs
+++ b/examples/dhcp_client.rs
@@ -38,6 +38,9 @@ fn main() {
     config.random_seed = rand::random();
     let mut iface = Interface::new(config, &mut device, Instant::now());
 
+    // Initialize with an unspecified address
+    set_ipv4_addr(&mut iface, Ipv4Cidr::new(Ipv4Address::UNSPECIFIED, 0));
+
     // Create sockets
     let mut dhcp_socket = dhcpv4::Socket::new();
 
@@ -86,9 +89,10 @@ fn main() {
     }
 }
 
+/// Clear any existing IP addresses & add the new one
 fn set_ipv4_addr(iface: &mut Interface, cidr: Ipv4Cidr) {
     iface.update_ip_addrs(|addrs| {
-        let dest = addrs.iter_mut().next().unwrap();
-        *dest = IpCidr::Ipv4(cidr);
+        addrs.clear();
+        addrs.push(IpCidr::Ipv4(cidr)).unwrap();
     });
 }
--- a/examples/dhcp_client.rs
+++ b/examples/dhcp_client.rs
@@ -38,9 +38,6 @@ fn main() {
     config.random_seed = rand::random();
     let mut iface = Interface::new(config, &mut device, Instant::now());
 
-    // Initialize with an unspecified address
-    set_ipv4_addr(&mut iface, Ipv4Cidr::new(Ipv4Address::UNSPECIFIED, 0));
-
     // Create sockets
     let mut dhcp_socket = dhcpv4::Socket::new();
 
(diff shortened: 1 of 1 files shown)


jlogan03:
Converting to draft for the moment. During testing, found that while this allows the interface to obtain an IP address, if there are sockets already configured on the interface, they will not update properly unless there was already an IP address in place before they were initialized and before an IP address was obtained via DHCP.

Solution in my case is to initialize the interface with an UNSPECIFIED address and then replace that address with the new one. This is more similar to what the original example did, but with an added step in initialization.

Dirbaio:
we should not have UNSPECIFIED addrs in the interface, it's not correct. The way to make the interface have no addresses is to make the address vec empty.

> while this allows the interface to obtain an IP address, if there are sockets already configured on the interface, they will not update properly

what do you mean with "update"? if there's some issue in the socket logic we should fix that instead of workarounding it with UNSPECIFIED addrs.

jlogan03:
@Dirbaio @thvdveld , this is ready for review, but I'm not sure why tests are failing. Able to build & pass clippy locally on stable (1.76) and MSRV without specifying features. The test breakage is not related to these changes.

On main, the test breakage still occurs:
```
jlogan@jlogan-MS-7C56:~/git/smoltcp$ cargo +1.65.0 test --no-default-features --features default
   Compiling futures-macro v0.3.30
   Compiling regex-automata v0.4.5
   Compiling is-terminal v0.4.11
   Compiling rstest_macros v0.17.0
   Compiling rand_chacha v0.3.1
   Compiling smoltcp v0.11.0 (/home/jlogan/git/smoltcp)
   Compiling idna v0.5.0
   Compiling bitflags v1.3.2
   Compiling managed v0.8.0
   Compiling termcolor v1.4.1
   Compiling futures-timer v3.0.2
   Compiling unicode-width v0.1.11
   Compiling humantime v2.1.0
error[E0432]: unresolved imports `std::os::fd::AsFd`, `std::os::fd::AsRawFd`
  --> /home/jlogan/.cargo/registry/src/github.com-1ecc6299db9ec823/is-terminal-0.4.11/src/lib.rs:40:19
   |
40 | use std::os::fd::{AsFd, AsRawFd};
   |                   ^^^^  ^^^^^^^ no `AsRawFd` in `os::fd`
   |                   |
   |                   no `AsFd` in `os::fd`

error[E0603]: module `fd` is private
  --> /home/jlogan/.cargo/registry/src/github.com-1ecc6299db9ec823/is-terminal-0.4.11/src/lib.rs:40:14
   |
40 | use std::os::fd::{AsFd, AsRawFd};
   |              ^^ private module
   |
note: the module `fd` is defined here

Some errors have detailed explanations: E0432, E0603.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `is-terminal` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...

```

Dirbaio:
Seems the cause is https://github.com/sunfishcode/is-terminal/issues/36. ~~Easiest fix is probably to bump our MSRV to 1.66.~~ should be fixed now, that was fast wow.

To reproduce locally, delete Cargo.lock to get upated deps.

jlogan03:
Yep, that worked locally - if you want to rerun the failed workflows, it should be fine now

Dirbaio:
have you seen this comment though? https://github.com/smoltcp-rs/smoltcp/pull/905#issuecomment-1936716913

jlogan03:
> we should not have UNSPECIFIED addrs in the interface, it's not correct. The way to make the interface have no addresses is to make the address vec empty.
> 
> > while this allows the interface to obtain an IP address, if there are sockets already configured on the interface, they will not update properly
> 
> what do you mean with "update"? if there's some issue in the socket logic we should fix that instead of workarounding it with UNSPECIFIED addrs.

I replaced both usages of setting an UNSPECIFIED address with just clearing the addrs vec, and I can confirm that works at least in my particular application (a bare-metal stm32h743 on ethernet).

I didn't dig too far into why that was happening, and I was casually inferring that the sockets had some state related to the IP address available when they were created. Removing that initial assignment of an UNSPECIFIED address has no effect today, and a UDP socket created between initialization and when a real IP address is acquired works fine. My best guess is that because I'm testing immediately after flashing a chip that is already connected to ethernet, I may have been seeing the dhcp socket get deconfigured by the router due to the link going down, causing an UNSPECIFIED address to be pushed.

Dirbaio (approved):
thanks!

PR #906, opened 2024-02-07
Add BPF support for FreeBSD

jamwaffles:
Adds support for FreeBSD to `bpf.rs`. This was tested on a FreeBSD system with an [EtherCrab](https://github.com/ethercrab-rs/ethercrab) example which involves some reasonably complex networking.

The FreeBSD [source code](https://github.com/freebsd/freebsd-src/blob/b8afdda360e5915be3c2cf0d1438f511779b03db/sbin/ipf/ipf/bpf-ipf.h#L159) defines `SIZEOF_BPF_HDR` as 18 bytes, however during testing on a FreeBSD 14 system it was found that the size is actually 24 bytes, so that is what is used here in `smoltcp` when targeting FreeBSD. I'm open to corrections as I don't like just going of empirical evidence, but 18 bytes definitely did not work for the code I was testing with.

The other constants were checked against the FreeBSD source code and are correct, so those were just a matter of adding the `target_os` `cfg` parameters.

### DIFF
--- a/src/phy/sys/bpf.rs
+++ b/src/phy/sys/bpf.rs
@@ -13,7 +13,8 @@ use crate::wire::ETHERNET_HEADER_LEN;
     target_os = "macos",
     target_os = "ios",
     target_os = "netbsd",
-    target_os = "openbsd"
+    target_os = "openbsd",
+    target_os = "freebsd"
 ))]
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

jamwaffles:
Rapid! Thank you!

PR #907, opened 2024-02-18
feat(congestion control): add CongestionController trait and example impl

ytakano:
Add `CongestionController` trait for congestion controllers of TCP.

By default, this PR uses `NoControl`, which do not control congestion,
and the PR does not affect conventional behavior.

To use `Cubic` or `Reno` congestion controllers,
`socket-tcp-cubic` or `socket-tcp-reno` features must be set or
`tcp::Socket::new_with_cc()` must be used.

Users can implement custom congestion controllers
by using `CongestionController ` trait and `tcp::Socket::new_with_cc()`.

`Cubic` or `Reno` can be tested as follows.

```
$ cargo test --features socket-tcp-cubic
$ cargo test --features socket-tcp-reno
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,21 @@ defmt = ["dep:defmt", "heapless/defmt-03"]
 "socket-dns" = ["socket", "proto-dns"]
 "socket-mdns" = ["socket-dns"]
 
+# Enable Cubic TCP congestion control algorithm, and it is used as a default congestion controller.
+#
+# Cubic relies on double precision (`f64`) floating point operations, which may cause issues in some contexts:
+#
+# * Small embedded processors (such as Cortex-M0, Cortex-M1, and Cortex-M3) do not have an FPU,
+#   and floating point operations consume significant amounts of CPU time and Flash space.
+# * Interrupt handlers should almost always avoid floating-point operations.
+# * Kernel-mode code on desktop processors usually avoids FPU operations to reduce the penalty of saving and restoring FPU registers.
+#
+# In all these cases, `CongestionControl::Reno` is a better choice of congestion control algorithm.
+"socket-tcp-cubic" = []
+
+# Enable Reno TCP congestion control algorithm, and it is used as a default congestion controller.
+"socket-tcp-reno" = []
+
 "packetmeta-id" = []
 
 "async" = []
@@ -85,7 +100,7 @@ default = [
 ]
 
 # Private features
-# Features starting with "_" are considered private. They should not be enabled by 
+# Features starting with "_" are considered private. They should not be enabled by
 # other crates, and they are not considered semver-stable.
 
 "_proto-fragmentation" = []
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -17,6 +17,8 @@ use crate::wire::{
     TCP_HEADER_LEN,
 };
 
+mod congestion;
+
 macro_rules! tcp_trace {
     ($($arg:expr),*) => (net_log!(trace, $($arg),*));
 }
@@ -390,6 +392,19 @@ impl Display for Tuple {
     }
 }
 
+/// A congestion control algorithm.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum CongestionControl {
+    None,
+
+    #[cfg(feature = "socket-tcp-reno")]
+    Reno,
+
+    #[cfg(feature = "socket-tcp-cubic")]
+    Cubic,
+}
+
 /// A Transmission Control Protocol socket.
 ///
 /// A TCP socket may passively listen for connections or actively connect to another endpoint.
@@ -464,6 +479,9 @@ pub struct Socket<'a> {
     /// Nagle's Algorithm enabled.
     nagle: bool,
 
+    /// The congestion control algorithm.
+    congestion_controller: congestion::AnyController,
+
     #[cfg(feature = "async")]
     rx_waker: WakerRegistration,
     #[cfg(feature = "async")]
@@ -522,11 +540,60 @@ impl<'a> Socket<'a> {
             ack_delay_timer: AckDelayTimer::Idle,
             challenge_ack_timer: Instant::from_secs(0),
             nagle: true,
+            congestion_controller: congestion::AnyController::new(),
 
             #[cfg(feature = "async")]
             rx_waker: WakerRegistration::new(),
             #[cfg(feature = "async")]
             tx_waker: WakerRegistration::new(),
+        }
+    }
+
+    /// Set an algorithm for congestion control.
+    ///
+    /// `CongestionControl::None` indicates that no congestion control is applied.
+    /// Options `CongestionControl::Cubic` and `CongestionControl::Reno` are also available.
+    /// To use Reno and Cubic, please enable the `socket-tcp-reno` and `socket-tcp-cubic` features
+    /// in the `smoltcp` crate, respectively.
+    ///
+    /// `CongestionControl::Reno` is a classic congestion control algorithm valued for its simplicity.
+    /// Despite having a lower algorithmic complexity than `Cubic`,
+    /// it is less efficient in terms of bandwidth usage.
+    ///
+    /// `CongestionControl::Cubic` represents a modern congestion control algorithm designed to
+    /// be more efficient and fair compared to `CongestionControl::Reno`.
+    /// It is the default choice for Linux, Windows, and macOS.
+    /// `CongestionControl::Cubic` relies on double precision (`f64`) floating point operations, which may cause issues in some contexts:
+    /// * Small embedded processors (such as Cortex-M0, Cortex-M1, and Cortex-M3) do not have an FPU, and floating point operations consume significant amounts of CPU time and Flash space.
+    /// * Interrupt handlers should almost always avoid floating-point operations.
+    /// * Kernel-mode code on desktop processors usually avoids FPU operations to reduce the penalty of saving and restoring FPU registers.
+    /// In all these cases, `CongestionControl::Reno` is a better choice of congestion control algorithm.
+    pub fn set_congestion_control(&mut self, congestion_control: CongestionControl) {
+        use congestion::*;
+
+        self.congestion_controller = match congestion_control {
+            CongestionControl::None => AnyController::None(no_control::NoControl),
+
+            #[cfg(feature = "socket-tcp-reno")]
+            CongestionControl::Reno => AnyController::Reno(reno::Reno::new()),
+
+            #[cfg(feature = "socket-tcp-cubic")]
+            CongestionControl::Cubic => AnyController::Cubic(cubic::Cubic::new()),
+        }
+    }
+
+    /// Return the current congestion control algorithm.
+    pub fn congestion_control(&self) -> CongestionControl {
+        use congestion::*;
+
+        match self.congestion_controller {
+            AnyController::None(_) => CongestionControl::None,
+
+            #[cfg(feature = "socket-tcp-reno")]
+            AnyController::Reno(_) => CongestionControl::Reno,
+
+            #[cfg(feature = "socket-tcp-cubic")]
+            AnyController::Cubic(_) => CongestionControl::Cubic,
         }
     }
 
@@ -1593,6 +1660,9 @@ impl<'a> Socket<'a> {
                 }
 
                 self.rtte.on_ack(cx.now(), ack_number);
+                self.congestion_controller
+                    .inner_mut()
+                    .on_ack(cx.now(), ack_len, &self.rtte);
             }
         }
 
@@ -1636,6 +1706,9 @@ impl<'a> Socket<'a> {
                         tcp_trace!("received SYNACK with zero MSS, ignoring");
                         return None;
                     }
+                    self.congestion_controller
+                        .inner_mut()
+                        .set_mss(max_seg_size as usize);
                     self.remote_mss = max_seg_size as usize
                 }
 
@@ -1681,6 +1754,9 @@ impl<'a> Socket<'a> {
                         return None;
                     }
                     self.remote_mss = max_seg_size as usize;
+                    self.congestion_controller
+                        .inner_mut()
+                        .set_mss(self.remote_mss);
                 }
 
                 self.remote_seq_no = repr.seq_number + 1;
@@ -1795,6 +1871,10 @@ impl<'a> Socket<'a> {
         let is_window_update = new_remote_win_len != self.remote_win_len;
         self.remote_win_len = new_remote_win_len;
 
+        self.congestion_controller
+            .inner_mut()
+            .set_remote_window(new_remote_win_len);
+
         if ack_len > 0 {
             // Dequeue acknowledged octets.
             debug_assert!(self.tx_buffer.len() >= ack_len);
@@ -1830,6 +1910,11 @@ impl<'a> Socket<'a> {
                 {
                     // Increment duplicate ACK count
                     self.local_rx_dup_acks = self.local_rx_dup_acks.saturating_add(1);
+
+                    // Inform congestion controller of duplicate ACK
+                    self.congestion_controller
+                        .inner_mut()
+                        .on_duplicate_ack(cx.now());
 
                     net_debug!(
                         "received duplicate ACK for seq {} (duplicate nr {}{})",
@@ -1995,6 +2080,9 @@ impl<'a> Socket<'a> {
             0
         };
 
+        // Compare max_send with the congestion window.
+        let max_send = max_send.min(self.congestion_controller.inner().window());
+
         // Can we send at least 1 octet?
         let mut can_send = max_send != 0;
         // Can we send at least 1 full segment?
@@ -2072,6 +2160,10 @@ impl<'a> Socket<'a> {
             self.remote_last_ts = Some(cx.now());
         }
 
+        self.congestion_controller
+            .inner_mut()
+            .pre_transmit(cx.now());
+
         // Check if any state needs to be changed because of a timer.
         if self.timed_out(cx.now()) {
             // If a timeout expires, we should abort the connection.
@@ -2095,6 +2187,11 @@ impl<'a> Socket<'a> {
 
                 // Inform RTTE, so that it can avoid bogus measurements.
                 self.rtte.on_retransmit();
+
+                // Inform the congestion controller that we're retransmitting.
+                self.congestion_controller
+                    .inner_mut()
+                    .on_retransmit(cx.now());
             }
         }
 
@@ -2315,6 +2412,9 @@ impl<'a> Socket<'a> {
         if repr.segment_len() > 0 {
             self.rtte
                 .on_send(cx.now(), repr.seq_number + repr.segment_len());
+            self.congestion_controller
+                .inner_mut()
+                .post_transmit(cx.now(), repr.segment_len());
         }
 
         if !self.seq_to_transmit(cx) && repr.segment_len() > 0 {
@@ -7308,5 +7408,25 @@ mod test {
             r.sample(100);
             assert_eq!(r.retransmission_timeout(), Duration::from_millis(rto));
         }
+    }
+
+    #[test]
+    fn test_set_get_congestion_control() {
+        let mut s = socket_established();
+
+        #[cfg(feature = "socket-tcp-reno")]
+        {
+            s.set_congestion_control(CongestionControl::Reno);
+            assert_eq!(s.congestion_control(), CongestionControl::Reno);
+        }
+
+        #[cfg(feature = "socket-tcp-cubic")]
+        {
+            s.set_congestion_control(CongestionControl::Cubic);
+            assert_eq!(s.congestion_control(), CongestionControl::Cubic);
+        }
+
+        s.set_congestion_control(CongestionControl::None);
+        assert_eq!(s.congestion_control(), CongestionControl::None);
     }
 }
--- a/src/socket/tcp/congestion.rs
+++ b/src/socket/tcp/congestion.rs
@@ -1,0 +1,101 @@
+use crate::time::Instant;
+
+use super::RttEstimator;
+
+pub(super) mod no_control;
+
+#[cfg(feature = "socket-tcp-cubic")]
+pub(super) mod cubic;
+
+#[cfg(feature = "socket-tcp-reno")]
+pub(super) mod reno;
+
+#[allow(unused_variables)]
+pub(super) trait Controller {
+    /// Returns the number of bytes that can be sent.
+    fn window(&self) -> usize;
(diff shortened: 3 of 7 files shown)


ytakano:
Avoid `cbrt()` and `powi()` for no_std.

pothos:
Just for your reference, there was also https://github.com/smoltcp-rs/smoltcp/pull/450 and https://github.com/cbranch/smoltcp/commits/cbranch/cc (other non-upstreamed things are also public: https://github.com/cbranch/smoltcp/branches/stale, and if you are looking into BBR there is https://github.com/quinn-rs/quinn/tree/0.8.0)

ytakano:
When I was testing smoltcp, it occupied traffic bandwidth, and other TCP connections starved due to the lack of congestion control.

By the way, I think delay based congestion control mechanisms like BBR should be also implemented by using `CongestionController`,
because it can know time and bytes of sending/receiving data
and RTT by using following methods.

```rust
pub trait CongestionController: Default {
    #[allow(unused_variables)]
    fn on_ack(&mut self, now: Instant, len: usize, rtt: &RttEstimator) {}

    #[allow(unused_variables)]
    fn post_transmit(&mut self, now: Instant, len: usize) {}
}
```

This is similar to quinn.
https://github.com/quinn-rs/quinn/blob/0.8.0/quinn-proto/src/congestion.rs

I know theory of BBR, but I cannot evaluate it sufficiently because it requires emulation of delays.

src/socket/tcp.rs
```diff
@@ -390,14 +402,23 @@ impl Display for Tuple {
...
+
 /// A Transmission Control Protocol socket.
 ///
 /// A TCP socket may passively listen for connections or actively connect to another endpoint.
 /// Note that, for listening sockets, there is no "backlog"; to be able to simultaneously
 /// accept several connections, as many sockets must be allocated, or any new connection
 /// attempts will be reset.
 #[derive(Debug)]
-pub struct Socket<'a> {
+pub struct Socket<'a, CC: CongestionController = DefaultCC> {
```
Dirbaio: there's no use in having this as a generic param, the user can't set it to something other than default due to how `SocketSet` works. 

I'd instead use the type directly on the field: `congestion_controller: DefaultCC`. And maybe rename it to something like "active" or "chosen": `ActiveCC` or `ActiveCongestionController.`

src/socket/tcp.rs
```diff
@@ -152,7 +156,7 @@ const RTTE_MAX_RTO: u32 = 10000;
 
 #[derive(Debug, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-struct RttEstimator {
+pub struct RttEstimator {
```
Dirbaio: why does this have to be pub?

src/socket/tcp.rs
```diff
@@ -17,6 +17,10 @@ use crate::wire::{
     TCP_HEADER_LEN,
 };
 
+use self::congestion_controller::CongestionController;
+
+pub mod congestion_controller;
```
Dirbaio: why does this have to be pub?

src/socket/tcp.rs
ytakano: All right.
I will fix to remove the generic parameter, and place it to the field.

Thank you for your review.

src/socket/tcp.rs
ytakano: Because this implementation let users to pass a congestion controller via a type parameter,
`RttEstimator` must be exposed.
However, I will fix to remove the type parameter as discussed above,
and it will be non-public.

src/socket/tcp.rs
ytakano: Same as above.
I will fix this.

src/socket/tcp.rs
ytakano: Prepare `ActiveCC` as follows.
https://github.com/ytakano/smoltcp/blob/179fd7e5c9f552eabd7da3e955d132d2cf4d8b97/src/socket/tcp/congestion_controller.rs#L14-L21

The active congestion controller can be replaced by `socket-tcp-cubic` or `socket-tcp-reno` features.

In addition to that, I removed the type parameter of `CC: CongestionController = DefaultCC` from `tcp::Socket`.

src/socket/tcp.rs
ytakano: fixed

src/socket/tcp.rs
ytakano: fixed

whitequark (changes requested):
I think we shouldn't add this with non-additive features. I think it's OK to have an enum of all possible congestion control methods, with some sensible default selection, but still allowing two crates that both use `smoltcp` to compile without needing to coordinate their feature sets.

ytakano:
> I think we shouldn't add this with non-additive features. I think it's OK to have an enum of all possible congestion control methods, with some sensible default selection, but still allowing two crates that both use `smoltcp` to compile without needing to coordinate their feature sets.

Right.
A general enum type of congestion control can be prepared.
I implemented it before, but discarded.
I will fix it later.

Thank you for your review.

whitequark:
To expand a bit, I'm thinking that the enum could include all of the necessary state, therefore making `tcp::Socket` as much bigger as the biggest possible to choose congestion controller (0 bytes if none are enabled). 

The way to create it would probably be methods on the `CongestionController` enum, whose fields are all private. Then a `Default` implementation would pick the "best" one available if several are enabled. (The decision on which congestion controller is the "best" one is probably contentious but I don't expect it to matter that much since most people will pick one.)

Once this PR is in, we should strongly recommend picking a congestion controller in the README, to be a good netizen.

ytakano:
I have defined `set_congestion_control()` and `get_congestion_control()` methods for `tcp::Socket` as follows.

```rust
pub enum CongestionControl {
    None,
    Cubic,
    Reno,
}

impl tcp::Scoket {
    /// Set an algorithm for congestion control.
    ///
    /// The default setting is `CongestionControl::None`, indicating that no congestion control is applied.
    /// Options `CongestionControl::Cubic` and `CongestionControl::Reno` are also available.
    ///
    /// `CongestionControl::Cubic` represents a modern congestion control algorithm designed to
    /// be more efficient and fair compared to `CongestionControl::Reno`.
    /// It is the default choice for Linux, Windows, and macOS.
    ///
    /// `CongestionControl::Reno` is a classic congestion control algorithm valued for its simplicity.
    /// Despite having a lower algorithmic complexity than `Cubic`,
    /// it is less efficient in terms of bandwidth usage.
    pub fn set_congestion_control(&mut self, congestion_control: CongestionControl) { /* omitted */ }

    /// Return the current congestion control algorithm.
    pub fn get_congestion_control(&self) -> CongestionControl { /* omitted */ }
}
```

> CongestionController enum, whose fields are all private.

Variants of a public enum cannot be private.
So, I have defined these 2 methods and `CongestionControl` enum to `tcp::Socket`.

If the best performance is required,  I think it should be a type parameter of `tcp::Socket`.

```Rust
pub struct Socket<'a, CC: CongestionController> {
    congestion_controller: CC
}
```

However, it is less compatible.
See https://github.com/smoltcp-rs/smoltcp/pull/907#discussion_r1493859449

If it is a type parameter, the required bytes of the empty congestion controller must be 0 as follows.

```Rust
let socket = tcp::Socket<EmptyCongestionController>();
```

whitequark:
Yeah it's not actually possible to use a type parameter here. A enum is perfectly fine.

src/socket/tcp.rs
```diff
@@ -17,6 +17,10 @@ use crate::wire::{
     TCP_HEADER_LEN,
 };
 
+use self::congestion_controller::CongestionController;
```
whitequark: Nit: I find `congestion_controller` very long for a module name. `mod congestion`?

src/socket/tcp.rs
```diff
@@ -474,7 +489,7 @@ const DEFAULT_MSS: usize = 536;
 
 impl<'a> Socket<'a> {
     #[allow(unused_comparisons)] // small usize platforms always pass rx_capacity check
-    /// Create a socket using the given buffers.
+    /// Create a socket using the given buffers
```
whitequark: This seems to have been changed by mistake.

src/socket/tcp.rs
```diff
@@ -530,6 +546,39 @@ impl<'a> Socket<'a> {
...
+
+        self.congestion_controller = match congestion_control {
+            CongestionControl::None => GenericController::None,
+            CongestionControl::Cubic => GenericController::Cubic(cubic::Cubic::default()),
+            CongestionControl::Reno => GenericController::Reno(reno::Reno::default()),
+        }
+    }
+
+    /// Return the current congestion control algorithm.
+    pub fn get_congestion_control(&self) -> CongestionControl {
```
whitequark: This should be called `fn congestion_control` when following our naming scheme for accessors.

src/socket/tcp/congestion_controller.rs
```diff
@@ -0,0 +1,90 @@
...
+
+pub(super) mod cubic;
+pub(super) mod reno;
+
+pub(super) trait CongestionController: Default {
+    /// Returns the number of bytes that can be sent.
+    fn window(&self) -> usize;
+
+    /// Set the remote window size.
+    #[allow(unused_variables)]
```
whitequark: These can just go on the trait.

src/socket/tcp.rs
ytakano: Need period after `buffers`.

src/socket/tcp/congestion_controller.rs
```diff
@@ -0,0 +1,90 @@
...
+    Reno(reno::Reno),
+}
+
+impl Default for GenericController {
+    fn default() -> Self {
+        GenericController::None
+    }
+}
+
+impl CongestionController for GenericController {
```
whitequark: Can this be implemented as something like the below rather than with this indirect impl?

```
impl GenericController {
    fn inner_mut(&mut self) -> &mut dyn CongestionController {
    }
}
```

src/socket/tcp/congestion_controller/cubic.rs
```diff
@@ -0,0 +1,312 @@
+use crate::time::Instant;
+
+use super::CongestionController;
+
+// Constants for the Cubic congestion control algorithm.
+// See RFC 8312.
+const BETA_CUBIC: f64 = 0.7;
+const C: f64 = 0.4;
```
whitequark: These `f64`s will be incredibly inefficient on most small embedded systems, being essentially unusable. (This isn't just execution time overhead, though it's that; code space overhead will impact many applications too.)

(By the way, it makes no sense to propose adding a generic parameter to reduce on call dispatch overhead while leaving in the massive f64 softfloat overhead. The latter will dwarf just about anything else...)

whitequark: A fixed point implementation would be ideal. Without it, the usefulness of the cubic congestion controller is quite limited, enough that it should be called out in the documentation as "do not use this on your STM32" or something like that (not the exact text).

whitequark (changes requested):

src/socket/tcp.rs
ytakano: OK. I will fix.

src/socket/tcp.rs
ytakano: I will fix. Thank you.

src/socket/tcp/congestion_controller.rs
ytakano: I tried it, but Rust compiler rejected it.

whitequark: I tried it just now in the playground and it seems to work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ee36a332e85a0b3d92f1a437f1b949da

src/socket/tcp/congestion_controller/cubic.rs
ytakano: It makes sense.

I want `f64` implementation for my operating system currently I implementing running on x86_64 or AArch64.
So, I will add more documents about `Cubic` and `Reno`.

src/socket/tcp/congestion_controller.rs
whitequark: If you do it like this, I think that in the case where the enum has only one element, the call might get devirtualized, and you'll get the benefits of the generic parameter without actually using one.

ytakano: Because `Default` trait requires `Self: Sized`, `inner_mut()` has been rejected.
I will try to avoid `Default`.

src/socket/tcp.rs
ytakano: features are removed.
See https://github.com/smoltcp-rs/smoltcp/pull/907#issuecomment-1956049830

whitequark: I think keeping futures makes sense to reduce code size on embedded devices.

whitequark: This is especially important if you're using `f64` since those will cause a large amount of code to be linked in for softfloat.

src/socket/tcp.rs
ytakano: fixed

src/socket/tcp.rs
ytakano: fixed

src/socket/tcp.rs
ytakano: fixed

src/socket/tcp/congestion_controller.rs
ytakano: fixed

src/socket/tcp/congestion_controller.rs
ytakano: Removing `Default` from `CongestionController`'s types, and `inner_mut()` works fine.
Thank you!

src/socket/tcp.rs
ytakano: fixed

src/socket/tcp/congestion_controller/cubic.rs
ytakano: I have inserted a warning message as follows to the document of `set_congestion_control()`.

```Rust
    /// `CongestionControl::Cubic` uses `f64` internally, which is less efficient on some platforms like STM32.
    pub fn set_congestion_control(&mut self, congestion_control: CongestionControl) {}
```

src/socket/tcp.rs
ytakano: OK.
I will add features again for control congestion control.

ytakano: I have added `socket-tcp-cubic` feature for Cubic.
Cubic is disabled by default.

I think Reno is not bad for small devices like STM32, and uses can use Cubic by specifying the feature.

Fixed-point numbers crate may help the improvement, but is should be another PR I think.
https://docs.rs/fixed/latest/fixed/

src/socket/tcp/congestion_controller/cubic.rs
ytakano: I have added `socket-tcp-cubic` feature as discussed https://github.com/smoltcp-rs/smoltcp/pull/907#discussion_r1497091986 .
Cubic is disabled by default, and users cannot use it unless `socket-tcp-cubic` feature is enabled.
So, unintentional behavior can be avoided.

src/socket/tcp.rs
ytakano: Inserted a message of Cubic as follows.
```rust
/// To use cubic, please enable the `socket-tcp-cubic` feature in the `smoltcp` crate.
pub fn set_congestion_control(&mut self, congestion_control: CongestionControl) {}
```

Cargo.toml
```diff
@@ -70,6 +70,8 @@ defmt = ["dep:defmt", "heapless/defmt-03"]
 "socket-dns" = ["socket", "proto-dns"]
 "socket-mdns" = ["socket-dns"]
 
+"socket-tcp-cubic" = [] # Enable Cubic TCP congestion control algorithm
```
whitequark: I'd like all congestion control algorithms to be pluggable since this is what we also use for all other features.

src/socket/tcp/congestion.rs
```diff
@@ -0,0 +1,93 @@
...
+            GenericController::None => None,
+            GenericController::Reno(r) => Some(r),
+
+            #[cfg(feature = "cubic")]
+            GenericController::Cubic(c) => Some(c),
+        }
+    }
+}
+
+impl CongestionController for GenericController {
```
whitequark: So I was actually thinking of something else: where the `None` controller option would belong to just another implementation of a congestion controller. The enum itself would not implement `CongestionController` at all, and the `&mut dyn CongestionController` would be used by `tcp::Socket` itself.

src/socket/tcp.rs
```diff
@@ -530,6 +548,45 @@ impl<'a> Socket<'a> {
...
+    /// To use cubic, please enable the `socket-tcp-cubic` feature in the `smoltcp` crate.
+    ///
+    /// `CongestionControl::Reno` is a classic congestion control algorithm valued for its simplicity.
+    /// Despite having a lower algorithmic complexity than `Cubic`,
+    /// it is less efficient in terms of bandwidth usage.
+    ///
+    /// `CongestionControl::Cubic` represents a modern congestion control algorithm designed to
+    /// be more efficient and fair compared to `CongestionControl::Reno`.
+    /// It is the default choice for Linux, Windows, and macOS.
+    /// `CongestionControl::Cubic` uses `f64` internally, which is less efficient on some platforms like STM32.
```
whitequark:
```suggestion
    /// `CongestionControl::Cubic` relies on double precision (`f64`) floating point operations, which may cause issues in some contexts:
    /// * Small embedded processors (such as Cortex-M0, Cortex-M1, and Cortex-M3) do not have an FPU, and floating point operations consume significant amounts of CPU time and Flash space.
    /// * Interrupt handlers should almost always avoid floating-point operations.
    /// * Kernel-mode code on desktop processors usually avoids FPU operations to reduce the penalty of saving and restoring FPU registers.
    /// In all these cases, `CongestionControl::Reno` is a better choice of congestion control algorithm.
```

whitequark (changes requested):

Cargo.toml
ytakano: `socket-tcp-reno` feature has been added.

src/socket/tcp/congestion.rs
ytakano: Use `GenericController`'s `inner()` and `inner_mut()` methods to call `CongestionController`'s methods.

src/socket/tcp/congestion.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    #[cfg(feature = "socket-tcp-cubic")]
+    Cubic(cubic::Cubic),
+}
+
+impl GenericController {
+    #[allow(unreachable_code)]
+    #[inline]
+    pub fn new() -> Self {
+        #[cfg(feature = "socket-tcp-cubic")]
```
whitequark, line 63: It is worth adding a comment here explaining that:
* We are selecting a "best" congestion controller intentionally (i.e. there is an ordering, we are not just picking any available).
* There is a reason why we are picking one over another.

src/socket/tcp/congestion.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    fn post_transmit(&mut self, now: Instant, len: usize) {}
+
+    /// Set the maximum segment size.
+    fn set_mss(&mut self, mss: usize) {}
+}
+
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub(super) enum GenericController {
```
whitequark: I feel that `GenericController` does not describe this enum well (even if it's internal). What about `AnyController`?

I'm open to other options, it is just that "Generic" in a programming language context usually means either "non-specific" (which isn't the case as the enum always specifies one particular controller) or "generic over type" (which isn't the case either).

src/socket/tcp/congestion.rs
```diff
@@ -0,0 +1,89 @@
...
+pub(super) mod no_control;
+
+#[cfg(feature = "socket-tcp-cubic")]
+pub(super) mod cubic;
+
+#[cfg(feature = "socket-tcp-reno")]
+pub(super) mod reno;
+
+#[allow(unused_variables)]
+pub(super) trait CongestionController {
```
whitequark: I think we can rename this just to `Controller`, because it will usually be named `congestion::Controller`, and so everything will be a little shorter.

Also, then it nicely pairs with the proposal for `AnyController` below; a `congestion::AnyController` enum can be lowered to a specific (but type-erased) `congestion::Controller` trait object.

whitequark (commented):
Thank you, I am happy with the functionality now, and I only have a few comments about naming.

src/socket/tcp/congestion.rs
ytakano: Thank you for your review.

Added a document as follows.

https://github.com/ytakano/smoltcp/blob/991db961e3c6ea08e4eb13b544525b1ef4ff793b/src/socket/tcp/congestion.rs#L48-L59

```rust
impl AnyController {
    /// Create a new congestion controller.
    /// `AnyController::new()` selects the best congestion controller based on the features.
    ///
    /// - If `socket-tcp-cubic` feature is enabled, it will use `Cubic`.
    /// - If `socket-tcp-reno` feature is enabled, it will use `Reno`.
    /// - If both `socket-tcp-cubic` and `socket-tcp-reno` features are enabled, it will use `Cubic`.
    ///    - `Cubic` is more efficient regarding throughput.
    ///    - `Reno` is more conservative and is suitable for low-power devices.
    /// - If no congestion controller is available, it will use `NoControl`.
    ///
    /// Users can also select a congestion controller manually by [`super::Socket::set_congestion_control()`]
    /// method at run-time.
    #[allow(unreachable_code)]
    #[inline]
    pub fn new() -> Self {
```

src/socket/tcp/congestion.rs
ytakano: Renamed.

src/socket/tcp/congestion.rs
ytakano: Renamed.

ytakano:
Now we can test `Cubic` and `Reno` by specifying the features as follows.

```
$ cargo test --features=socket-tcp-reno
$ cargo test --features=socket-tcp-cubic
```

whitequark (approved):
LGTM. Thank you for this excellent work!

@Dirbaio Once this is in, I think we ought to consider recommending enabling a congestion controller in the docs, to be a good netizen.

whitequark:
Can you squash everything into one commit please?

ytakano:
Squashed.

whitequark:
I think this includes some unrelated commits?

ytakano:
Sorry. How about this?

whitequark:
Thanks!

whitequark:
@Dirbaio Once this is merged I think we should strongly recommend that Reno be selected unless there is an extremely good reason not to use any congestion control.

PR #914, opened 2024-04-02
Implement `join/leave_multicast_group()` for IPv6

lucasvr:
This patch rebases @jgallagher's PR #602 to the tip of the main branch, adding support for IPv6 multicast groups.

As in the original PR, it is possible to join groups by sending an initial MLDv2 Report packet. In addition to PR #602, this PR also implements `leave_multicast_group()`.

It is not yet possible to resend the change report or respond to queries.


lucasvr:
Folks, I would appreciate if someone could review this PR. For symmetry, in addition to @jgallagher's original pull request, this PR implements `leave_multicast_group()`. This pair of functions is enough to commission and control devices that use the [Matter](https://github.com/project-chip/rs-matter) protocol.

src/iface/interface/igmp.rs
```diff
@@ -10,6 +10,8 @@ pub enum MulticastError {
     GroupTableFull,
     /// IPv6 multicast is not yet supported.
     Ipv6NotSupported,
```
thvdveld: I think you can remove the `Ipv6NotSupported` now?

thvdveld (changes requested):
I think everything looks good. Just a small thing that I noticed.

src/iface/interface/igmp.rs
lucasvr: Done. Thanks for looking into this!

src/wire/ipv6hbh.rs
```diff
@@ -105,6 +106,18 @@ impl<'a> Repr<'a> {
             buffer = &mut buffer[opt.buffer_len()..];
         }
     }
+
+    /// The hop-by-hop header containing a PadN and a MLDv2 router alert option.
+    pub fn mldv2_router_alert(n: u8) -> Self {
```
thvdveld: Is there a case where the padding is not equal to PadN(0)?

Cargo.toml
```diff
@@ -296,6 +296,10 @@ required-features = ["log", "medium-ethernet", "proto-ipv4", "socket-tcp"]
 name = "multicast"
 required-features = ["std", "medium-ethernet", "medium-ip", "phy-tuntap_interface", "proto-ipv4", "proto-igmp", "socket-udp"]
 
+[[example]]
+name = "multicast6"
+required-features = ["std", "medium-ethernet", "medium-ip", "phy-tuntap_interface", "proto-ipv6", "socket-udp"]
```
thvdveld, line 301: Shouldn't `proto-igmp` be required as well?

src/iface/interface/ipv6.rs
```diff
@@ -472,4 +473,67 @@ impl InterfaceInner {
...
+            IpCidr::Ipv6(cidr) => {
+                let addr = cidr.address();
+                if addr.is_link_local() {
+                    Some(addr)
+                } else {
+                    None
+                }
+            }
+        })
+    }
```
thvdveld: Could you move this function under the `fn ipv6_addr` function?

src/iface/interface/tests/ipv6.rs
```diff
@@ -1169,3 +1169,126 @@ fn get_source_address_empty_interface() {
...
+            .insert(
+                0,
+                Ipv6OptionRepr::Unknown {
+                    type_: Ipv6OptionType::Unknown(IpProtocol::Icmpv6.into()),
+                    length: 0,
+                    data: &[],
+                },
+            )
+            .unwrap();
+        assert_eq!(hbh_repr, expected_hbh_repr);
```
thvdveld: I'm not sure why this is needed? Are we unable to parse the Hop-by-Hop option?

src/wire/mld.rs
```diff
@@ -294,6 +294,63 @@ impl<T: AsRef<[u8]> + AsMut<[u8]>> AddressRecord<T> {
...
+    pub fn emit<T: AsRef<[u8]> + AsMut<[u8]>>(&self, record: &mut AddressRecord<T>) {
+        record.set_record_type(self.record_type);
+        record.set_aux_data_len(self.aux_data_len);
+        record.set_num_srcs(self.num_srcs);
+        record.set_mcast_addr(self.mcast_addr);
+        let record_payload = record.payload_mut();
+        if self.payload.len() == record_payload.len() {
+            // TODO: handle the case where the payload sizes are different
+            record_payload.copy_from_slice(self.payload);
+        }
```
thvdveld: Should we maybe set the payload outside of this function, since the `buffer_len` function returns the buffer length excluding the payload length?

thvdveld (changes requested):

Cargo.toml
lucasvr: Since `proto-igmp` belongs to the `default` dependencies I didn't spot this before. Fixed.

src/wire/ipv6hbh.rs
lucasvr: If the only option being transmitted is `RouterAlert` then `PadN(0)` would suffice. But if one wishes to append more options after `RouterAlert` then a different padding value might be needed.

src/iface/interface/ipv6.rs
lucasvr: Done.

src/iface/interface/tests/ipv6.rs
lucasvr: This was mostly out of habit. I have updated the test to check the output of the parsed representation instead.

src/wire/mld.rs
lucasvr: Good idea -- I have now moved the payload copy to `wire::Repr::emit()`, which fits nicely with the way that `Repr::Query` and `Repr::Report` payloads are handled.

src/wire/ipv6hbh.rs
thvdveld: Shouldn't the padding be calculated at the end, when emitting the options?

src/wire/mld.rs
```diff
@@ -386,6 +440,21 @@ impl<'a> Repr<'a> {
...
+                packet.set_nr_mcast_addr_rcrds(records.len() as u16);
+                let mut payload = packet.payload_mut();
+                for record in *records {
+                    if record.payload.len() == payload.len() {
+                        // TODO: handle the case where the payload sizes are different
+                        payload.copy_from_slice(record.payload);
+                    }
+                    record.emit(&mut AddressRecord::new_unchecked(&mut *payload));
+                    payload = &mut payload[record.buffer_len()..];
+                }
```
thvdveld: Is this correct? I am not sure what you are trying to achieve with the if block, because I guess that the emit function of the record is going to override some parts of the payload variable.

src/wire/ipv6hbh.rs
lucasvr: To keep the API surface small, I opted to let `mldv2_router_alert()` handle both options. But I agree that going forward it may be good to have a dedicated function to emit the padding option. I have pushed a commit addressing that.

src/wire/mld.rs
lucasvr: This was actually part of @jgallagher's original PR and, indeed, it can go. The source payload is always empty in the current implementation and, even if it weren't, the copied data would be overwritten when emitting the address record. I have just fixed this. Thanks for the code review!

thvdveld:
I think we can merge this now. Thank you for your PR!

PR #917, opened 2024-04-08
fix(tcp): only reset remote_last_ts if some data is enqueued

copy:
I call `send` with a callback that often returns `0` (e.g. a real tcp socket that returns `WouldBlock`). This causes `remote_last_ts` being set to `None` and `process` being called even though nothing on the socket has changed.


Dirbaio (approved):
thanks!

PR #919, opened 2024-04-10
ignore checksum during loopback

skpupil:
There is no need to check the checksum during loopback.
According to tests, this optimization improves performance by more than 30%.

### DIFF
--- a/src/phy/loopback.rs
+++ b/src/phy/loopback.rs
@@ -1,7 +1,7 @@
 use alloc::collections::VecDeque;
 use alloc::vec::Vec;
 
-use crate::phy::{self, Device, DeviceCapabilities, Medium};
+use crate::phy::{self, ChecksumCapabilities, Device, DeviceCapabilities, Medium};
 use crate::time::Instant;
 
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):

whitequark:
Do we... care about loopback performance?

skpupil:
> Do we... care about loopback performance?

We don't check the checksum during loopback according to the Linux network stack. The performance improvement then becomes just a fringe benefit. BTW, if we aim to identify the performance bottleneck of smoltcip, it's easier to do so after excluding checksum checking.

whitequark:
That is true in some cases, though I would expect most performance bottlenecks to come from the relatively (compared to e.g. Linux) immature TCP implementation, not the basic and straightforward bit manipulation code.

skpupil:
> That is true in some cases, though I would expect most performance bottlenecks to come from the relatively (compared to e.g. Linux) immature TCP implementation, not the basic and straightforward bit manipulation code.

emmm, perhaps both are needed. I'll try my best to do more. Thx.

Issue #443, opened 2021-03-19
Why does RxToken require a &mut [u8] as the argument?


JakkuSakura:
I only got &[u8] somewhere, and find that RxToken requires a &mut [u8]. I can't see any use case of &mut u8. Is it for future-proof or something else?

whitequark:
It was added in case the internals of smoltcp might want to modify the incoming packet, which is usually exclusive to the TCP/IP stack due to how networking hardware is accessed. We can probably downgrade it to `&[u8]` to support more use cases since I don't think anything in smoltcp actually used that option.

JakkuSakura:
I transmuted the reference to mut, and it works just fine in this case.

Sympatron:
If it's not used mutably by smoltcp it will work currently. It's just dangerous because it may break, when smoltcp decides to mutate it some day. So it would probably be better to change it to `&[u8]` if smoltcp ended up not needing it and probably won't use it in the future.
Current implementations would still work fine with this change without any change.

I started working on a smoltcp device implementation recently too and think this would be easier to implement safely if it were not mutable.

whitequark:
> I transmuted the reference to mut, and it works just fine in this case.

I think this might be unsound.

JakkuSakura:
I'm fully aware of the unsoundness, but I have to get my job done anyway. If the code break, I can review all unsafe code then

PR #924, opened 2024-04-23
Change mutability of `RxToken`'s `consume` argument.

lmbollen:
This pull request drops the requirement that we can mutate the receive buffer.

I noticed issue #443  already discussed this, but was closed without making the change. Even though the suggestion was to change it.

### DIFF
--- a/src/phy/fuzz_injector.rs
+++ b/src/phy/fuzz_injector.rs
@@ -1,3 +1,6 @@
+extern crate alloc;
+use alloc::vec::Vec;
+
 use crate::phy::{self, Device, DeviceCapabilities};
 use crate::time::Instant;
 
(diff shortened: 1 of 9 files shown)


whitequark (approved):

lmbollen:
Had to add:
```rust
extern crate alloc;
use alloc::vec::Vec;
```
to `fuzz_injector`, which should not be a problem since it's for development and not production.
@whitequark Feel free to merge or ping me if you still agree with the changes.

src/phy/fuzz_injector.rs
```diff
@@ -1,6 +1,8 @@
 use crate::phy::{self, Device, DeviceCapabilities};
 use crate::time::Instant;
 
+extern crate alloc;
+use alloc::vec::Vec;
```
whitequark: Style: These should probably be added above `use crate::`, not below.

lmbollen:
@whitequark I think we're ready to merge :) But I have no rights.

PR #933, opened 2024-05-23
Enable dynamic TCP receive window resizing

### LINKED ISSUE #927: [Feature Request] Resize TCP socket buffer dynamically after TCP socket has been created
Currently, the buffer for TCP socket needs to be preallocated in `socket::tcp::Socket::new()` and I found nowhere to change its size. However, a configurable window size is necessary to achieve a balance between good bandwidth and low memory footprint. I notice that TCP window scaling (RFC 1323) is advertised to be implemented but it seems it's only properly implemented for the remote endpoint. Can this feature be implemented in local side?

XOR-op:
Closes #927.

This PR introduces the ability to resize TCP receive buffer after the connection is established, as well as helper getter functions. This feature will help implement "rwnd auto-tuning", for example [in Linux kernel](https://github.com/torvalds/linux/blob/e091caf99f3a5006c95baec24330bac6f7f17193/net/ipv4/tcp_input.c#L630), which can reduce memory footprint when establishing a lot of connections without sacrificing the maximum achievable throughput.

Current requirements include:
1. The new buffer must be larger than the length of remaining data in the current buffer.
2. The new buffer must be multiple of (1 << self.remote_win_shift), which is the window scaling negotiated in the handshake.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -2660,9 +2660,7 @@ impl<'a> Socket<'a> {
                 .unwrap_or(&PollAt::Ingress)
         }
     }
-}
 
-impl Socket<'static> {
     /// Replace the receive buffer with a new one.
     ///
     /// The requirements for the new buffer are:
@@ -2678,10 +2676,10 @@ impl Socket<'static> {
     ///
     /// See also the [new_with_window_scaling](struct.Socket.html#method.new_with_window_scaling) and
     /// [local_recv_win_scale](struct.Socket.html#method.local_recv_win_scale) methods.
-    pub fn replace_recv_buffer<T: Into<SocketBuffer<'static>>>(
+    pub fn replace_recv_buffer<T: Into<SocketBuffer<'a>>>(
         &mut self,
         new_buffer: T,
-    ) -> Result<SocketBuffer<'static>, SocketBuffer<'static>> {
+    ) -> Result<SocketBuffer<'a>, SocketBuffer<'a>> {
         let mut replaced_buf = new_buffer.into();
         /* Check if the new buffer is valid
          * Requirements:
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -63,6 +63,30 @@ impl Display for ConnectError {
 #[cfg(feature = "std")]
 impl std::error::Error for ConnectError {}
 
+/// Error returned by set_*
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ArgumentError {
+    InvalidArgs,
+    InvalidState,
+    InsufficientResource,
+}
+
+impl Display for crate::socket::tcp::ArgumentError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            crate::socket::tcp::ArgumentError::InvalidArgs => write!(f, "invalid arguments by RFC"),
+            crate::socket::tcp::ArgumentError::InvalidState => write!(f, "invalid state"),
+            crate::socket::tcp::ArgumentError::InsufficientResource => {
+                write!(f, "insufficient runtime resource")
+            }
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for crate::socket::tcp::ArgumentError {}
+
 /// Error returned by [`Socket::send`]
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
@@ -523,8 +547,6 @@ impl<'a> Socket<'a> {
             panic!("receiving buffer too large, cannot exceed 1 GiB")
         }
         let rx_cap_log2 = mem::size_of::<usize>() * 8 - rx_capacity.leading_zeros() as usize;
-        let remote_win_shift = rx_cap_log2.saturating_sub(16) as u8;
-        let (rx_buffer, tx_buffer) = (rx_buffer.into(), tx_buffer.into());
 
         Socket {
             state: State::Closed,
@@ -545,7 +567,7 @@ impl<'a> Socket<'a> {
             remote_last_ack: None,
             remote_last_win: 0,
             remote_win_len: 0,
-            remote_win_shift,
+            remote_win_shift: rx_cap_log2.saturating_sub(16) as u8,
             remote_win_scale: None,
             remote_has_sack: false,
             remote_mss: DEFAULT_MSS,
@@ -790,11 +812,16 @@ impl<'a> Socket<'a> {
     /// It may be reset to 0 during the handshake if remote side does not support window scaling.
     ///
     /// # Errors
-    /// Returns an error if the socket is not in the `Closed` or `Listen` state, or if the
-    /// receive buffer is smaller than (1<<scale) bytes.
-    pub fn set_local_recv_win_scale(&mut self, scale: u8) -> Result<(), ()> {
+    /// `Err(ArgumentError::InvalidArgs)` if the scale is greater than 14.
+    /// `Err(ArgumentError::InvalidState)` if the socket is not in the `Closed` or `Listen` state.
+    /// `Err(ArgumentError::InsufficientResource)` if the receive buffer is smaller than (1<<scale) bytes.
+    pub fn set_local_recv_win_scale(&mut self, scale: u8) -> Result<(), ArgumentError> {
+        if scale > 14 {
+            return Err(ArgumentError::InvalidArgs);
+        }
+
         if self.rx_buffer.capacity() < (1 << scale) as usize {
-            return Err(());
+            return Err(ArgumentError::InsufficientResource);
         }
 
         match self.state {
@@ -802,7 +829,7 @@ impl<'a> Socket<'a> {
                 self.remote_win_shift = scale;
                 Ok(())
             }
-            _ => Err(()),
+            _ => Err(ArgumentError::InvalidState),
         }
     }
 
@@ -8254,8 +8281,8 @@ mod test {
     #[test]
     fn test_too_large_window_scale() {
         let mut socket = Socket::new(
-            SocketBuffer::new(vec![0; 128]),
-            SocketBuffer::new(vec![0; 128]),
+            SocketBuffer::new(vec![0; 8 * (1 << 15)]),
+            SocketBuffer::new(vec![0; 8 * (1 << 15)]),
         );
         assert!(socket.set_local_recv_win_scale(15).is_err())
     }
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -523,7 +523,17 @@ impl<'a> Socket<'a> {
             panic!("receiving buffer too large, cannot exceed 1 GiB")
         }
         let rx_cap_log2 = mem::size_of::<usize>() * 8 - rx_capacity.leading_zeros() as usize;
+        let remote_win_shift = rx_cap_log2.saturating_sub(16) as u8;
+        Self::new_with_window_scaling(rx_buffer, tx_buffer, remote_win_shift)
+    }
 
+    /// Create a socket using the given buffers and window scaling factor defined in [RFC 1323].
+    ///
+    /// See also the [local_recv_win_scale](#method.local_recv_win_scale) method.
+    pub fn new_with_window_scaling<T>(rx_buffer: T, tx_buffer: T, recv_win_scale: u8) -> Socket<'a>
+    where
+        T: Into<SocketBuffer<'a>>,
+    {
         Socket {
             state: State::Closed,
             timer: Timer::new(),
@@ -543,7 +553,7 @@ impl<'a> Socket<'a> {
             remote_last_ack: None,
             remote_last_win: 0,
             remote_win_len: 0,
-            remote_win_shift: rx_cap_log2.saturating_sub(16) as u8,
+            remote_win_shift: recv_win_scale,
             remote_win_scale: None,
             remote_has_sack: false,
             remote_mss: DEFAULT_MSS,
@@ -772,6 +782,14 @@ impl<'a> Socket<'a> {
             // until the next packet, unless we wind up the timer explicitly.
             self.timer.set_keep_alive();
         }
+    }
+
+    /// Return the local receive window scaling factor defined in [RFC 1323].
+    ///
+    /// The value will become constant after the connection is established.
+    /// It may be reset to 0 during the handshake if remote side does not support window scaling.
+    pub fn local_recv_win_scale(&self) -> u8 {
+        self.remote_win_shift
     }
 
     /// Return the time-to-live (IPv4) or hop limit (IPv6) value used in outgoing packets.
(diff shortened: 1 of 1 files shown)


src/socket/tcp.rs
```diff
@@ -508,6 +508,26 @@ impl<'a> Socket<'a> {
...
+    }
+
+    /// Create a socket using the given buffers and window scaling factor defined in [RFC 1323].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the window scaling factor is greater than 14.
+    ///
+    /// See also the [local_recv_win_scale](#method.local_recv_win_scale) method.
+    pub fn new_with_window_scaling<T>(rx_buffer: T, tx_buffer: T, recv_win_scale: u8) -> Socket<'a>
```
tomDev5: You may want to add a `set_window_scaling` function, instead of `new_with_window_scaling`.
Forcing this choice to the intializing line differs from similar configurations:

- `set_ack_delay`
- `set_nagle_enabled`
- `set_keep_alive`
- `set_hop_limit`

XOR-op: By RFC 1323, window scaling can only be set in a SYN segment. Therefore we should force user to set this field in advance rather than give them illusion that it can be changed anytime.

tomDev5: That makes sense.
I just think about the future situation where you have another feature that is from the start only, and you want a socket with both features (like the TCP Timestamp feature I added a PR for).
A setter which checks for the TCP state (CLOSED/LISTEN) will solve this

XOR-op: That makes sense to me. Will update it later.

XOR-op:
@Dirbaio Would you mind giving a review?

XOR-op:
@whitequark / @thvdveld Would you mind giving some forms of review? It has been quite some time, but I haven't receive any feedback from maintainers yet.

whitequark:
Ping me again on a workday on UK time and I'll see if I can do it. Ideally Mon/Tue.

XOR-op:
@whitequark Thanks. It's workday today.

src/socket/tcp.rs
```diff
@@ -63,6 +63,30 @@ impl Display for ConnectError {
 #[cfg(feature = "std")]
 impl std::error::Error for ConnectError {}
 
+/// Error returned by set_*
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ArgumentError {
```
whitequark, line 69: I don't think the creation of a new error type that is used only in this API follows our API design principles.

src/socket/tcp.rs
```diff
@@ -749,6 +773,41 @@ impl<'a> Socket<'a> {
         }
     }
 
+    /// Return the local receive window scaling factor defined in [RFC 1323].
+    ///
+    /// The value will become constant after the connection is established.
+    /// It may be reset to 0 during the handshake if remote side does not support window scaling.
```
whitequark, line 804: I don't like that this API has a bidirectional data flow, where a value set by the consumer is read by the TCP/IP stack, but also then is updated by the TCP/IP stack. Is there any precedent (whether in smoltcp or in BSD TCP/IP) to have an option like this? I would there to be two API entry points, one to request an option, one to see if it was applied.

src/socket/tcp.rs
```diff
@@ -2516,6 +2580,53 @@ impl<'a> Socket<'a> {
...
+    ///
+    /// The requirements for the new buffer are:
+    /// 1. The new buffer must be larger than the length of remaining data in the current buffer
+    /// 2. The new buffer must be multiple of (1 << self.remote_win_shift)
+    ///
+    /// If the new buffer does not meet the requirements, the new buffer is returned as an error;
+    /// otherwise, the old buffer is returned as an Ok value.
+    ///
+    /// See also the [local_recv_win_scale](struct.Socket.html#method.local_recv_win_scale) methods.
+    pub fn replace_recv_buffer<T: Into<SocketBuffer<'a>>>(
```
whitequark, line 2679: This looks like a generic method that has no direct correspondence to TCP receive window resizing and as such I don't see why it should be a part of this PR at all.

whitequark (changes requested):
There is a number of design issues with this API and I think it needs to be discussed further before being implemented. As it is I don't think it is a good fit for smoltcp.

src/socket/tcp.rs
XOR-op: I don't expect a new error type for this API, too. However, I didn't find an appropriate existent error type to add enum variants. Do you have any suggestion for which of the existent {Send, Recv, Listen, Connect}Error should I add variants?

src/socket/tcp.rs
XOR-op: This method does enable the TCP receive window resizing. Smoltcp uses `rx_buffer.capacity()-rx_buffer.len()` to compute the current receive window. Without this method, it's impossible to adjust `rx_buffer.capacity()`. By replaced with a larger receive buffer, smoltcp knows it can advertise a larger receive window size safely without correctness issue.

XOR-op:
I believe some explanations are needed for this PR: the original purpose of this PR is to enable using a larger (or smaller) receive window after a TCP connection is established for better resource utilization. However, in the current implementation of smoltcp, the window scale is automatically computed, which will be the minimum value that can fully utilize the initial receive buffer. Assuming the initial buffer is 4MB, than the auto-computed window scale is log_2(4MB/64K)=6, which means we cannot replace with a buffer > 4MB. Thus we need to have ability to set window scale manually.

TL;DR:
1. We want to replace receive buffer dynamically (controlled by users).
2. To make sure we can use a larger buffer than the initial one, window scale should be able to be set manually.

whitequark:
I feel like these two goals shouldn't be addressed in the same PR. For example, given that a buffer is replaced (which I'm still not sure is the right way to handle this, but I'm not saying it's not) why not recompute the window size for the buffer size automatically?

src/socket/tcp.rs
XOR-op: In google's [gVisor userspace TCP/IP stack](https://github.com/google/gvisor/blob/ddfbb50ecc2d9b062d96b93a06aef86441e158d1/pkg/tcpip/transport/tcp/endpoint.go#L3077), they have similar options. I believe the issue is that, smoltcp has its default, and we want to override this default. That's why we have such a bidirectional data flow. Our TCP/IP stack reads the value (after connect/accept), and sets the value if no appropriate default value is set (before connect/accept).

As for the "two API entry points", I don't quite get that. Could you give some examples?

XOR-op:
> I feel like these two goals shouldn't be addressed in the same PR. For example, given that a buffer is replaced (which I'm still not sure is the right way to handle this, but I'm not saying it's not) why not recompute the window size for the buffer size automatically?

If the new buffer is always smaller than the initial buffer, than it has been automatically recomputed. But that will be less useful, since in most situations we want to increase the buffer size rather than reduce it. To be able to use a larger buffer after the establishment, we need to allow specifying window scale manually. (I don't add code for automatic re-computation, since that has been implemented even if without this PR.)

Also, to ensure the correctness of replacement, we need to access window scale anyway. That's why I see the two goals are related. And I'm afraid the smaller-only replacement will be accepted as a standalone PR due to lack of real usefulness. But I'll respect your opinion if you still think two PRs are needed.

XOR-op:
@whitequark Can I know your current opinion on our previous discussion (API designs and whether to split PR)? If you insist, I can split this change into two separate PRs (ability to replace with a smaller buffer is the first, ability to replace with a larger buffer & change window scale before establishing connections is the second). Without your comments it's difficult for me to advance this PR. Thanks in advance for your time and effort.

whitequark:
I'm overwhelmed at work currently so I'll only be able to handle this 1-2 weeks later. Feel free to ping me.

XOR-op:
@whitequark It has been two weeks now. Are you not so busy recently?

whitequark:
I'm on vacation currently, so I might be able to look at this PR in depth soon.

XOR-op:
@whitequark It has been another month since our last activity in this PR and hope you well. I think the two unresolved concerns are 1) whether we should add a new Error type; 2) how should we override the default window_scaling for API design. I hope to know thoughts from maintainers given the design challenges I faced, which are important for me to change my implementation.

whitequark:
I'll do my best to address this PR soon.

fire332:
Interested in getting a feature like this (not necessarily with the same impl) merged so I'm commenting here. I've thought through this quite a bit. smoltcp already uses the strategy pattern for congestion control so naturally, being able to use the same pattern for receive window scaling would be a good fit. My idea:

### `RingBuffer` should be fallibly resizable.

This would allow `tcp::Socket` to try and resize the buffer as needed.

- Add a `new_resizable` function that acts like `new` with a 2nd parameter taking a user-provided buffer factory.
- Add a `try_resize` method. Fallible to align with `smoltcp`'s API design.

```rs
/// The requested size.
/// `resizer` factory may return a buffer with a larger size than requested.
pub enum ResizeTo {
  Shrink(usize),
  Grow(usize),
}

pub fn new_resizable<S, F>(storage: S, resizer: F) -> RingBuffer<'a, T>
where
  S: Into<ManagedSlice<'a, T>>,
  F: FnMut(ManagedSlice<'a, T>, ResizeTo) -> S;
```

Then you would have this on `tcp::Socket`:

```rs
pub enum WindowSizing {
  Static,
  DynamicRightSize,
}

pub fn set_window_sizing(&mut self, strategy: WindowSizing);
```

This design allows the factory to return the same buffer (e.g. failed allocation). There would also be a provision for allowing the socket to swap its own strategy back to `WindowSizing::Static` upon resize failure.

whitequark:
> ### `RingBuffer` should be fallibly resizable.
> This would allow `tcp::Socket` to try and resize the buffer as needed.
> 
> * Add a `new_resizable` function that acts like `new` with a 2nd parameter taking a user-provided buffer factory.
> * Add a `try_resize` method. Fallible to align with `smoltcp`'s API design.

I like this approach.

fire332:
Would it be acceptable for me to file PRs implementing this? I'm thinking of splitting the task into separate PRs as follows:
1. feat(`tcp::Socket`): window size gains independence from buffer size. This will allow shrinking or expanding the window as required up to the current buffer's size.
2. feat(`RingBuffer`): can be resizable when created with a new constructor function that takes a buffer factory which behaves similar to `realloc`.
3. feat(`tcp::Socket`): add get/setters for a window sizing strategy similar to the current methods for congestion control strategy. Should it be fallible?

I can also file separate issues for each of the above tasks so we can nail down the API design before implementation.

whitequark:
> 1. feat(`tcp::Socket`): window size gains independence from buffer size. This will allow shrinking or expanding the window as required up to the current buffer's size.

Seems reasonable and uncontroversial.

> 2. feat(RingBuffer): can be resizable when created with a new constructor function that takes a buffer factory which behaves similar to realloc.

Seems reasonable also, though it might be nice to look at what the ecosystem has to offer for this design pattern.

> 3\. feat(`tcp::Socket`): add get/setters for a window sizing strategy similar to the current methods for congestion control strategy. Should it be fallible?

This is something I'd rather see a proposed design for before a PR.

XOR-op:
@fire332 Glad to see more attention on this matter. However I have some concerns over your design.

>    Add a `new_resizable` function that acts like `new` with a 2nd parameter taking a user-provided buffer factory.

I'm against this factory API design. The reasons are below:
1. It turns the explicit storage allocation to an implicit one. Developers now have no guarantee of when a resizing will happen. This adds complexity for resource management.
2. This API relies `S::drop` for resource reclamation. However, 1) sometimes it's much easier to implement an external reclamation instead of a callback; 2) even if another closure is provided, asynchronous reclamation is difficult. One example is that buffers are statically partitioned, and handles to the buffers are transferred by async channels. A closure style API makes the implementation just more complicated.

>   There would also be a provision for allowing the socket to swap its own strategy back to WindowSizing::Static upon resize failure.

Why? Unless you specified, you cannot know if a failure is temporarily unavailable resource, permanently insufficient resource or a bad argument. I don't think the implementation should infer anything here.

I guess your argument is that we may implement an auto-managed window if we use factory-style API. I agree that we can have one, but the factory should be managed by tcp socket itself, instead of the buffer(s). Developers can opt-in to use auto-managed window resizing if they can provide such a factory, or manually control if they want a finer-grained control or the management is too complicated.

fire332:
> It turns the explicit storage allocation to an implicit one. Developers now have no guarantee of when a resizing will happen. This adds complexity for resource management.

Tbh, I think the TCP impl can be refactored out of the socket struct/impl such that you can have completely zero copy with explicit resource management. If you use 2 pairs of SPSC rings, 2 for rx & 2 for tx, we can think of portions/slices of the ring as either owned by the app, the hardware -> ring transfer mechanism, or the TCP impl. Here's the embedded case's example for receive:

1. User/DMA peripheral (ethernet -> ring) writes raw packets directly to the receive ring. Interestingly enough, a high-end MCU like the STM32F7 has an ethernet MAC peripheral that can automatically timestamp frames and manage its own side of the ring buffers. If using a cheaper MCU, can always timestamp yourself with an interrupt.
2. User/DMA peripheral advances the tail, then calls the fallible receive fn on the TCP impl. Slice is either now owned by the TCP impl, or the tail advancement fails because missing packets are preventing the window from advancing.
3. In the same function call, TCP impl reads the packet and does its book keeping. During this, it will advance the ring's head if the window can advance past in-order packets. Those packets will now be owned by the user again (ring -> app).
4. User's app does w/e it wants with the data in the packet. Those slices should then be released back to the side managing transfers from ethernet -> ring.

This actually fits in perfectly with embedded ethernet peripherals that can advance and swap rings for you in a double-buffer per queue setup. You'll get the full benefit of the hardware peripheral, and the user can also resize the ring in the pair that's not being used according to feedback from the TCP impl (e.g. Dynamic Right-Size window scaling). A facade can then be thrown on top of all that to make it look like a unix socket and maintain complete backwards compatibility with the current API.

> This API relies S::drop for resource reclamation. However, 1) sometimes it's much easier to implement an external reclamation instead of a callback; 2) even if another closure is provided, asynchronous reclamation is difficult. One example is that buffers are statically partitioned, and handles to the buffers are transferred by async channels. A closure style API makes the implementation just more complicated.

It doesn't as the `realloc` function would be `FnMut` meaning it can take ownership of the passed in buffer while returning a new one for the socket. I see your point though. It's definitely nowhere close to a perfect API design, especially with how much fighting with the borrow checker that will be needed to implement a buffer factory that's compatible with the required lifetimes. With the socket's current design, it's really either have a dynamically resizable ring buffer with a `realloc` function, or have a leaky abstraction such that the user can see the "wanted" ring buffer size. Don't forget that the dynamic window scaling algorithm can just be left to slam up against the capacity of the ring buffer if the user doesn't want to play buffer hot potato (they can also allocate an extremely ring buffer upfront).

> Why? Unless you specified, you cannot know if a failure is temporarily unavailable resource, permanently insufficient resource or a bad argument. I don't think the implementation should infer anything here.

It can, but I haven't really thought that part through. If we want to continue with my initial idea, we should definitely nail the design down in a separate issue as @whitequark mentioned.

> Developers can opt-in to use auto-managed window resizing if they can provide such a factory

Just for clarity, I'll re-state again that users can have dynamic window sizing ***EVEN WITHOUT*** a `realloc`/buffer factory. The user can always allocate a much bigger buffer upfront, and the DRS algorithm will naturally shrink or grow the window up to the buffer's capacity for optimal latency.

whitequark:
> Tbh, I think the TCP impl can be refactored out of the socket struct/impl such that you can have completely zero copy with explicit resource management.

This is the skb approach that Linux takes (and some other TCP/IP stacks). I have intentionally not designed smoltcp to use it because I do not think that the complexity is justified. Also, smoltcp's entire reason for existence (a hard-to-trace memory safety "bug" in lwip) turned out to most likely be caused by a cache coherence issue with DMA. Placing the TCP/IP stack's control structures in potential harm's way of misbehaving DMA to pursue a small amount of efficiency gains isn't something I find inspiring.

(In other words: I see no particular reason to change the policy of "no unsafe code outside of `smoltcp::phy`".)

> Just for clarity, I'll re-state again that users can have dynamic window sizing _**EVEN WITHOUT**_ a `realloc`/buffer factory.

I agree that window management is orthogonal to buffer resizing, although I won't be able to say offhand how much of an issue (if at all) our current window management strategy is.

fire332:
> This is the skb approach that Linux takes

My idea actually stemmed from how io_uring's provided buffers work.

> I have intentionally not designed smoltcp to use it because I do not think that the complexity is justified.

Ofc, the ring buffer's design can be greatly simplified for the use case where a single thread completely manages the stack from top to bottom with no parallelism from other CPU cores or peripherals. That's why I was thinking of having SPSC/ring buffer traits in the crate and smoltcp can continue using its current ring impl for the simple case, and anyone who needs max speed can rip off the socket facade and implement their own rings compatible with the traits.

whitequark:
That doesn't (fully) eliminate complexity as smoltcp would still have to be redesigned around using these traits. It is not clear to me exactly how that would work so I cannot comment on how invasive it would be.

fire332:
I've already spent a few hours studying smoltcp's `tcp::Socket` implementation to think the architecture through so I feel like the idea has a good chance of being viable with minimal changes to the actual TCP code. If there isn't a hard no, I can outline the APIs and changes needed in a separate issue for such an endeavor.

whitequark:
I would like to see the API outline at least; if it really can be done without invasive changes I don't see a particular reason not to have this capability.

PR #934, opened 2024-05-23
SocketSet::add optimization 

VegetalDev:
Optimize `SocketSet::add` for the common use case of adding sockets in a loop, see https://github.com/smoltcp-rs/smoltcp/issues/926.


chenzhuoyu:
@Steanky Any updates on this ...?

thvdveld:
Thank you!

whitequark:
I think this PR needs a rebase to proceed.

VegetalDev:
There we go! Apologies for the slow response :)

Dirbaio:
After discussing this a bit more i'm going to close it. Sorry for making you rebase it for nothing.

The reasoning is:

*Everything* in smoltcp iterates the SocketSet linearly. `Interface` iterates it on ingress to match packets to sockets, and on egress to see which socket wants to transmit something. Even if we optimize `SocketSet::add()` so adding 40k sockets is fast, smoltcp will still be unusably slow because it'll iterate those 40k sockets *for every single sent or received packet*. Therefore optimizing `add` increases complexity for little benefit since it doesn't make high socket count feasible anyway.

`SocketSet` is designed like this because `smoltcp` primarily targets embedded systems, where you typically have RAM for only a handful of sockets so linearly iterating them is not an issue. Keeping the `SocketSet` code dumb and straightforward helps with code size, which is often also important in embedded.

*However* I agree it'd be nice to make smoltcp usable in larger systems. Doing this properly will require completely  redesigning `SocketSet`. It would have to
- hash sockets by IP protocol / endpoint / 4-tuple so lookup on ingress is fast
- quickly answer queries like "which sockets have a timer expired / want to send data" on egress, somehow.

This would massively increase code size so we'd have to keep the old linear implementation as well, and allow users to select one with Cargo features.

Since it's a large change, if you (or anyone else reading this) are willing to work on it please open an issue to discuss first.

PR #937, opened 2024-05-25
Use `core` instead of `std` for net types.

liamkinne:
Helpful for using `.into()` with these types on no_std platforms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "smoltcp"
 version = "0.11.0"
 edition = "2021"
-rust-version = "1.65"
+rust-version = "1.77"
 authors = ["whitequark <whitequark@whitequark.org>"]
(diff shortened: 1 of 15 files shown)


liamkinne:
To fix MSRV should this be gated behind an `ip_in_core` feature?

Dirbaio:
IMO we can just increase the MSRV, unless someone has some objection.

liamkinne:
Have brought the msrv up. If it's okay, I will fix the clippy lints in another PR to not pollute this one with unrelated changes.

crawford:
Bumping the MSRV brings with it the changes to Clippy's lints. In my opinion, it's fine to have the linting changes in this PR. If they're done in a separate commit, it's still easy to follow.

liamkinne:
Ready for review.

Dirbaio:
there's more places where MSRV should be updated, like the readme. please grep for "1.65" and update them all.

PR #939, opened 2024-06-01
Support TCP Timestamp option

tomDev5:
This is currently not relating to the RTT estimation, just for supporting remotes who do support it.
The generator isn't a lambda to avoid a generic.
Tests include a logic test of an established session and tests which make sure this option is only enabled if the both sides support it.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1,6 +1,6 @@
 // Heads up! Before working on this file you should read, at least, RFC 793 and
-// the parts of RFC 1122 that discuss TCP. Consult RFC 7414 when implementing
-// a new feature.
+// the parts of RFC 1122 that discuss TCP, as well as RFC 7323 for some of the TCP options.
+// Consult RFC 7414 when implementing a new feature.
 
 use core::fmt::Display;
 #[cfg(feature = "async")]
@@ -14,7 +14,7 @@ use crate::storage::{Assembler, RingBuffer};
 use crate::time::{Duration, Instant};
 use crate::wire::{
     IpAddress, IpEndpoint, IpListenEndpoint, IpProtocol, IpRepr, TcpControl, TcpRepr, TcpSeqNumber,
-    TCP_HEADER_LEN,
+    TcpTimestampGenerator, TcpTimestampRepr, TCP_HEADER_LEN,
 };
 
 mod congestion;
@@ -482,6 +482,12 @@ pub struct Socket<'a> {
     /// The congestion control algorithm.
     congestion_controller: congestion::AnyController,
 
+    /// tsval generator - if some, tcp timestamp is enabled
+    tsval_generator: Option<TcpTimestampGenerator>,
+
+    /// 0 if not seen or timestamp not enabled
+    last_remote_tsval: u32,
+
     #[cfg(feature = "async")]
     rx_waker: WakerRegistration,
     #[cfg(feature = "async")]
@@ -540,6 +546,8 @@ impl<'a> Socket<'a> {
             ack_delay_timer: AckDelayTimer::Idle,
             challenge_ack_timer: Instant::from_secs(0),
             nagle: true,
+            tsval_generator: None,
+            last_remote_tsval: 0,
             congestion_controller: congestion::AnyController::new(),
 
             #[cfg(feature = "async")]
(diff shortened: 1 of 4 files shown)


whitequark:
Looks good at a quick glance.

tomDev5:
Hi, I would appreciate if someone could review this PR (or approve if it indeed looks good)

src/socket/tcp.rs
```diff
@@ -549,6 +557,16 @@ impl<'a> Socket<'a> {
         }
     }
 
+    /// Enable or disable TCP Timestamp.
+    pub fn set_tsval_generator(&mut self, generator: Option<fn() -> u32>) {
```
whitequark: Should this be a `fn` or `FnMut`?

src/wire/tcp.rs
```diff
@@ -624,6 +625,7 @@ pub enum TcpOption<'a> {
     WindowScale(u8),
     SackPermitted,
     SackRange([Option<(u32, u32)>; 3]),
+    TimeStamp(u32, u32),
```
whitequark: This would be improved with named fields, though given the manipulation of this struct above (e.g. `reply_repr.timestamp = match (repr.timestamp, self.tsval_generator) {`) I think it would be even better as a proper structure with an inherent method to build a response struct.

whitequark (changes requested):
I would like you to also add RFC 7323 to the `src/socket/tcp.rs` file header, since it is also now necessary to understand the semantics. Once you're done please squash the commits.

src/socket/tcp.rs
tomDev5: having `generator` an `FnMut` would allow closures, which would make it much more flexible, but will require every user of `socket` to have a generic.
Another option is to have it behind a `Box`, but that would be an indirection on a function that will be called a lot (every packet)
I don't have a strong opinion, if you think it should be one of those two options, that can be easily done

tomDev5: having `generator` an `FnMut` would allow closures, which would make it much more flexible, but will require every user of socket to have a generic.
Another option is to have it behind a `Box`, but that would be an extra indirection on a function that will be called for every TCP packet.
I don't have a strong opinion, if you think I should change it to one of these options, that can be easily done

whitequark: Let's ship it as `fn()` and deal with it later if it becomes an issue.

tomDev5:
Added tests and re-squashed. Codecov didn't update with the new data, but (almost) all lines complained upon are now integrated in tests.

tomDev5:
Ok, the codecov window is refusing to update.
all complained upon lines are integrated in tests

whitequark:
> Ok, the codecov window is refusing to update.

Yeah so codecov is having some infra issues. That's ok. I talked to their CTO, they'll fix it eventually but not yet.

whitequark (changes requested):
This is a lot better but I'd still much prefer there to be a struct for the timestamp itself with inherent method(s) to manipulate it to the current situation with extensive tuple access. I can take a closer look and propose a specific structure if you'd like.

tomDev5:
Agreed, that made for cleaner access.

src/wire/tcp.rs
```diff
@@ -806,9 +819,30 @@ pub struct Repr<'a> {
     pub max_seg_size: Option<u16>,
     pub sack_permitted: bool,
     pub sack_ranges: [Option<(u32, u32)>; 3],
+    pub timestamp: Option<TcpTimestampRepr>,
     pub payload: &'a [u8],
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct TcpTimestampRepr {
+    tsval: u32,
```
whitequark: These fields should be public, since the accessors don't have any added behavior; all of the defined inherent methods don't bring anything over normal field access.

src/socket/tcp.rs
```diff
@@ -2246,6 +2286,14 @@ impl<'a> Socket<'a> {
             max_seg_size: None,
             sack_permitted: false,
             sack_ranges: [None, None, None],
+            timestamp: if let Some(ts_generator) = self.tsval_generator {
```
whitequark: I think this logic, as well as the logic I highlighted above, should live in an inherent method on `TcpTimestampRepr` which accepts the two timestamps and a TSVal generator and returns one you can use in response or None.

src/socket/tcp.rs
```diff
@@ -1331,6 +1350,14 @@ impl<'a> Socket<'a> {
     fn ack_reply(&mut self, ip_repr: &IpRepr, repr: &TcpRepr) -> (IpRepr, TcpRepr<'static>) {
         let (mut ip_reply_repr, mut reply_repr) = Self::reply(ip_repr, repr);
 
+        reply_repr.timestamp = match (repr.timestamp, self.tsval_generator) {
```
whitequark: (Also here.)

whitequark (changes requested):
Sorry, I was unclear, see inline comments.

tomDev5:
Leaving this unsquashed for easier diff, will squash after deemed ok

whitequark:
I'm on vacation for a week. Later!

tomDev5:
@whitequark gentle ping, regarding the fixes I've made

whitequark (approved):
Looks great! Thanks for the work you put into this and the fixes.

Issue #570, opened 2021-11-17
SLAAC support


Rahix:
I'd be interested in implementing IPv6 SLAAC support for smoltcp.  I wonder what the best way to do this would be?  I see DHCPv4 is implemented as a socket type.  Would it be a good idea to go the same route here?

I did some first tests, starting an implementation outside of smoltcp, but got stuck as the IcmpSocket does not allow receiving the router advertisements (because its filters are too strict).

astro:
What I really want for SLAAC is to participate in IPv6 Multicast Listener Discovery to that IPv6-aware Ethernet switches don't filter out the advertisements on the port that smoltcp is connected to.

That's why I started working on multicast in the first place. Sorry that I didn't make it beyond IPv4.

DHCPv4 was started as a RawSocket, and was later switched to its own socket type for the reasons discussed in #459.

therealprof:
DHCPv4 is special as it kind of sits in between layers (which is also the reason why the responsibility for handling it is often divvied into a kernel and userspace part). SLAAC and DHCPv6 are much clearer cut and should be implementable with the just a few generic prerequisites:

- Support of multiple addresses per network interface
- Support for multicast addressing
- DAD (which in turn requires use of ND and ideally MLD as mentioned by @astro)

dlrobertson:
@Rahix I never got around to implementing SLAAC when I implemented the basics of IPv6, but I was interested in it. I'd be happy to help where needed and could definitely assist in reviews.

dscso:
Are there some updates related to this issue?
I am currently doing some very hacky thing to achieve IPv6 connectivity by compiling a static IPv6 address into my source code, and I'd like to do it the right way.
Are there already some implementation attempts yet? If so I'd be glad. Otherwise I'd might try my luck to implement SLAAC.

I was looking where to implement this, and I found this piece of code:

https://github.com/smoltcp-rs/smoltcp/blob/ef67e7b46cabf49783053cbf68d8671ed97ff8d4/src/iface/interface/ipv6.rs#L376-L387


I think reacting on a Router Solicitation by adding an IP address could be a solution. The only problem is how to select the suffix.

I see 2 choices:
- use mac address
- user defined

the problem of the mac address is that it would not work if the router cdir is too small and the user defined option would be add a lot of complexity. Alternatively the user defined could be "seed based" so the user provides a seed to the network stack and the network stack generates the suffix based on the seed.

[...]

PR #948, opened 2024-07-03
Adding SLAAC support

dscso:
Could you please take a look if I am going the right way before I implement the whole thing :D

These are my ideas:
- creating a new socket type (heavily DHCP inspired)
- leave the IPv6 generation to the user of the library

My questions:
- how to handle DNS? SLAAC does not have DNS advertisement support

ref #570

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -342,7 +342,7 @@ impl InterfaceInner {
 
     pub(super) fn process_icmpv6<'frame>(
         &mut self,
-        _sockets: &mut SocketSet,
+        sockets: &mut SocketSet,
         ip_repr: Ipv6Repr,
         ip_payload: &'frame [u8],
     ) -> Option<Packet<'frame>> {
@@ -360,7 +360,7 @@ impl InterfaceInner {
         #[cfg(feature = "socket-icmp")]
         {
             use crate::socket::icmp::Socket as IcmpSocket;
-            for icmp_socket in _sockets
+            for icmp_socket in sockets
                 .items_mut()
                 .filter_map(|i| IcmpSocket::downcast_mut(&mut i.socket))
             {
@@ -393,9 +393,9 @@ impl InterfaceInner {
             #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
             Icmpv6Repr::Ndisc(repr) if ip_repr.hop_limit == 0xff => match self.caps.medium {
                 #[cfg(feature = "medium-ethernet")]
-                Medium::Ethernet => self.process_ndisc(ip_repr, repr),
+                Medium::Ethernet => self.process_ndisc(sockets, ip_repr, repr),
                 #[cfg(feature = "medium-ieee802154")]
-                Medium::Ieee802154 => self.process_ndisc(ip_repr, repr),
+                Medium::Ieee802154 => self.process_ndisc(sockets, ip_repr, repr),
                 #[cfg(feature = "medium-ip")]
                 Medium::Ip => None,
             },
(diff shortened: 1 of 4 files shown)


chrysn:
Leaving the IP generation to the user is a bit odd, but then again, better to have this in first and then add a way for the user to pick an automatism for it.

DNS through SLAAC just works (and has been standardized since 2010, but people kept repeating the talking point): The same Router Advertisement that gives you the prefix information and the routes will also give you (in [option 25](https://www.rfc-editor.org/rfc/rfc8106)) the IP address of a recursive DNS server.

chrysn:
Does this really get simpler by using a new socket type? There is already the ICMP socket, and that can be bound to the relevant ICMP identifiers for Router Advertisement and Neighbor Advertisement, and send Router / Neighbor Solicitation.

dscso:
> DNS through SLAAC just works (and has been standardized since 2010, but people kept repeating the talking point): The same Router Advertisement that gives you the prefix information and the routes will also give you (in [option 25](https://www.rfc-editor.org/rfc/rfc8106)) the IP address of a recursive DNS server.

image
True you are right. 

Anyways DNS parsing has to be added to smoltcp

dscso:
1. The ICMP socket is a great idea I didn't see that. Do you think SLAAC shouldn't be handled from a users perspective? My thought was to handle it similarly to DHCP (see below), so that the user adds a SLAAC-like Socket, that requests and awaits a Router Advertisement and returns the Prefix, DNS and a IPv6 address suggestion (see 2.). The user of the library is then responsible to assign the IPv6 address + DNS to the interface. What do you think about this idea?

https://github.com/smoltcp-rs/smoltcp/blob/1ff5bc41eec296da12be8dd6a251569c1ad7f5e8/examples/dhcp_client.rs#L60-L77

2. Regarding IPv6 address generation, I could do the following:
2.1 Either use (see below) for random address generation
2.2 Or just use the mac address in the last 6 bytes of the IPv6 address 

https://github.com/smoltcp-rs/smoltcp/blob/99ddd22cdbe4ca59f894c4ba2588b8d4f52a08a6/src/iface/interface/mod.rs#L132

chrysn:
ad 1., I think that many developers will eventually expect a choice akin to what NetworkManager gives you ("use MAC address based", "generate both and primarily use {which of them}"). I don't know enough about the scope of smoltcp yet to know whether that's a choice that smoltcp should offer, or embassy_net (although I'd hope that the crate that offers the options would take a smoltcp network stack, plus maybe an impl RngCore and impl embedded_hal_async::timer something, so it could be used outside embassy_net as well).

If DHCPv4 is implemented this way, it probably makes sense here as well; DHCPv4 also needs to have at least access to a timer, so whatever is done there sounds like a good template.

ad 2.1: You may not need a random number generator, just a stable per-device secret. [RFC7217](https://www.rfc-editor.org/rfc/rfc7217.html) is state of the art there.

ad 2.2: MAC addresses may just have 6 bytes, but there is an 8-byte form called the EUI-64 where they are embedded, and there are precise rules on how those wind up in the interface identifier (low 8 bytes) of an IPv6 address [in section 2.5.1 of RFC4291](https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.1). These are what gives you the typical "fe80::xXxx:xxff:fexx:xxxx" IPv6 addresses you may have seen. (Note how if you look at the output of `ip -6 a`, in the position of the X, the address differs from the MAC address; this was done so that you can assign addresses like "fe80::1" manually without conflicting with MAC addresses).

---

But having SLAAC implemented with a "proposing IPv6 addresses for DAD is up to the user" interface sounds totally sufficient to me as a first iteration, as long as the SLAAC can ask that user provided function for another address (eg. providing a DAD_Counter).

Note that I'm not really an expert on any of this, and have barely contributed to smoltcp at all, so take all I'm writing here with a grains of salt, minimum.

PR #950, opened 2024-07-08
Use `assert!()` to check `prefix_len` in `Ipv4Cidr::new`

wetheredge:
Updated to match [`Ipv6Cidr::new`](https://github.com/smoltcp-rs/smoltcp/blob/99ddd22cdbe4ca59f894c4ba2588b8d4f52a08a6/src/wire/ipv6.rs#L485)


thvdveld:
Thank you!

PR #954, opened 2024-07-15
Limiting the number of poll loops to prevent DoS events

### LINKED ISSUE #848: Packet flooding can cause `interface.poll()` to require excessive processing time
The logic behind `interface.poll()` repeatedly processes packet ingress and egress. However, in cases where packets may be arriving faster than they can be read and processed (i.e. heavy multicast/broadcast traffic, slow transfer times with the MAC/PHY), this can cause the `interface.poll()` function to block for indefinite periods of time.

I specifically encountered this while streaming UDP multicast traffic on a LAN at ~40MBps on a separate device, and observed an unrelated piece of hardware (on the same physical network, both running smoltcp) then encounter a watchdog and reset (the watchdog had a 4s timer, streaming was occurring for ~10 seconds).

I believe this could potentially be used as a fault injector and/or DoS mechanism on firmware projects. Would it make sense to limit the infinite process loop in `interface.poll()` to some theoreticaly maximum packet count, and then allow the user to call the function again?

ryan-summers:
This PR fixes #848 by limiting the number of ingress/egress tokens processed by observing the characteristics of the sockets contained in the polled socket set.

---
Edit: Further information on hardware testing can be found in https://github.com/quartiq/booster/pull/423

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -396,6 +396,13 @@ impl Interface {
     /// This function returns a boolean value indicating whether any packets were
     /// processed or emitted, and thus, whether the readiness of any socket might
     /// have changed.
+    ///
+    /// # Note
+    /// This function performs a bounded amount of work per call to avoid
+    /// starving other tasks of CPU time. If it returns true, there may still be
+    /// packets to be received or transmitted. Depending on system design,
+    /// calling this function in a loop may cause a denial of service if
+    /// packets cannot be processed faster than they arrive.
     pub fn poll<D>(
         &mut self,
         timestamp: Instant,
@@ -429,23 +436,12 @@ impl Interface {
             }
         }
 
-        let mut readiness_may_have_changed = false;
+        let mut readiness_may_have_changed = self.socket_ingress(device, sockets);
+        readiness_may_have_changed |= self.socket_egress(device, sockets);
 
-        loop {
-            let mut did_something = false;
-            did_something |= self.socket_ingress(device, sockets);
-            did_something |= self.socket_egress(device, sockets);
-
-            #[cfg(feature = "proto-igmp")]
-            {
-                did_something |= self.igmp_egress(device);
-            }
-
-            if did_something {
-                readiness_may_have_changed = true;
-            } else {
-                break;
-            }
+        #[cfg(feature = "proto-igmp")]
+        {
+            readiness_may_have_changed |= self.igmp_egress(device);
         }
 
         readiness_may_have_changed
@@ -507,67 +503,65 @@ impl Interface {
     {
         let mut processed_any = false;
 
-        while let Some((rx_token, tx_token)) = device.receive(self.inner.now) {
-            let rx_meta = rx_token.meta();
-            rx_token.consume(|frame| {
-                if frame.is_empty() {
-                    return;
-                }
+        let Some((rx_token, tx_token)) = device.receive(self.inner.now) else {
+            return processed_any;
+        };
 
-                match self.inner.caps.medium {
-                    #[cfg(feature = "medium-ethernet")]
-                    Medium::Ethernet => {
-                        if let Some(packet) = self.inner.process_ethernet(
-                            sockets,
-                            rx_meta,
-                            frame,
-                            &mut self.fragments,
-                        ) {
-                            if let Err(err) =
-                                self.inner.dispatch(tx_token, packet, &mut self.fragmenter)
-                            {
-                                net_debug!("Failed to send response: {:?}", err);
-                            }
-                        }
-                    }
-                    #[cfg(feature = "medium-ip")]
-                    Medium::Ip => {
-                        if let Some(packet) =
-                            self.inner
-                                .process_ip(sockets, rx_meta, frame, &mut self.fragments)
+        let rx_meta = rx_token.meta();
+        rx_token.consume(|frame| {
+            if frame.is_empty() {
+                return;
+            }
+
+            match self.inner.caps.medium {
+                #[cfg(feature = "medium-ethernet")]
+                Medium::Ethernet => {
+                    if let Some(packet) =
+                        self.inner
+                            .process_ethernet(sockets, rx_meta, frame, &mut self.fragments)
+                    {
+                        if let Err(err) =
+                            self.inner.dispatch(tx_token, packet, &mut self.fragmenter)
                         {
-                            if let Err(err) = self.inner.dispatch_ip(
-                                tx_token,
-                                PacketMeta::default(),
-                                packet,
-                                &mut self.fragmenter,
-                            ) {
-                                net_debug!("Failed to send response: {:?}", err);
-                            }
-                        }
-                    }
-                    #[cfg(feature = "medium-ieee802154")]
-                    Medium::Ieee802154 => {
-                        if let Some(packet) = self.inner.process_ieee802154(
-                            sockets,
-                            rx_meta,
-                            frame,
-                            &mut self.fragments,
-                        ) {
-                            if let Err(err) = self.inner.dispatch_ip(
-                                tx_token,
-                                PacketMeta::default(),
-                                packet,
-                                &mut self.fragmenter,
-                            ) {
-                                net_debug!("Failed to send response: {:?}", err);
-                            }
+                            net_debug!("Failed to send response: {:?}", err);
                         }
                     }
                 }
-                processed_any = true;
-            });
-        }
+                #[cfg(feature = "medium-ip")]
+                Medium::Ip => {
+                    if let Some(packet) =
+                        self.inner
+                            .process_ip(sockets, rx_meta, frame, &mut self.fragments)
+                    {
+                        if let Err(err) = self.inner.dispatch_ip(
+                            tx_token,
+                            PacketMeta::default(),
+                            packet,
+                            &mut self.fragmenter,
+                        ) {
+                            net_debug!("Failed to send response: {:?}", err);
+                        }
+                    }
+                }
+                #[cfg(feature = "medium-ieee802154")]
+                Medium::Ieee802154 => {
+                    if let Some(packet) =
+                        self.inner
+                            .process_ieee802154(sockets, rx_meta, frame, &mut self.fragments)
+                    {
+                        if let Err(err) = self.inner.dispatch_ip(
+                            tx_token,
+                            PacketMeta::default(),
+                            packet,
+                            &mut self.fragmenter,
+                        ) {
+                            net_debug!("Failed to send response: {:?}", err);
+                        }
+                    }
+                }
+            }
+            processed_any = true;
+        });
 
         processed_any
     }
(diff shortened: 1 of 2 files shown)


jordens:
Why not get rid of the loop and let the application decide on the strategy? Or, asking differently, why is `1` not the appropriate iteration count here? Or why is it not something like the number of rx/tx buffer slots of the device (i.e. either derived from some qty known by smiltcp or to be chosen by the application)?

ryan-summers:
I like the idea of basing the poll iteration count based on buffer sizing and data available to smoltcp, so I've updated it to calculate a maximum iteration count based on the sockets (where this makes sense). For protocols that do not buffer data in a packetized format (i.e. TCP, DHCP, DNS), I've simply set the maximum iteration count to 1, since these protocols could theoretically have an infinite sequence of packet exchange and the user application should likely figure out how to handle polling.

src/iface/interface/mod.rs
```diff
@@ -396,6 +396,9 @@ impl Interface {
     /// This function returns a boolean value indicating whether any packets were
     /// processed or emitted, and thus, whether the readiness of any socket might
     /// have changed.
+    ///
+    /// # Note
+    /// If this function returns true, there may still be packets to be transmitted or received.
```
whitequark: This note should explain *why*, because as it is the obvious response to reading this note is to do it in a loop, but this just reintroduces the original DoS.

src/iface/interface/mod.rs
```diff
@@ -429,11 +432,51 @@ impl Interface {
...
+                Socket::Udp(s) => s.packet_send_capacity(),
+                #[cfg(feature = "socket-tcp")]
+                Socket::Tcp(_) => 1,
+                #[cfg(feature = "socket-dhcpv4")]
+                Socket::Dhcpv4(_) => 1,
+                #[cfg(feature = "socket-dns")]
+                Socket::Dns(_) => 1,
+            })
+            .max()
+            .unwrap_or(1);
```
whitequark: Is this actually a reasonable way to calculate the number of polls? Or in other words, besides helping to prevent DoS, does this help the API consumers which care about completing the poll in any well-defined amount of time, for example to do some other real-time processing?

The function can't be redefined to do one iteration since this will be O(n^2) with a large amount of sockets, therefore it should probably have an additional variant that polls up to a predefined amount of times. It's not clear to me what the existing poll function should do, maybe call the new variant with `usize::MAX`?

whitequark:
@jordens While I agree in principle with the idea of basing the amount of polling on some function of the total buffer size I don't feel that smoltcp can reasonably calculate this on its own (the strategy in this PR doesn't seem reasonable to me), and this should probably be left to the application, since the final application itself is the only entity which is aware of the performance requirements and the characteristics of the platform. If we do go with smoltcp calculating the iteration count on the buffer size this should be the PHY buffer size rather than the stack buffer size.

An alternative approach, more principled but also requiring considerably more work, is to make an opaque polling iterator for the socket set (i.e. use external iteration instead of the internal iteration implicitly done by `poll` currently) which can be saved between polling attempts, so that the application can know it has done a full cycle of polling, ignoring the pauses caused by TX buffer becoming full. I feel like we may run into issues with lifetimes for this approach though.

jordens:
The iteration approach sounds neat!
But if that turns out to be difficult or only long-term, let's go for a single iteration now. Then the application can decide the strategy.
Always assuming that a low iteration count has no negative impact: I see there is some the fragmentation stuff and some ipv4/sixlowpan egress outside the loop.

whitequark:
> But if that turns out to be difficult or only long-term, let's go for a single iteration now. Then the application can decide the strategy.

Imagine that you have 1000 sockets and each of them has a single packet to send (and the TX buffer never becomes full, e.g. you're on a PC). You'll then have to iterate through the socket set 499500000 times to send all of the packets, am I wrong?

jordens:
Maybe. But I may be misundertanding you. I tought we are talking about removing the `loop {}` around this block here:

https://github.com/smoltcp-rs/smoltcp/blob/7b125ef6010311ea6cbd432496cace3b299e6b29/src/iface/interface/mod.rs#L434-L449

Ignoring the fragmentation and timestamp stuff before the loop, I can't see how allowing the application to do that loop over `poll()` (thus "deciding the strategy" as we both proposed) instead of having `poll()` loop over `ingress+egress` (as it is now) changes anything performance-wise.

whitequark:
@jordens Oh, you are right. I was relying too much on my memory and wasn't paying enough attention. I agree that only running a single iteration of that loop would be fine.

ryan-summers:
> Maybe. But I may be misundertanding you. I tought we are talking about removing the `loop {}` around this block here:
> 
> https://github.com/smoltcp-rs/smoltcp/blob/7b125ef6010311ea6cbd432496cace3b299e6b29/src/iface/interface/mod.rs#L434-L449

There's actually another loop in `socket_ingress()` that repeatedly polls the PHY for RX tokens that has to be broken, otherwise removal of the outermost loop doesn't solve the problem.

whitequark:
So if we change both loops to a single iteration then the problem I described does arise.

But if you change the loop in `socket_ingress()` to poll up to the number of PHY RX buffers, it should be fine. I think we have the number of PHY TX buffers available in PHY properties; do we have it for RX buffers? If not we can add it.

(Also, if we do limit the inner loop, is the outer loop still an issue?)

ryan-summers:
> So if we change both loops to a single iteration then the problem I described does arise.

I suspect so, although I didn't completely follow your math :)
 
> But if you change the loop in `socket_ingress()` to poll up to the number of PHY RX buffers, it should be fine. I think we have the number of PHY TX buffers available in PHY properties; do we have it for RX buffers? If not we can add it.

I didn't know if this property was available on the PHY trait, but this was what I was thinking in the back of my mind when drafting up the changes. I'll take a look at this shortly, thanks for the feedback! 


> (Also, if we do limit the inner loop, is the outer loop still an issue?)

I believe so, yes. Since essentially the problem we observe is that the PHY is essentially generating RX tokens faster than they can be processed. Any kind of `loop {}` over RX token processing will always block infinitely.

whitequark:
@ryan-summers I'm happy to proceed changing both loops to a single iteration; like Robert says this should not have detrimental side effects, as far as I can tell.

ryan-summers:
Yup sorry, meant to get to this on Thursday, but am now traveling until Monday. I'm planning on going with the single loop approach, thanks for clarifying :)

src/iface/interface/mod.rs
ryan-summers: I expanded the why in 6492c522feb4020de3fa8aa7d509c2ce9f250af8, let me know if you have other verbiage you'd prefer and I'm happy to incorporate it :)

src/iface/interface/mod.rs
ryan-summers: You make a good point here, and after discussion below we have changed it to poll a single loop of the socket set instead.

ryan-summers:
@whitequark Let me know if these updated changes look appropriate. I'll run this PR on hardware in my office in the next day or two to verify it as well.

src/iface/interface/mod.rs
```diff
@@ -396,6 +396,12 @@ impl Interface {
     /// This function returns a boolean value indicating whether any packets were
     /// processed or emitted, and thus, whether the readiness of any socket might
     /// have changed.
+    ///
+    /// # Note
+    /// If this function returns true, there may still be packets to be transmitted or received.
+    /// Users should should perform other tasks before calling this function again. If executed in
+    /// a loop, its possible that packets can arrive faster than they can be processed, which could
+    /// cause an infinite loop.
```
whitequark:
```suggestion
    /// This function performs a bounded amount of work per call to avoid exhausting
    /// starving other tasks of CPU time. If it returns true, there may still be packets
    /// to be received or transmitted. Depending on system design, calling this function
    /// in a loop may cause a denial of service: it is safe if packets can be processed
    /// faster than they arrive, and unsafe otherwise.
```

src/iface/interface/mod.rs
whitequark: Done

src/iface/interface/mod.rs
```diff
@@ -398,10 +398,11 @@ impl Interface {
     /// have changed.
     ///
     /// # Note
-    /// If this function returns true, there may still be packets to be transmitted or received.
-    /// Users should should perform other tasks before calling this function again. If executed in
-    /// a loop, its possible that packets can arrive faster than they can be processed, which could
-    /// cause an infinite loop.
+    /// This function performs a bounded amount of work per call to avoid exhausting
```
whitequark: Oh, typo:

```suggestion
    /// This function performs a bounded amount of work per call to avoid
```

(I guess this also makes formatting uglier... sigh)

ryan-summers: Yeah sorry, I saw this already and removed it, and indeed the formatting is ugly

ryan-summers:
Alright, I have now tested this on hardware and confirmed the problem is still fixed with these changes.

src/iface/interface/mod.rs
```diff
@@ -396,6 +396,14 @@ impl Interface {
...
     /// processed or emitted, and thus, whether the readiness of any socket might
     /// have changed.
+    ///
+    /// # Note
+    /// This function performs a bounded amount of work per call to avoid
+    /// starving other tasks of CPU time. If it returns true, there may still be
+    /// packets to be received or transmitted. Depending on system design,
+    /// calling this function in a loop may cause a denial of service: it is
+    /// safe if packets can be processed faster than they arrive, and unsafe
+    /// otherwise.
```
dnadlinger: Hmm, wouldn't "safe"/"unsafe" sound nebulously dangerous to many users? Maybe just `Depending on system design, calling this function in a loop may cause a denial of service (if packets cannot be processed faster than they arrive).`?

ryan-summers: That's a good point, since rust has a formal definition of `unsafe`, and it's perfectly `safe` in the rust definition to hit the DoS conditions.

ryan-summers: Fixed in 2bcae4a6985d580a8bb6569b5a9d0e02f5be8602

whitequark: > Hmm, wouldn't "safe"/"unsafe" sound nebulously dangerous to many users?

I think a denial of service is plenty unsafe though. (Memory unsafety is one particular type of unsafety.)

whitequark (commented):
Looks good, please squash and I'll merge.

ryan-summers:
> Looks good, please squash and I'll merge.

I've done the squash manually. However, I thought Github added the ability for contributors / admins to squash before merge? If it's not showing up on the merge options list, you can probably enable it in the Github repository settings as well :)

whitequark:
> However, I thought Github added the ability for contributors / admins to squash before merge? If it's not showing up on the merge options list, you can probably enable it in the Github repository settings as well :)

Nope, this is not available when merge queue is used. Which is annoying but unfortunately the merge queue is a lot more important.

PR #955, opened 2024-07-18
[Breaking] Update to bitflags 2.x

GnomedDev:
Gets rid of this old duplicate dependency in my dep tree.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ managed = { version = "0.8", default-features = false, features = ["map"] }
 byteorder = { version = "1.0", default-features = false }
 log = { version = "0.4.4", default-features = false, optional = true }
 libc = { version = "0.2.18", optional = true }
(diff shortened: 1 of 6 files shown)


whitequark:
Does defmt no longer provide a derive macro?

GnomedDev:
bitflags 2.x swapped the internals of each bitflag type from just a newtype around the integer to a newtype around a private newtype. This means that derives that just try to touch the inner value can't work, hence having to make our own macro to implement formatting the bits.

whitequark:
Thanks. @Dirbaio what do you think about the macro? it feels a bit heavyweight but maybe that's just me.

GnomedDev:
I can change the macro to be separate and not replace the bitflags invocation, if that would be preferred, as in
```rust
bitflags::bitflags! {
	pub struct Flags: u8 {
		...
	}
}

impl_bitflags_defmt!(Flags);
```

PR #958, opened 2024-07-24
Fix performance related pedantic clippy lints

GnomedDev:
This PR enables a couple of lints from the "pedantic" group that can improve performance.

### DIFF
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -95,7 +95,7 @@ fn main() {
     };
     config.random_seed = rand::random();
 
-    let mut iface = Interface::new(config, &mut device, Instant::now());
+    let mut iface = Interface::new(&config, &mut device, Instant::now());
     iface.update_ip_addrs(|ip_addrs| {
         ip_addrs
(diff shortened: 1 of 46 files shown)


whitequark:
As far as I can see this is mostly noise. (I'm also personally opposed to use of clippy in general, although opt-in lints are the least bad mode.)

whitequark:
For example, "redundant_else" cannot ever have any impact on performance as it does not change codegen. This is pure overreach with no useful outcome, only churn.

GnomedDev:
Your dismissal and closing without discussion have come across as super hostile to new contributions, you could have just told me to undo said change. I often solve clippy lints to become familiar with a codebase but I'm not going to be looking into this any more.

whitequark:
I feel the same way about your PR; I see it as a statement that boils down to "this codebase isn't to my taste and I would like it to be, without regard to the time of people maintaining it". This is why I responded the way I did. Historically, interactions like this contributed to burnout, so I've decided to make sure I have less of them.

PR #962, opened 2024-07-28
socket: dns: expose MAX_DNS_SERVER_COUNT

mirkomartn:
Provide a `pub const fn` that returns the const `MAX_DNS_SERVER_COUNT`. This will allow users of the library to guard against triggering a panic by passing a too long slice to `Socket::update_servers()`.

### DIFF
--- a/src/socket/dns.rs
+++ b/src/socket/dns.rs
@@ -167,7 +167,7 @@ impl<'a> Socket<'a> {
     ///
     /// # Panics
     ///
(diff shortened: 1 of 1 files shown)


Dirbaio:
this is already exposed here https://docs.rs/smoltcp/latest/smoltcp/config/constant.DNS_MAX_SERVER_COUNT.html

mirkomartn:
Pardon, didn't look carefully enough!

PR #964, opened 2024-07-31
Added waker_registered query method

tomDev5:
Simple PR which will allow not leaving unawaken `Waker`s

### DIFF
--- a/src/socket/icmp.rs
+++ b/src/socket/icmp.rs
@@ -160,6 +160,12 @@ impl<'a> Socket<'a> {
         self.rx_waker.register(waker)
     }
 
+    /// Returns `true` if a recv waker is registered.
+    #[cfg(feature = "async")]
+    pub fn recv_waker_registered(&self) -> bool {
+        self.rx_waker.registered()
+    }
+
     /// Register a waker for send operations.
     ///
     /// The waker is woken on state changes that might affect the return value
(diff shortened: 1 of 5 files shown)


Dirbaio:
What is the use case for this? in my experience it's rare to want to know whether a waker is registered, you just register it every time you poll.

tomDev5:
If I am waiting for the socket to become ready for writing, and then something makes my application want to send more data, nothing prevents me from stealing the Waker, and leaving the other writing task undone.
With these methods I could at least know to return some error, or wait for a period of time.

Dirbaio:
The waker is woken when anything about the socket changes. Having no waker doesn't mean the other task's operation is done, the other task might still get poled laterand register the waker again because the operation it was waiting for is not done yet.

So you can still have bugs like:

- Task A starts one operation, registers the waker
- Smoltcp task polls the interface, something in the socket (unrelated to the operation!) changes, the waker gets woken+deregistered.
- Task B starts another operation. It sees there's no waker, so it happily registers its own.
- Task A gets polled again because it was woken before. It checks the socket state, it sees the thing it was waiting for hasn't completed yet. Then what? it registers the waker again overwriting B's? It errors out?

In general you can't safely write to a socket from two tasks at the same time with just one waker. It's not a substitute for a mutex. I'd recommend you make writing/reading require `&mut self` in the socket in your higher-level async wrapper, to prevent concurrent usage from multiple tasks, which fixes these waker races. This is [what embassy-net does](https://github.com/embassy-rs/embassy/blob/main/embassy-net/src/tcp.rs#L440), for example. Then if you want to share the socket between tasks, wrap it in an async Mutex.

tomDev5:
In the use case you proposed, when task A is polled and deregistered, if it cannot complete it should re-register itself to the waker, so that it's called again.

Assuming single async thread works on any given socket (which I understand may be a problem, but it kind of already is - smoltcp is not protecting the current wakers), task B will not have a chance of waking while the A's waker is not registered.

Am I getting it correctly?

Dirbaio:
When a task gets woken it gets added to the executor's queue, but there's no guarantee that it's going to be polled immediately next. It's definitely possible for task B to sneak in between the smoltcp task and task A.

This has nothing to do with threads, it is the case with both single-threaded and multi-threaded executors.

tomDev5:
Understood, thanks for the clarification. Will close this PR.

PR #968, opened 2024-08-05
Fix wrong feature flag for UDP broadcast test

thvdveld:
The UDP broadcast test was using the wrong feature flag. This commit fixes the feature flag to use the correct one. This must have slipped through the cracks when I updated the tests to use rstest.


Dirbaio (approved):
those attrs are starting to be intimidating heh

PR #969, opened 2024-08-05
Fix target-arch names and remove unknonwn fuzzing cfg warning

thvdveld:
The endianness of the target architecture cannot be specified using target_arch. The endianness can be specified using target_endian.

The target description for mipsel-unknown-linux-gnu is found here:
https://github.com/rust-lang/rust/blob/master/compiler/rustc_target/src/spec/targets/mipsel_unknown_linux_gnu.rs
In this specification, the arch is set to "mips" and not "mipsel".

@blechschmidt: since you added this cfg for these architectures, could you check if smoltcp still compiles for you?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ license = "0BSD"
 # ensure that the correct features are enabled.
 autoexamples = false
 
+[lints.rust]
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
nice! thanks

PR #970, opened 2024-08-06
Fix mismatching src ip address in TCP socket transmissions

### LINKED ISSUE #466: TCP sockets can transmit packets with wrong source IP

MathiasKoch:
This is an attempt to close TCP sockets if the src address changes, while the socket is open.

I have been testing this on my application, and it seems to do the trick. I have also tried with both `socket.close()` and `socket.abort()` in place of the `socket.reset()` call currently in here, but neither of them quite seems to do the trick, as the tuple is not set to `none`?

I am not a fan of making `reset` public here, so this is mostly an attempt at getting some feedback from you guys?

Fixes #466


whitequark (commented):
I don't understand what the motivation is, could you elaborate on that please?

MathiasKoch:
Ahh, sorry. I assumed the linked issue was enough context. Sorry about that.

In my case, I am running a single smoltcp stack, but changing between two PPP based interfaces in a super crude way (because smoltcp does not support actual multi-interface routing).
I only ever need one interface active at a time, and use the other as fallback (cellular connection). I have no means of notifying higher level consumer libraries of my sockets, that the interface has changed and that they should thus drop the current connection and reconnect.

This means that the src address on any open socket, will no longer match the address when the socket was opened, but the socket stays established and still transmits packets with the old source IP address.

This PR is an attempt at making sure a host never transmits packets with a source IP that's not its own, and at the same time closes the socket, allowing socket consumers to immediately notice and reconnect.

whitequark:
This feels like a fairly niche use case that I'm not sure we should support in the core library. In particular the lack of ability to notify the upper levels of connectivity change feels like a design issue that isn't necessarily something that should have a workaround in smoltcp.

MathiasKoch:
While I agree, my own use case is niche, this PR doesn't really address my use case in any particular way.
This addresses the fact that, as @Dirbaio said in the issue: "No matter what, a host should never transmit packets with a source IP that's not its own".

This could just as well happen from plugging an ethernet cable into a different network, which feels like something that should absolutely be handled in smoltcp, as the alternative is to emit garbage into the network?

The discussion on whether this is the right approach to handling this, or not, is the primary reason I opened this PR still with a "FIXME" comment in the code.

whitequark:
Thanks for elaborating on the context!

Dirbaio:
yeah I'm seeing the bug in a standard ethernet+dhcp network, when you plug into a different network, or when the DHCP IP changes. That's not niche :upside_down_face:

MathiasKoch:
How do we drive this any further?

whitequark:
I'm travelling this week but could take another look after.

MathiasKoch:
Any progress on this?

whitequark:
This PR is still in my queue. I'll look at it after I quit my current job at the end of Nov.

MathiasKoch:
Ping on this one :)

whitequark:
I'm back from my vacation so I'll try to look at this soon.

Dirbaio:
I'm somewhat concerned about doing the check on the interface for any egress. It'll apply to all socket types, and people probably want to be able to spoof src IPs with the raw socket. 

I've opened this alternative which does the check in the TCP socket itself: https://github.com/smoltcp-rs/smoltcp/pull/1113

PR #972, opened 2024-08-09
Enable support for DHCP option 15 (domain name)

korken89:
This makes `smoltcp` work in industrial networks with Windows DHCPs as they do not support option 119.

I'm not sure if the result should be directly tied into the `dns` socket, but the IPs for DNS servers are manually supplied so I follow the same pattern here and assume that users will build the FQDN before doing a DNS request.

Feedback welcome!


korken89:
I see clippy is a bit angry with the size of the domain name and the impact it has.
Should I move the max size into `gen_config.py` and set a default that is smaller than the maximum to make `clippy` happy?
E.g. a domain name size of 64 is probably enough for almost everything in my experience.

korken89:
I added it in my latest commit. Also realized I was doing wrong in `DhcpRepr` and moved to use a reference there instead of an owned `String`.

korken89:
This PR builds on top of https://github.com/smoltcp-rs/smoltcp/pull/974 to make `no_std` apps build

korken89:
Also, here is an example of it working from a network I'm testing this PR on

image

Dirbaio:
the original design of DHCP was to provide the raw bytes in `Config::packet`, so the user can parse the options they're interested in from there. The reason for this is I think adding explicit fields in Repr and Config for every single option a user might want is not going to scale, there's a lot of them.

korken89:
Hi @Dirbaio , I fully agree on the idea of having special options being handled outside as it is designed.
In the PR I was working on for `embassy-net` I started that path, but changed direction to add it here.

The motivation is that without domain name the DNS implementation is only working on non-FQDN networks or with manual FQDN lookups, and the only discrepancy was networks with domain name and automatically forming non-FQDN -> FQDN.

To me this warranted the direct inclusion in `smoltcp` instead of a special external handling as it's so fundamental to lookup.
The line-in-the-sand I formed for myself while working on this was that vendor extensions were candidates for direct inclusion in `smoltcp` and anything else should be handled in the aforementioned API.
If this is not the case I can make a PR to document where the line goes between `smoltcp` inclusion and not. :rocket:

korken89:
Another idea is to feature gate this so people do not pay for it unless they need it.
That would alleviate the scaling issue. What are your thoughts @Dirbaio ?

korken89:
I made it optional in my last commit just to show what I mean. I can drop the commit if it's considered as not needed.

Cargo.toml
```diff
@@ -60,6 +60,7 @@ defmt = ["dep:defmt", "heapless/defmt-03"]
 "proto-sixlowpan" = ["proto-ipv6"]
 "proto-sixlowpan-fragmentation" = ["proto-sixlowpan", "_proto-fragmentation"]
 "proto-dns" = []
+"proto-domainname" = []
```
astro, line 63: The feature's name could be confusing as "domainname protocol" sounds a lot like DNS. I suggest `proto-dhcpv4-domainname`.

astro (changes requested):
I am torn about feature-gating this. Low memory usage is important but we could end up with an unwieldy lot of feature flags when more DHCP options are added over time.

Cargo.toml
korken89: Good suggestion!

korken89:
I agree on your view. I'm open to a different approach!
This currently follows how other features are done, but before doing anything rash I'm awaiting review to see if it even has any chance of being accepted. :)

PR #979, opened 2024-08-23
fix raw_socket  on macos

hikaricai:
Currently BfpDevice::interface_mtu() return 4096, and RawSocket::new() would add it with EthernetFrame::header_len(), so BfpDevice::recv() would return  "Invalid argument (os error 22)".
This patch avoids adding mtu with ether_hdr_len for raw_socket on macos, maybe we will also avoid adding Ieee802154 extral hdr_len.

### DIFF
--- a/src/phy/raw_socket.rs
+++ b/src/phy/raw_socket.rs
@@ -43,7 +43,10 @@ impl RawSocket {
             mtu += 2;
         }
 
-        #[cfg(feature = "medium-ethernet")]
+        #[cfg(all(
+            feature = "medium-ethernet",
+            not(any(target_os = "macos", target_os = "openbsd"))
+        ))]
         if medium == Medium::Ethernet {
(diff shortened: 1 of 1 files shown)


whitequark:
I wonder if we should have a macOS test runner?

thvdveld:
smoltcp currently doesn't have any device implementation for macOS, I think (correct me if I'm wrong)? The TUN/TAP devices are only compiled for Linux or Android. Is the `BfpDevice` something we can add to smoltcp? Or could you share where I could find this implementation?

thvdveld:
@whitequark I made changes in my fork to add macOS test runners (https://github.com/thvdveld/smoltcp/pull/8), however, I'm not sure if it is really necessary. I don't think smoltcp contains code that is specific for macOS. As explained in my previous comment, smoltcp does not have a macOS specific device, such as the tuntap devices for Linux.

Let me know if I need to open a PR to add the macOS runners.

PR #983, opened 2024-09-10
Fixes #982

### LINKED ISSUE #982: Undefined behavior on 16 bit architectures
When I checked out this crate and used it as a path dependency, cargo warns be about several places with undefined behavior.

To reproduce:
```
git clone git@github.com:smoltcp-rs/smoltcp.git --branch v0.10.0
```
In a crate that uses a target with `"target-pointer-width": "16"` specify in `Cargo.toml` (assuming the directories are next to each other):
```
[dependency.smoltcp]
path= "../smoltcp"
```
When building/checking in the crate with the 16 bit target, cargo yields:
```
error: this arithmetic operation will overflow
   --> /home/dweber/gh/smoltcp/src/socket/tcp.rs:489:26
    |
489 |         if rx_capacity > (1 << 30) {
    |                          ^^^^^^^^^ attempt to shift left by `30_i32`, which would overflow
    |
    = note: `#[deny(arithmetic_overflow)]` on by default

error: this arithmetic operation will overflow
   --> /home/dweber/gh/smoltcp/src/socket/tcp.rs:597:13
    |
597 |             (1 << 16) - 1,
    |             ^^^^^^^^^ attempt to shift left by `16_i32`, which would overflow

error: this arithmetic operation will overflow
    --> /home/dweber/gh/smoltcp/src/socket/tcp.rs:2158:63
     |
2158 |                 repr.window_len = self.rx_buffer.window().min((1 << 16) - 1) as u16;
     |                                                               ^^^^^^^^^ attempt to shift left by `16_i32`, which would overflow
```

Note: if I specify the dependency as usual (`smoltcp= "0.10.0"`), these warnings won't appear but the behavior is still undefined.

d2weber:


src/socket/tcp.rs
```diff
@@ -1868,7 +1866,10 @@ impl<'a> Socket<'a> {
         let assembler_was_empty = self.assembler.is_empty();
 
         // Try adding payload octets to the assembler.
-        let Ok(contig_len) = self.assembler.add_then_remove_front(payload_offset, payload_len) else {
+        let Ok(contig_len) = self
+            .assembler
+            .add_then_remove_front(payload_offset, payload_len)
+        else {
```
d2weber: This was just formatting..

Dirbaio:
Fix looks good!

Could you add to CI a 16bit target that would've triggered the errors, so that CI catches similar busg in the future?

d2weber:
I could try to reproduce with the msp430-none-elf target (so we wouldn't need a separate target.json). But there is no std. Currently in `ci.sh test` all feature combinations are tested. One option would be maintaining a separate set of "no-std" feature combinations. Or I check if it'd be enough to run clippy with the msp430 target.

d2weber:
Just checked, it's reproducible with the `msp430-none-elf` target. But the standard library cannot be compiled with 16 bit pointer-width, so only no-std feature sets can be tested. I'd need help figuring out, which feature sets make sense here.
 
Then I could add something like this in [`ci.sh`](https://github.com/smoltcp-rs/smoltcp/blob/main/ci.sh)

```
clippy() {
    rustup +nightly component add rust-src
    cargo +nightly clippy -Z build-std=core,alloc --target msp430-none-elf --no-default-features --features="socket-tcp proto-ipv4 medium-ethernet"
}
```

By the way, I'm getting lots of warnings with `cargo +beta clippy`.

Dirbaio:
the errors are caught by rustc, not clippy, right? If so, just doing a `cargo build` for msp430 should be enough? 

For the features i'd just do one build, enabling all features that do work on nostd and ignoring the others.

d2weber:
Yes you are right, its `build`.

src/socket/tcp.rs
```diff
@@ -592,10 +593,7 @@ impl<'a> Socket<'a> {
     ///
     #[inline]
     fn scaled_window(&self) -> u16 {
-        cmp::min(
-            self.rx_buffer.window() >> self.remote_win_shift as usize,
-            (1 << 16) - 1,
-        ) as u16
+        u16::try_from(self.rx_buffer.window() >> self.remote_win_shift).unwrap_or(u16::MAX)
```
d2weber, line 596: I'm stuck here. Somethings off with ci. The newly added build fails exactly here where I already fixed an overflow:
```
error: this arithmetic operation will overflow
   --> src/socket/tcp.rs:699:67
    |
699 |         Some(cmp::min(last_win_adjusted >> self.remote_win_shift, (1 << 16) - 1) as u16)
    |                                                                   ^^^^^^^^^ attempt to shift left by `16_i32`, which would overflow
    |
```

d2weber:
I can't find the location where this error originates. Line numbers seem to be off. Maybe there is some macro magic going on? I also can't reproduce this error locally, no idea what's going on:
```
error: this arithmetic operation will overflow
   --> src/socket/tcp.rs:699:67
    |
699 |         Some(cmp::min(last_win_adjusted >> self.remote_win_shift, (1 << 16) - 1) as u16)
    |                                                                   ^^^^^^^^^ attempt to shift left by `16_i32`, which would overflow
    |
    = note: `#[deny(arithmetic_overflow)]` on by default
    ```

**Comment by thvdveld** [comment]
@d2weber I think `actions/checkout` was drunk and used an old commit. I opened a PR to update to the latest version of `actions/checkout`. You might want to rebase on main once that PR is merged.

File: .github/workflows/test.yml [hunk 3]
```diff
@@ -48,6 +48,14 @@ jobs:
       - name: Run Tests nightly
         run: ./ci.sh test nightly
 
+  test-build-16bit:
+    runs-on: ubuntu-22.04
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v2
```
thvdveld: Could you use `actions/checkout@v4`?

ci.sh
```diff
@@ -42,6 +42,10 @@ FEATURES_CHECK=(
     "defmt,alloc,medium-ip,medium-ethernet,proto-ipv6,proto-ipv6,proto-igmp,proto-dhcpv4,socket-raw,socket-udp,socket-tcp,socket-icmp,socket-dns,async"
 )
 
+FEATURES_NO_STD=(
+    "alloc,defmt,medium-ethernet,medium-ip,medium-ieee802154,proto-ipv4,proto-ipv4-fragmentation,proto-igmp,proto-dhcpv4,proto-ipv6,proto-rpl,proto-sixlowpan,proto-sixlowpan-fragmentation,proto-dns,socket,socket-raw,socket-udp,socket-tcp,socket-icmp,socket-dhcpv4,socket-dns,socket-mdns,packetmeta-id,async"
+)
+
```
thvdveld: Can you remove this? I don't think you are using it. Or you could add it to `FEATURES_CHECK`, which already contains some `no_std` tests.

d2weber: You are indeed right. I wanted to use it but then copy and pasted `FEATURES_CHECK`. The intention of `FEATURES_NO_STD` was to only build once with all features that work with no-std. But we can also use the existing  `FEATURES_CHECK`.

thvdveld:
~~I don't understand why CI is using the wrong source code. All jobs are checking out 19464653, which looks fine, but then when the tests are running, the source doesn't look like what's in 19464653.~~

Never mind, you actually have `(1 << 16) - 1` on line 699 in `src/socket/tcp.rs`.

d2weber:
> ~I don't understand why CI is using the wrong source code. All jobs are checking out [1946465](https://github.com/smoltcp-rs/smoltcp/commit/1946465366606b0913553da3c55dbf0da760e0b1), which looks fine, but then when the tests are running, the source doesn't look like what's in [1946465](https://github.com/smoltcp-rs/smoltcp/commit/1946465366606b0913553da3c55dbf0da760e0b1).~
> 
> Never mind, you actually have `(1 << 16) - 1` on line 699 in `src/socket/tcp.rs`.

No idea how I missed that. I have to check later, what went wrong, maybe my editor didn't reload the file after the rebase...

thvdveld:
I didn't catch it at first sight either! Thanks for fixing it.

PR #984, opened 2024-09-11
Use own source address for ARP and NDISC Solicitations

### LINKED ISSUE #940: ARP Request Packets generated with source address of triggering packet instead of interface
This is most relevant when using Raw sockets.

If you attempt to send a packet via a raw socket with a source address that does not match the interface address (i.e. a spoofed packet), smoltcp may generate an ARP request if it doesn't have the destination MAC address cached. This ARP request will have a source address matching the spoofed packet NOT the interface. As a result, smoltcp will ignore the ARP response.

The problem appears to be that `InterfaceInnner::lookup_hardware_addr()` uses the packet's source IPv4 address instead of looking up the interfaces's source IPv4 address at 989:
https://github.com/smoltcp-rs/smoltcp/blob/57b1dd4b158327c31edebd3260d90bf5c3b1452d/src/iface/interface/mod.rs#L977-L992

It seems likely that `self.get_source_address_ipv4(dst_addr)` should be used instead.

thvdveld:
When looking up the destination link-layer address for a given destination address, we need to send an ARP or ICMPv6 Neighbor Solicitation if the destination link-layer address is not in the cache.

When transmitting an ARP or ICMPv6 Neighbor Solicitation packet, use the source address of the interface instead of the source address of the packet that is tried to be sent.

By using the source address of the interface, we are sure that a response will be received on the same interface.

Fixes #940 .

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -898,7 +898,6 @@ impl InterfaceInner {
     fn lookup_hardware_addr<Tx>(
         &mut self,
         tx_token: Tx,
-        src_addr: &IpAddress,
         dst_addr: &IpAddress,
         fragmenter: &mut Fragmenter,
     ) -> Result<(HardwareAddress, Tx), DispatchError>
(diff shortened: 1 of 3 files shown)


whitequark (approved):
LGTM

PR #985, opened 2024-09-11
iface: do not require device and timestamp for multicast join/leave.

Dirbaio:
Instead of eagerly sending the join/leave packet when the user calls join/leave,
we update internal state and send the packet when the interface is polled.

Advantages:
- If the device is exhausted, the packet gets sent later instead of just failing and returning an error to the user.
- Makes the API consistent with everything else in smoltcp: operations only update internal state, poll is what sends/receives packets.
- Enables wrappers to offer simpler APIs with less generics. See https://github.com/embassy-rs/embassy/pull/3329 for an example, which is my original motivation.

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -82,6 +82,16 @@ pub struct Interface {
     fragmenter: Fragmenter,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum MulticastGroupState {
+    /// Joining group, we have to send the join packet.
+    Joining,
+    /// We've already sent the join packet, we have nothing to do.
+    Joined,
+    /// We want to leave the group, we have to send a leave packet.
+    Leaving,
+}
+
 /// The device independent part of an Ethernet network interface.
 ///
 /// Separating the device from the data required for processing and dispatching makes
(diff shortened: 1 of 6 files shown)


src/iface/interface/mod.rs
```diff
@@ -82,6 +82,16 @@ pub struct Interface {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum MulticastGroupState {
+    /// Joining group, we have to send the join packet.
+    Joining,
+    /// We've already sent the join packet, we have nothing to do.
+    Joined,
+    /// We want to leave the group, we have to send a leave packet.
+    Leaving,
+}
+
```
thvdveld, line 94: Shouldn't we move this to `interface/igmp.rs`?

Dirbaio: it doesn't work because `multicast_groups` exists if `any(feature = "proto-igmp", feature = "proto-ipv6")`, but `mod igmp` exists if `feature = "proto-igmp"` only. So it would break if you enable ipv6 but not igmp.

(fwiw I think these feature flags are a bit confusing. I'd like to refactor this so there's a single `multicast` Cargo feature that enables multicast for ipv4 and/or ipv6 depending on what you have enabled. Then rename `mod igmp` to `mod multicast` and move all multicast-related stuff there. I'd rather do this in a next PR tho)

thvdveld: I was thinking about the same thing.

Dirbaio:
(pushed small update, removing MulticastError::Exhausted that's unused now)

src/iface/interface/mod.rs
Dirbaio: got nerdsniped, just did it in #988

thvdveld (approved):

PR #987, opened 2024-09-12
Don't panic if no valid source address is found for the DNS query

### LINKED ISSUE #941: Panic when solving URI with mDNS
I tried to enable the following feature in embassy-net #[cfg(feature = "socket-mdns")] to be able to resolve this URI _homeassistant.local_ in rp pico. I was able to enable mdns feature but when trying to resolve the URI smoltcp panics with 
```
smoltcp-0.11.0/src/iface/interface/mod.rs:882:14:
called `Option::unwrap()` on a `None` value
```

thvdveld:
Don't panic if there is no valid source address for the DNS query. The state of the query is set to `Failed` instead, and the query is abandoned.

Fixes #941 .

### DIFF
--- a/src/socket/dns.rs
+++ b/src/socket/dns.rs
@@ -612,7 +612,15 @@ impl<'a> Socket<'a> {
                 };
 
                 let dst_addr = servers[pq.server_idx];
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
thanks!

PR #988, opened 2024-09-12
multicast: use single Cargo feature for both ipv4 and ipv6.

Dirbaio:
**Builds on top of #985**

multicast: use single Cargo feature for both ipv4 and ipv6.

The multicast feature was somewhat broken before, there was no way to do IPv6-only multicast
since a lot of multicast code was gated behind `proto-igmp` which force-enables ipv4.

Solution:

- Added feature `multicast` which enables multicast for ipv4 and/or ipv6, whatever the user has enabled.
- Removed feature `proto-igmp`. IGMP support in `wire` is always built if ipv4 is enabled. This causes no bloat if unused, and the increase in compile time is negligible. `iface` only uses it if the `multicast` feature is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,6 @@ defmt = ["dep:defmt", "heapless/defmt-03"]
 
 "proto-ipv4" = []
 "proto-ipv4-fragmentation" = ["proto-ipv4", "_proto-fragmentation"]
-"proto-igmp" = ["proto-ipv4"]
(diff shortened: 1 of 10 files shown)


Dirbaio:
ready to go now that https://github.com/smoltcp-rs/smoltcp/pull/985 is merged

whitequark (approved):
Looks okay at a glance

PR #991, opened 2024-09-16
iface: make poll() process all packets, add fine-grained poll functions.

### LINKED ISSUE #990: Throughput has decreased drastically after polling updates
According to @thvdveld and based on observations from @Dirbaio, it was noted that the throughput of Smoltcp has been significantly impacted by #954. Specifically, it appears that this is due to the requirement that egress is checked for every ingressed packet, where this wasn't previously the case.

It was also noted that the current return value of `bool` from `Interface::poll()` doesn't provide clear information to a user as to what they should do with that value. We likely want to provide some `Enum` type instead that is explicit in what the poll result is and what action the user should take. This would future-proof the network stack implementation.

Dirbaio:
This makes `.poll()` behave the same as before #954. Users affected by DoS
concerns can use the finer-grained egress-only and single-packet-ingress-only fns.

Fixes #990

- **wtf**
- **tests: use two queues for TestingDevice.**
- **iface: make poll() process all packets, add fine-grained poll functions.**


Dirbaio:
running the `benchmark` example, two attempts each:

```
smoltcp 0.11.0
    reader: 3.726 Gbps 3.676 Gbps
    writer: 11.102 Gbps 11.187 Gbps
smoltcp git main 469dccdba
    reader: 3.731 Gbps 3.788 Gbps
    writer: 4.260 Gbps 4.246 Gbps
smoltcp this PR:
    reader: 3.522 Gbps 3.558 Gbps
    writer: 11.524 Gbps 11.484 Gbps
```

Dirbaio:
embassy-net [stm32 eth perf test](https://github.com/embassy-rs/embassy/blob/main/tests/stm32/src/bin/eth.rs), stm32h755zi, 64k tcp buffer:

```
smoltcp 0.11.0:
    download:        6273 kB/s  6264 kB/s
    upload:          9619 kB/s  9619 kB/s
    upload+download: 1297 kB/s  1575 kB/s

smoltcp git main 469dccd
    download:        5357 kB/s  5354 kB/s
    upload:          9678 kB/s  9588 kB/s
    upload+download: 1098 kB/s  1122 kB/s

smoltcp this PR
    download:        5826 kB/s  5822 kB/s
    upload:          9142 kB/s  9145 kB/s
    upload+download: 3043 kB/s  2729 kB/s
```

thvdveld (approved):

PR #993, opened 2024-09-19
Export `PollResult`

lrh2000:
https://github.com/smoltcp-rs/smoltcp/pull/991 adds `PollResult` and `PollIngressSingleResult`. However, these two types are not accessible because they are in a private mod `interface`:
https://github.com/smoltcp-rs/smoltcp/blob/9cc06575ea03515fc4674a3e4f362688c5921ee3/src/iface/mod.rs#L8

This needs to be fixed, otherwise no one can check the result returned by `poll` or `poll_ingress_single`:
https://github.com/smoltcp-rs/smoltcp/blob/9cc06575ea03515fc4674a3e4f362688c5921ee3/src/iface/interface/mod.rs#L431-L436

### DIFF
--- a/src/iface/mod.rs
+++ b/src/iface/mod.rs
@@ -18,7 +18,9 @@ mod packet;
 
 #[cfg(feature = "multicast")]
 pub use self::interface::multicast::MulticastError;
-pub use self::interface::{Config, Interface, InterfaceInner as Context};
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
oops, nice catch. thanks!

PR #994, opened 2024-09-20
wire: use core::net types for IP addresses.

Dirbaio:
- **wire: move prefix_len impl down to v4 and v6 addrs.**
- **wire: use core::net::Ipv4Addr as the ipv4 address type.**
- **Bump MSRV to 1.80.**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "smoltcp"
 version = "0.11.0"
 edition = "2021"
-rust-version = "1.77"
+rust-version = "1.80"
 authors = ["whitequark <whitequark@whitequark.org>"]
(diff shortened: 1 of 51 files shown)


thvdveld:
I opened https://github.com/rust-lang/libs-team/issues/447, as requested in https://github.com/rust-lang/rust/pull/130629#issuecomment-2366979396.

Dirbaio:
this is ready to review.

I've "fixed" the `unstable_name_collisions` warnings by renaming the affected methods to `x_`. it's not pretty but it's the most reasonable solution I think. (note it doesn't affect the public API, the extension trait is private).

I still think we should merge this, moving to the stdlib's ip addr types is worth it even if it requires carrying a few warts internally.

thvdveld (approved):

PR #995, opened 2024-09-22
fix: internal sACK flag not set correctly for client socket

XOR-op:
### Bug Description
Selective ACK is used for TCP retransmission. In the current implementation, sACK option will be enabled by default for a client TCP socket. However, the `remote_has_sack` is not correctly set when the server sends a SYN-ACK packet back. Since `remote_has_sack` is false by default, client TCP sockets will not send sACK when packet loss happens. 

However, when the sACK-permitted option is set in the initial SYN packet, the server will (typically) only expect sACK for fast retransmission. Since no sACK is received, the server (as a sender) will wait until retransmission timer expires. This will trigger congestion control, and users will observe an inconsistent download speed even if the network link is relatively stable.

### Solution
Set `remote_has_sack` correctly in `SynSent` state. `Listen` state is not affected by this bug.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1806,6 +1806,7 @@ impl<'a> Socket<'a> {
                 self.remote_seq_no = repr.seq_number + 1;
                 self.remote_last_seq = self.local_seq_no + 1;
                 self.remote_last_ack = Some(repr.seq_number);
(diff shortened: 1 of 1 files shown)


whitequark (approved):

Dirbaio:
can you add a test?

XOR-op:
@Dirbaio Corresponding test has been added.

PR #996, opened 2024-09-27
fix: enter unreachable when discarding HBH Option with multicast destination address

thvdveld:
It was previously possible to enter an unreachable path when the HBH option type was 0b11 and the destination address was multicast. This is now fixed in the interface.

I also changed the way an HBH option type is parsed. Only the 2 highest order bits indicate what action must be taken, meaning that we can't have an unknown state.

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -288,12 +288,11 @@ impl InterfaceInner {
                         Ipv6OptionFailureType::DiscardSendAll => {
                             return HopByHopResponse::Discard(param_problem());
                         }
-                        Ipv6OptionFailureType::DiscardSendUnicast
-                            if !ipv6_repr.dst_addr.is_multicast() =>
-                        {
-                            return HopByHopResponse::Discard(param_problem());
+                        Ipv6OptionFailureType::DiscardSendUnicast => {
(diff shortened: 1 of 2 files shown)


jordens (approved):

src/iface/interface/ipv6.rs
```diff
@@ -288,12 +288,11 @@ impl InterfaceInner {
...
                             return HopByHopResponse::Discard(param_problem());
                         }
-                        Ipv6OptionFailureType::DiscardSendUnicast
-                            if !ipv6_repr.dst_addr.is_multicast() =>
-                        {
-                            return HopByHopResponse::Discard(param_problem());
+                        Ipv6OptionFailureType::DiscardSendUnicast => {
+                            if !ipv6_repr.dst_addr.is_multicast() {
+                                return HopByHopResponse::Discard(param_problem());
+                            }
```
Dirbaio, line 296: we still have to return here if the sender is multicast, to discard the packet, right?

thvdveld: Yes, indeed, I misread this in the RFC. I thought only the option needs to be discarded, but like you say, the full packet needs to be discarded, without sending an ICMP parameter problem back.

Dirbaio (approved):

PR #998, opened 2024-10-16
Improve logging on rejected packets

thomaseizinger:
I spent a bunch of time debugging why my packets weren't processed by `smoltcp`. After I finally found the code that performs these checks, I figured that these boolean conditions were quite hard to understand, so I split them up. This made it quite easy to add log statements that should guide the user to the problem as to why their packets aren't being processed.


crawford (commented):
Nice! This looks really helpful.

thvdveld:
Thank you, this is indeed helpful!

PR #1000, opened 2024-10-16
use core::net types for IpAddress

aurelj:
Similar to what was done for Ipv4Address and Ipv6Address.

- use core::net::IpAddr as the ip address type
- remote v4() and v6() constructor, use core IpAddr::V4(Ipv4Addr::new(_)) instead
- Remove IpVersion from public API

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -13,9 +13,9 @@ mod wire {
     extern crate test;
 
     #[cfg(feature = "proto-ipv6")]
-    const SRC_ADDR: IpAddress = IpAddress::Ipv6(Ipv6Address::new(0xfe80, 0, 0, 0, 0, 0, 0, 1));
+    const SRC_ADDR: IpAddress = IpAddress::V6(Ipv6Address::new(0xfe80, 0, 0, 0, 0, 0, 0, 1));
     #[cfg(feature = "proto-ipv6")]
-    const DST_ADDR: IpAddress = IpAddress::Ipv6(Ipv6Address::new(0xfe80, 0, 0, 0, 0, 0, 0, 2));
+    const DST_ADDR: IpAddress = IpAddress::V6(Ipv6Address::new(0xfe80, 0, 0, 0, 0, 0, 0, 2));
 
     #[cfg(all(not(feature = "proto-ipv6"), feature = "proto-ipv4"))]
     const SRC_ADDR: IpAddress = IpAddress::Ipv4(Ipv4Address::new(192, 168, 1, 1));
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -11,7 +11,7 @@ use smoltcp::iface::{Config, Interface, SocketSet};
 use smoltcp::phy::{wait as phy_wait, Device, Medium};
 use smoltcp::socket::tcp;
 use smoltcp::time::{Duration, Instant};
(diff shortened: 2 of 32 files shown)


Dirbaio:
I intentionally did not do this because it will cause extra RAM usage when you enable IPv4 only. We conditionally enable the enum cases so `IpAddress` is only 4 bytes if you enable IPv4 only, while `core::net::IpAddr` will be 16 bytes + the enum tag.

aurelj:
OK, I see, this is a fair point.

My vision is that embedded systems that use an IP stack are generally not the most memory constrained one, and that they usually won't handle more that one or two `IpAddress`, so I hardly see this as a practical issue. Also, I guess more and more systems are enabling IPv6 and thus won't save the 12 bytes or RAM anyway.

I'm not sure what is the policy in smoltcp, but if it is to try to squeeze up to the last bit than can be squeezed, then I agree this PR is a bad idea.

Dirbaio:
there's also the question of what happens when you pass an IP addr to the stack that is of a version that is not enabled. Before this was imposssible, now it'll need to either return an error or panic.

crawford:
> My vision is that embedded systems that use an IP stack are generally not the most memory constrained one

My system has 4kiB of RAM, so I’d like just about every byte I can find.

aurelj:
> > My vision is that embedded systems that use an IP stack are generally not the most memory constrained one
> 
> My system has 4kiB of RAM, so I’d like just about every byte I can find.

Wow, a 4kiB MCU with an IP stack ! That's seriously impressive ! Curious what kind of MCU it is ? And does it have an internal ethernet MAC device or do you use an external ethernet MAC ?

Anyway, I agree that this PR is not desirable for such use case.

crawford:
Hang on. I’m an idiot. I’m mixing up my projects. This latest one uses the EFM32GG11, which has 512 KiB. On top of that, I see I only allocate space for a single address, so that 12-byte overhead doesn’t seem that bad (for my configuration).

I retract my statement. :)

thvdveld:
> @Dirbaio there's also the question of what happens when you pass an IP addr to the stack that is of a version that is not enabled. Before this was imposssible, now it'll need to either return an error or panic.

I don't think we have many places where this occurs? So it might not be such a big problem?

Dirbaio:
It occurs in every function in the public API that takes an IpAddress. 

For example, what happens when smoltcp is configured to ipv4-only and you try to set a v6 address in an interface, or try to connect a smoltcp socket to a v6 address?

Frostie314159:
We could introduce a trait `ToIpAddress`, which we only implement for either `Ipv4Addr` or `Ipv6Addr`, depending on which feature is enabled. If both are enabled, we could also implement it for `IpAddr.`

Dirbaio:
I'm going to close this, since it really seems there's no solution for this other than accepting wasting RAM, which I'd rather not do.

aside: `trait ToIpAddress` doesn't fix the issue:
- For functions where the user passes an IP addr to smoltcp, doing `fn foo(T: impl IntoIpAddress)` will cause it to get monomorphized if you ever call it with multiple IP types, increasing code size.
- It doesn't do anything for the case where smoltcp passes an IP back to the user. There we still need a type that's "either ipv4, or ipv6, or both" depending on feature flags.

PR #1001, opened 2024-10-19
Support TCP simultaneous open

lrh2000:
[RFC 9293 states:](https://datatracker.ietf.org/doc/html/rfc9293#section-3.5)
> A TCP implementation MUST support simultaneous open attempts (MUST-10).

It also details the related state machine ([Figure 7: Simultaneous Connection Synchronization](https://datatracker.ietf.org/doc/html/rfc9293#figure-7)):
```
TCP Peer A                                       TCP Peer B

1.  CLOSED                                           CLOSED

2.  SYN-SENT     --> <SEQ=100><CTL=SYN>              ...

3.  SYN-RECEIVED <-- <SEQ=300><CTL=SYN>              <-- SYN-SENT

4.               ... <SEQ=100><CTL=SYN>              --> SYN-RECEIVED

5.  SYN-RECEIVED --> <SEQ=100><ACK=301><CTL=SYN,ACK> ...

6.  ESTABLISHED  <-- <SEQ=300><ACK=101><CTL=SYN,ACK> <-- SYN-RECEIVED

7.               ... <SEQ=100><ACK=301><CTL=SYN,ACK> --> ESTABLISHED
```

In short:
 1. If a TCP socket is in SYN-SENT state and receives a SYN packet (_not_ a SYN|ACK packet), it should change its state to SYN-RECEIVED. **This is not previously implemented, and such packets are silently dropped. This PR implements the above behavior.**
 2. If it then receives an ACK packet acknowledging the SYN packet, it should change its state to ESTABLISHED. **This is already implemented.**

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1508,11 +1508,6 @@ impl<'a> Socket<'a> {
             (State::Listen, _, None) => (),
             // This case is handled in `accepts()`.
             (State::Listen, _, Some(_)) => unreachable!(),
-            // Every packet after the initial SYN must be an acknowledgement.
-            (_, _, None) => {
-                net_debug!("expecting an ACK");
-                return None;
-            }
             // SYN|ACK in the SYN-SENT state must have the exact ACK number.
             (State::SynSent, TcpControl::Syn, Some(ack_number)) => {
                 if ack_number != self.local_seq_no + 1 {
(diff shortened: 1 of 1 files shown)


whitequark (approved):

whitequark:
I remember reading about simultaneous open in the original TCP RFC and thinking that it's a pretty silly thing that I've never seen used.

lrh2000:
> I remember reading about simultaneous open in the original TCP RFC and thinking that it's a pretty silly thing that I've never seen used.

Yeah, agree. It could be useful for [TCP hole-punching](https://en.wikipedia.org/wiki/TCP_hole_punching#Details_of_a_typical_TCP_connection_instantiation_with_TCP_Hole_Punching), but it is still a very limited scenario. However, per the RFC, this is required to be supported...

By the way, thanks for your review!

PR #1008, opened 2024-11-02
Fix the error of specific length IP packets not being fragmented

Jc0x7D3:
There is an obvious mistake at `src/iface/interface/mod.rs:1204`:  `total_ip_len` represents the length of the IP packet, while `self.caps.max_transmission_unit` is the max length of link-layer packets. The correct comparison should be with self.caps.ip_mtu(). 

When using Medium Ethernet, since `self.caps.max_transmission_unit` is 14 octets larger than `self.caps.ip_mtu()`, IP packets with length of 1501~1514 octets should be fragmented, but mistakenly entered the non-fragment branch.

Additionally, I added a test case sending IPv4 packets of different lengths, which can trigger this error.


Dirbaio (approved):
thanks!

PR #1009, opened 2024-11-08
feat: Add MLDv2 query response support

bergzand:
This add support for responding to MLDv2 Listener Queries. Both general queries and multicast specific queries are supported. The implementation tries to match the igmpv2 implementation where the spec allows it.

- Querier's robustness variable is not handled
- Multicast address and source specific queries are treated as regular multicast address specific queries.
- Maximum Response Delays [above 32768](https://www.rfc-editor.org/rfc/rfc3810.html#section-5.1.3) are not calculated to spec.
- Delay between the query and the report is not random, but a simple fixed delay proportional to the Maximum Response Code

I've tested this via the multicast6 example with a Linux bridge device configured with `multicast_querier` enabled. Multicast traffic outside the subscriptions (such as neighbor solicitations for other solicited node multicast addresses) are filtered at the bridge, and subscribed addresses are forwarded to the tap device as expected. 
Unfortunately this also means only the general listener queries are tested, I'm not sure how to test multicast address specific queries at this point.

### Open questions

- Any automated tests that can and should be added to this pr?
- @thvdveld Is there anything in this code that needs some exception on ieee802.15.4 links?

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -422,6 +422,16 @@ impl InterfaceInner {
                 #[cfg(feature = "medium-ip")]
                 Medium::Ip => None,
             },
+            #[cfg(feature = "multicast")]
+            Icmpv6Repr::Mld(repr) => match repr {
+                // [RFC 3810 § 6.2], reception checks
+                MldRepr::Query { .. }
+                    if ip_repr.hop_limit == 1 && ip_repr.src_addr.is_link_local() =>
+                {
+                    self.process_mldv2(ip_repr, repr)
+                }
+                _ => None,
+            },
 
             // Don't report an error if a packet with unknown type
             // has been handled by an ICMP socket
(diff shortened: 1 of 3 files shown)


thvdveld:
Thank you for adding this!

> Delay between the query and the report is not random, but a simple fixed delay proportional to the Maximum Response Code.

`InterfaceInner` holds a random number generator which you can use I think.

> Any automated tests that can and should be added to this pr?

Maybe you could add something in `src/iface/interface/tests/ipv6.rs`? You could check if `process_mldv2` reacts correctly to an incoming packet? 

> Is there anything in this code that needs some exception on ieee802.15.4 links?

I don't think so. It should behave similarly. Unless I missed an RFC.

bergzand:
> `InterfaceInner` holds a random number generator which you can use I think.

That makes things simple, using that for the random delay now

> Maybe you could add something in `src/iface/interface/tests/ipv6.rs`? You could check if `process_mldv2` reacts correctly to an incoming packet?

added a new test to that file. I think it could be merged into the existing multicast test function, but I'll leave the decision whether that makes sense to a reviewer.

Dirbaio (approved):

PR #1012, opened 2024-11-14
 feat: Automatically join solicited-node multicast addresses 

bergzand:
IPv6 over Ethernet should join the solicited-node multicast
addresses required by the configured IPv6 addresses, as neighbor
solicitations for these addresses have the solicited-node multicast
address as destination address.

This commit automatically leaves old solicited-node multicast addresses
and joins the new set when the IP addresses on the interface are
updated.

### DIFF
--- a/src/iface/interface/multicast.rs
+++ b/src/iface/interface/multicast.rs
@@ -164,15 +164,8 @@ impl Interface {
             .multicast
             .groups
             .keys()
-            .filter_map(|group_addr| match group_addr {
-                IpAddress::Ipv6(address)
-                    if address.is_solicited_node_multicast()
-                        && !self.inner.has_solicited_node(*address) =>
-                {
-                    Some(*group_addr)
-                }
-                _ => None,
-            })
+            .cloned()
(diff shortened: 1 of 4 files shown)


bergzand:
One effect of this pr is that an ethernet-based interface wants to send out multicast reports as soon as an IPv6 address is configured. Not sure if there's strong opinions on that.
According to [rfc 6775 § 5.2 ](https://datatracker.ietf.org/doc/html/rfc6775#section-5.2) joining the solicited-node multicast address is not required on IEEE 802.15.4.

src/iface/interface/multicast.rs
```diff
@@ -156,6 +156,36 @@ impl Interface {
...
+        // Remove old solicited-node multicast addresses
+        let removals: Vec<_, IFACE_MAX_MULTICAST_GROUP_COUNT> = self
+            .inner
+            .multicast
+            .groups
+            .keys()
+            .filter_map(|group_addr| match group_addr {
+                IpAddress::Ipv6(address)
+                    if address.is_solicited_node_multicast()
+                        && self.inner.has_solicited_node(*address) =>
```
Dirbaio: should be `&& !has_solicited_node(...)` . fixed it for you.

Dirbaio (approved):

Issue #942, opened 2024-06-25
Sending custom ETH packets


maxgr0:
### Description
We have the requirement to send/receive ethernet frames with a custom type/protocol for [HomePlug AV](https://www.slideshare.net/slideshow/sstic-2019-v2g-injector-whispering-to-cars-and-charging-units-through-the-powerline/149677087#20) (`0x88e1`). Using a raw socket does not really work as it drops packets which are neither IPv4 or IPv6. 
At the moment, we use smoltcp v0.11.0 through [embassy-net](https://github.com/embassy-rs/embassy/blob/main/embassy-net/README.md) in an embedded device.

In my view, there are multiple possible solutions to the issue:

#### 1. Directly talk to the ETH device
This would work in theory and we already verified that this works. The problem here is that we also need to expose an IPv6 socket for follow-up communication (after the HomePlug matching procedure is done) and its a way better solution to use this amazing library instead of crafting something ourselves.

Another possible combination is that we first talk directly with the ETH device and then spin up our IPv6 socket using smoltcp. This actually runs into timing issues as we then might receive some packets on IPv6 before the IPv6 socket is created.

#### 2. Refactor the `RawSocket`
In order to make this raw socket accept non-IP packets, it would require some refactoring in the following parts:

Make `ip_version` and `ip_protocol` both an `Option`
https://github.com/smoltcp-rs/smoltcp/blob/57b1dd4b158327c31edebd3260d90bf5c3b1452d/src/socket/raw.rs#L87-L88

Update the `process` and `dispatch` methods to check if we have version/protocol defined
https://github.com/smoltcp-rs/smoltcp/blob/57b1dd4b158327c31edebd3260d90bf5c3b1452d/src/socket/raw.rs#L341

https://github.com/smoltcp-rs/smoltcp/blob/57b1dd4b158327c31edebd3260d90bf5c3b1452d/src/socket/raw.rs#L370

#### 3. Create a `HpgpSocket`
This would mean to create an entire new socket with all validation and features to be fully compliant with the HomePlug AV protocol. From the number of changes and complexity/effort most probably similar to https://github.com/smoltcp-rs/smoltcp/pull/902 so also a lot of effort for the maintainers to review as its kinda specific.

#### 4. Create a `CustomSocket`
This is similar to the third option above but might not require so much specific knowledge for reviewing as the user could just specify e.g. an `ether_type` and the routing will be based on this while putting the validation and "handling" of packets to the user.

### Summary
[...]

PR #1013, opened 2024-11-24
Add new socket type for raw ethernet protocols

ususdei:
Hi everyone!

I'm currently working on a rust library for a protocol which runs on plain ethernet using a custom ethertype (TECMP on 0x99FE).
It also uses an additional IP-based protocol for discovery and configuration.

I would also like to use this library on some STM32, so I tried to port it from native linux sockets to the smoltcp stack.

Thus I started to implement some basic support for raw ethernet sockets.
This is still a _work in progress_ but the basic functionality already works on an STM32F7.
It is still missing support for Metadata and the send API could be more convenient
but I would like to get this out there and gather some early feedback as to whether you would even consider merging something like this.

So just some initial questions:

- Is this feature even in scope for the smoltcp project?
- Do you agree with my overall design?
- Do you already have any other initial feedback?

Thank you all for your feedback and your great contribution to the embedded rust ecosystem!


nikvoid:
I missed this PR and made [similar](https://github.com/nikvoid/smoltcp/tree/ether-socket) thing, so... +1 for raw ethernet sockets

i404788:
We nearly did the same as @nikvoid (hey btw), +1. Also should link https://github.com/smoltcp-rs/smoltcp/issues/942 to help the next person

not-Ryan:
Something you might want to add the `socket-eth` feature to is the compile error on [lib.rs:122](https://github.com/ususdei/smoltcp/blob/1146477db4aa3e34174554193f0ec34fe4839039/src/lib.rs#L101-L112)

Right now I'm getting the compile error:
```rust
compile_error!("If you enable the socket feature, you must enable at least one of the following features: socket-raw, socket-udp, socket-tcp, socket-icmp, socket-dhcpv4, socket-dns");
```

ususdei:
It's cool to hear that others would like to use this lib in similar ways as we do.

But I'm not sure if it's really worth investing more time into polishing this PR.
There hasn't been any feedback to this in almost a year now.

thvdveld:
@ususdei sorry it took so long before the first reply. I think this PR looks good. Could you, as @not-Ryan suggested, edit the compile error as well? I don't have any other remarks.

@Dirbaio @whitequark any remarks? I think this feature was requested already in the issues (#942, #902 could be implemented as a `Eth` socket).

src/socket/eth.rs
```diff
@@ -0,0 +1,525 @@
...
+}
+
+impl core::fmt::Display for EthMetadata {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        #[cfg(feature = "packetmeta-id")]
+        return write!(f, "PacketID: {:?}", self.meta);
+    }
+}
+
+/// A Eth packet metadata.
```
whitequark:
```suggestion
/// An Ethernet packet metadata.
```

src/socket/eth.rs
```diff
@@ -0,0 +1,525 @@
...
+pub type PacketMetadata = crate::storage::PacketMetadata<EthMetadata>;
+
+/// A Eth packet ring buffer.
```
whitequark:
```suggestion
/// An Ethernet packet ring buffer.
```

whitequark:
@ususdei Apologies for the lack of feedback. I don't see any reason not to merge this; it fits the overall theme of the library and doesn't add any overhead when disabled.

The implementation is basically copied from raw sockets, right? That's also an easy thing to approve.

not-Ryan:
Since your force-push of [833d319](https://github.com/smoltcp-rs/smoltcp/commit/833d31946e27e3c6cdcb808f5eeffb6cf66cabe7) it seems to be broken at:

https://github.com/ususdei/smoltcp/commit/833d31946e27e3c6cdcb808f5eeffb6cf66cabe7#diff-57320a1d1f958a93a7d782cab2e9e6409dd32f50a98a86e854e88b28559de4b0R62-R65

This is only when you disable the default features:
```
  --> .../smoltcp-426d9b36010b4974/833d319/src/socket/eth.rs:62:56
   |
62 |     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
   |        ---                                             ^^^^^^^^^^^^^^^^^ expected `Result<(), Error>`, found `()`
   |        |
   |        implicitly returns `()` as its body has no tail or `return` expression
   |
   = note:   expected enum `core::result::Result<(), core::fmt::Error>`
           found unit type `()`
```

ususdei:
Thanks @not-Ryan !
I found some more feature combinations which don't work yet.
I am working on them while also improving the test coverage.

ususdei:
I just finished some remaining TODOs from my list and reran the code on my test-device.

So from my side, this MR is now ready for a thorough review.
Let me know if you are unhappy with anything.

ususdei:
> The implementation is basically copied from raw sockets, right? That's also an easy thing to approve.

Exactly. I didn't see a way to extend raw sockets to also handle plain ethernet frames without breaking the
API, type safety and performance. So eth-sockets are now almost a copy of raw sockets, just on a different layer.

And then copied over metadata support from UDP sockets (which is still missing from raw sockets but I need the metadata-id for accurate timestamps).

whitequark (commented):
Can you explain the changes to ARP logic you've made?

ususdei:
> Can you explain the changes to ARP logic you've made?

None, I hope...

Since the ethernet packet handling cannot generate any automatic reply, we can still process ARP packages using process_arp() as we did before (same for all IP packets).
For any other ethertype, we can ignore ARP completely.

Or did I miss something here?

Dirbaio:
i'm going to close this for the same reasons as https://github.com/smoltcp-rs/smoltcp/pull/1121#issuecomment-3874638487 ie it'd be better to add ethernet support to the existing raw socket but it requires refactoring a bit.

ususdei:
Hi @Dirbaio,

I can totally get behind the decision to trade API compatibility (and maybe a bit of performance) for significantly less code duplication as outlined in https://github.com/smoltcp-rs/smoltcp/pull/1121#issuecomment-3874638487. That's a reasonable choice.

That said, I can’t help but wonder whether this was really something that couldn’t have been communicated more than a year ago...

PR #1015, opened 2024-11-28
fix(tcpdump): partial checksum

thvdveld:
In some cases, the checksum calculation is offloaded to hardware and only the partial checksum is calculated (only the pseudo-header). The tcpdump example was not taking this into account and was flagging some packets with incorrect checksums.

### DIFF
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -779,9 +779,17 @@ pub mod checksum {
     }
 
     // We use this in pretty printer implementations.
-    pub(crate) fn format_checksum(f: &mut fmt::Formatter, correct: bool) -> fmt::Result {
+    pub(crate) fn format_checksum(
+        f: &mut fmt::Formatter,
+        correct: bool,
+        partially_correct: bool,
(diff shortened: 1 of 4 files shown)


src/wire/ip.rs
```diff
@@ -833,7 +833,10 @@ pub fn pretty_print_ip_payload<T: Into<Repr>>(
                             )?;
                             let valid =
                                 udp_packet.verify_checksum(&repr.src_addr(), &repr.dst_addr());
-                            format_checksum(f, valid)
+                            let partially_valid = udp_packet
+                                .verify_partial_checksum(&repr.src_addr(), &repr.dst_addr());
+
+                            format_checksum(f, valid || partially_valid)
```
whitequark: I'd prefer different formatting for valid vs partially valid checksum, since this is a diagnostic interface.

thvdveld: I modified the message when the partial checksum is correct.

whitequark (approved):

PR #1020, opened 2024-12-23
TCP exponential backoff fix

tomDev5:
Currently, after a TCP retransmit, the socket timer was set to idle, which means that no actions except for keep-alive are required.
This is incorrect in the case where another TCP retransmission will be required.
This PR adds a test and fixes the problem.

And just to make sure I didn't miss anything - what if `set_for_retransmit` sets a timeout larger than the keep alive?
I assume that's OK because the socket won't close with unacknowledged data?

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -344,9 +344,10 @@ impl Timer {
                 }
             }
             Timer::Retransmit { expires_at, delay } if timestamp >= expires_at => {
+                let delay = delay * 2;
                 *self = Timer::Retransmit {
                     expires_at: timestamp + delay,
-                    delay: delay * 2,
+                    delay,
                 }
             }
             Timer::Retransmit { .. } => (),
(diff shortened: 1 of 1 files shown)


src/socket/tcp.rs
```diff
@@ -2285,11 +2286,9 @@ impl<'a> Socket<'a> {
...
                 self.remote_last_seq = self.local_seq_no;
 
-                // Clear the `should_retransmit` state. If we can't retransmit right
-                // now for whatever reason (like zero window), this avoids an
-                // infinite polling loop where `poll_at` returns `Now` but `dispatch`
-                // can't actually do anything.
-                self.timer.set_for_idle(cx.now(), self.keep_alive);
+                // Set next time of retransmission. `set_for_retransmit` will exponentially increase it.
+                self.timer
+                    .set_for_retransmit(cx.now(), self.rtte.retransmission_timeout());
```
Dirbaio, line 2291: i'm not sure if restarting the retransmit timer here is 100% correct, because it's not yet guaranteed we're going to send a packet. Actually transmitting is done further below, which might fail because there's not enough buffers in the device for example. 

i've opened  #1023 which does this a slightly different way: it still does `set_for_idle` here so we only restart the retransmit timer if we did actually send a retransmission packet (because it's done after sending), and keeps the `delay` variable outside the timer (in the RTTE struct) so it doesn't get cleared.

i've included your test from this PR in it.

PR #1025, opened 2024-12-30
Several TCP fixes, improvements.

Dirbaio:
- **tcp: set ack_all when remote acks bytes sent in earlier retransmissions.**
- **tcp/tests: make sure line number in recv_nothing points to the test code.**
- **tcp: start retransmit on first data sent, fix FIN retransmit in CLOSING.**
- **tcp: add zero window probe support.**


Dirbaio:
merged as part of #1026

PR #1027, opened 2025-01-03
Fix RawHardwareAddress panic when parsing to specific link layer address

thvdveld:
The `RawHardwareAddress::parse()` method panics if the length of the address is invalid. More specific, for an Ethernet address, the length should exactly be 6 bytes, and for an IEEE 802.15.4 address, the length should exactly be 8 bytes. Previously, we only checked if the size of the input was at least the size of the link layer address. Since `Ethernet::from_bytes()` does a copy_from_slice, the length of the input should be exactly 6 bytes. A panic can be triggered when the length is for example 7 bytes, while the `medium-ethernet` and `medium-ieee802154` features are enabled. This commit fixes the panic by checking if the length of the input is exactly the size of the link layer address.

This panic was discovered by people from Radically Open Security.

### DIFF
--- a/src/wire/mod.rs
+++ b/src/wire/mod.rs
@@ -478,6 +478,10 @@ pub struct RawHardwareAddress {
 
 #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
 impl RawHardwareAddress {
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Good catch!

PR #1029, opened 2025-01-03
Fix DHCPv4 panic when T1 < T2 < lease duration is not respected

### LINKED ISSUE #1028: DHCPv4 panics when lease duration < renew duration
When processing an ACK in the DHCP socket, we calculate the renew duration (T1) and rebind duration (T2). These are provided as an option. Ideally, the server SHOULD provide both of them, with the following constraint: T1 < T2 < lease duration. When only T1 is provided, [we calculate T2 as T1 + (lease duration - T1) * 0.75](https://github.com/smoltcp-rs/smoltcp/blob/3a30d0c891f79c9f46dfcda6f5532fd370bd7d74/src/socket/dhcpv4.rs#L518-L525). However, this panics when the lease duration is smaller than T1. This is because the implementation of the subtraction of two durations uses `checked_sub` and `expect()`.

I'm not sure what the correct approach is to fix this, as I'm not familiar with DHCPv4. I think we should use default values for T1 and T2 when the provided values of T1 and T2 do not follow T1 < T2 < lease duration.

From [RFC 2131](https://datatracker.ietf.org/doc/html/rfc2131#section-4.4.5): 
> T1 MUST be earlier than T2, which, in turn, MUST be earlier than the time at
> which the client's lease will expire.

Credits to the people from Radically Open Security for finding this panic.

thvdveld:
This is an attempt to try fixing #1028. I'm not sure if this is the correct approach. When invalid values for T1 and T2 are received, default values are calculated instead. Maybe the ACK should be ignored instead of using default values?

If using default values is the approach we want, then this PR should still add tests.

Fixes #1028

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -502,6 +502,8 @@ impl<'a> Socket<'a> {
         // Times T1 and T2 are configurable by the server through
         // options. T1 defaults to (0.5 * duration_of_lease). T2
         // defaults to (0.875 * duration_of_lease).
+        // When receiving T1 and T2, they must be in the order:
+        // T1 < T2 < lease_duration
         let (renew_duration, rebind_duration) = match (
             dhcp_repr
                 .renew_duration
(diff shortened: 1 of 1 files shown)


whitequark (commented):
Seems reasonable to me. There isn't much more we can do against a malicious or broken DHCP server.

whitequark (approved):

whitequark:
> Maybe the ACK should be ignored instead of using default values?

We could do that, but (assuming a genuinely broken server) debugging IP assignment failures is always really unpleasant, so accepting it feels like the right choice.

Dirbaio:
yeah, accepting it will most likely result in everything working fine (DHCP servers usually tolerate clients renewing/rebinding earlier/later than expected), while ignoring it will most likely result in things not working at all. Retrying likely won't fix it since the server is probably consistently misconfigured/buggy.

whitequark:
Oh, merged with no tests...

PR #1032, opened 2025-01-14
Remove some redundant code

ususdei:
Just some automatically applied suggestions from clippy 0.1.84.


whitequark:
I think this change doesn't really provide value, only adds churn.

ususdei:
It doesn't add much by itself, agreed.
Maybe just someone else would get annoyed by the warnings.
Other than that I just think, a clean codebase with fewer distractions makes development easier.
But it is going to converge anyway. I just have to wait until MSRV hits 1.84. ;-)

Issue #1031, opened 2025-01-12
6LoWPAN over BLE (RFC7668)


Dominaezzz:
I'm looking at implementing this with an ESP32 and [trouble](https://github.com/embassy-rs/trouble).

From reading various RFCs it seems to be a somewhat straightforward affair.
One just needs a L2CAP channel and a minimal GATT service for advertising, which trouble supports.

The network forms a star topology where the "Central" is the router and the "Peripheral"s are nodes in the network.
All communications between nodes/peripherals happen via the router/central.
Central is responsible for connecting to nodes advertising themselves as nodes.
Each node (including the router) has a link-local IPv6 address derived from the BLE mac address, but can still register additional non-link-local addresses using standard IPv6 protocols like Neighbour Solicitation (NS).

The L2CAP channel takes care throttling (with the credits system) and disassembly/re-assembly.

The IPv6 headers sent over BLE are compressed in the same format used for IEEE 802.15.4, but the conditions for when to compress are different.

Multicast is an O(n) operation (where n is the number of nodes) rather than a O(1) operation, due to the star topology.

I've mostly gotten the mac logic down, I just need a networking stack to handle the networking.

I looked into implementing the [Device](https://docs.rs/smoltcp/latest/smoltcp/phy/trait.Device.html) trait and I'm trying to figure out which medium to use. Do I need to PR a BLE medium or can I just use the Ip medium.

I'm still trying to figure out if I need to modify smoltcp to work with BLE or if I can do it all behind the `Device` trait.

Dominaezzz:
From discussions over Matrix with @Dirbaio it appears I just need a new medium for 6LowPan without ieee802154 headers, this can then be shipped over the L2CAP channel.

I just need to figure out if the IPv6 header compression is similar enough to just work on BLE as well

Dominaezzz:
- [ ] The [IPHC wire file](https://github.com/smoltcp-rs/smoltcp/blob/main/src/wire/sixlowpan/iphc.rs#L9) needs to have `ieee802154` removed from it. Not sure what the protocol-agnostic replacement is yet, maybe it could be generic. Or a EUI64 perhaps.

thvdveld:
[...]

PR #1033, opened 2025-01-15
Remove ieee802154 from IphcRepr

Dominaezzz:
First part of #1031 .

The `IphcRepr` type was written in terms of ieee802154 addresses.
I've replaced the ieee802154 address for structures that closer match the RFC and can be used on other mediums like BLE.

After this, I need to do the same for `sixlowpan/mod.rs` and move the ieee802154 bits into the `wire/ieee802154.rs` file.

### DIFF
--- a/fuzz/fuzz_targets/sixlowpan_packet.rs
+++ b/fuzz/fuzz_targets/sixlowpan_packet.rs
@@ -39,12 +39,10 @@ fuzz_target!(|fuzz: SixlowpanPacketFuzzer| {
         }
         Ok(SixlowpanPacket::IphcHeader) => {
             if let Ok(frame) = SixlowpanIphcPacket::new_checked(fuzz.data) {
-                if let Ok(iphc_repr) = SixlowpanIphcRepr::parse(
-                    &frame,
-                    fuzz.ll_src_addr.map(Into::into),
-                    fuzz.ll_dst_addr.map(Into::into),
-                    &[],
-                ) {
+                if let Ok(iphc_repr) = SixlowpanIphcRepr::parse(&frame) {
+                    let src_addr = iphc_repr.src_addr.resolve(fuzz.ll_src_addr.map(Into::into), &[], iphc_repr.context_identifier).unwrap();
+                    let dst_addr = iphc_repr.dst_addr.resolve(fuzz.ll_dst_addr.map(Into::into), &[], iphc_repr.context_identifier).unwrap();
+
                     let mut buffer = vec![0; iphc_repr.buffer_len()];
                     let mut iphc_frame = SixlowpanIphcPacket::new_unchecked(&mut buffer[..]);
                     iphc_repr.emit(&mut iphc_frame);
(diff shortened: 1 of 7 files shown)


Dominaezzz:
Was hoping to get some feedback on the structure of things before investing time into improving coverage.

Dirbaio:
Could you explain why is this needed to support 6lowpan over BLE? 

IIUC address compression works the same way in both 6lo over BLE and 6lo over ieee802.15.4. The info you use to compress/decompress is an EUI64 that can be derived from either the ieee802.15.4 link address or the BLE link address. So I think it should be possible to pass in the EUI64, and have higher layers calculate it differently depending on whether the link is ieee802.15.4 or BLE?  Then the code can stay in the same place, the refactor is not needed.

I'm not a 6lowpan expert though. Hopefully @thvdveld can weigh in.

(The reason I'm a bit concerned with the refactor is that "look at packet fields, write Rust enum, copy it around, match it to do thing" typically increases code size vs "look at packet fields, do thing")

Dominaezzz:
> The info you use to compress/decompress is an EUI64 that can be derived from either the ieee802.15.4 link address or the BLE link address.

Yes, this is largely the case for non-contextual compression.
The problem with the existing `IphcRepr` is, it has ieee802.15.4 link addresses hard coded into it, which cannot be used by BLE. BLE uses mac addresses which are a different length (48 bits) to ieee802.15.4 link addresses (64 bits).
This PR just makes the type more agnostic.

> So I think it should be possible to pass in the EUI64, and have higher layers calculate it differently depending on whether the link is ieee802.15.4 or BLE? Then the code can stay in the same place, the refactor is not needed.

Due to the things I've mentioned above, some kind of refactor will still be needed to remove ieee802.15.4 from `IphcRepr` to allow it to be used for BLE.
See [this](https://github.com/smoltcp-rs/smoltcp/blob/49cc60a368ba2d726d5dbec537261f178adf0206/src/wire/sixlowpan/iphc.rs#L684).

There's more info in the linked issue.

Dirbaio:
My question is: why can't we keep the current code structure, just replace `LlAddress` with an EUI64? Higher layers derive this EUI64 from either the ieee802154 or BLE address. 

This way `wire::sixlowpan` is agnostic from ieee802154 *and* we keep the current code structure which is simpler and uses less flash. No need for the enums you added.

Dominaezzz:
I unfortunately don't know enough about ieee802154 to give you a good answer. @thvdveld [thinks an EUI64 isn't enough](https://github.com/smoltcp-rs/smoltcp/issues/1031#issuecomment-2586544771) for ieee802154.

thvdveld:
I had a look at [RFC6282](https://datatracker.ietf.org/doc/html/rfc6282) for the address compression, and I think working just with EUI64 will work (sorry for not checking that before). We still have to implement the part where an EUI-64 address can be formed from a short address (we only have the implementation for IEEE 802.15.4 extended addresses), which is defined in [section 3.2.2](https://datatracker.ietf.org/doc/html/rfc6282#section-3.2.2). This means that the full implementation of the compression needs to be rewritten to actually do it with the EUI-64 address. I'm not sure if the case where nothing can be elided from the IPv6 address and the full address is going to be in-lined in the packet is going to occur in BLE.

For multicast IPv6, as I understand it, the multicast compression format for IEEE 802.15.4 is not used in BLE, since there is no multicast/broadcast in BLE. IPv6 multicast packets need to be replicated to each link. However, multicast uses broadcast in IEEE 802.15.4.

Dominaezzz:
Ah I see. In that case this PR isn't what smoltcp is looking for.

PR #1038, opened 2025-01-28
log and drop ipv6 packets requiring fragmentation

Easyoakland:
Currently packets which exceed the MTU are still transmitted for ipv6 without regard for their size.

This pr changes the behavior for ipv6 to match the behavior for ipv4: drop packet and `net_debug` instead of trying to transmit. This prevents panics in `Device` implementations that assume the given packet doesn't exceed an MTU in size.

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -1299,16 +1299,24 @@ impl InterfaceInner {
             }
             // We don't support IPv6 fragmentation yet.
             #[cfg(feature = "proto-ipv6")]
-            IpRepr::Ipv6(_) => tx_token.consume(total_len, |mut tx_buffer| {
-                #[cfg(feature = "medium-ethernet")]
-                if matches!(self.caps.medium, Medium::Ethernet) {
-                    emit_ethernet(&ip_repr, tx_buffer)?;
-                    tx_buffer = &mut tx_buffer[EthernetFrame::<&[u8]>::header_len()..];
-                }
+            IpRepr::Ipv6(_) => {
+                // Check if we need to fragment it.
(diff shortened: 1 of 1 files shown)


src/iface/interface/mod.rs
```diff
@@ -1299,16 +1299,24 @@ impl InterfaceInner {
...
-            IpRepr::Ipv6(_) => tx_token.consume(total_len, |mut tx_buffer| {
-                #[cfg(feature = "medium-ethernet")]
-                if matches!(self.caps.medium, Medium::Ethernet) {
-                    emit_ethernet(&ip_repr, tx_buffer)?;
-                    tx_buffer = &mut tx_buffer[EthernetFrame::<&[u8]>::header_len()..];
-                }
+            IpRepr::Ipv6(_) => {
+                // Check if we need to fragment it.
+                if total_ip_len > self.caps.ip_mtu() {
+                    net_debug!("ipv6 fragmentation support is unimplemented. Dropping.");
```
whitequark: nit: capitalization

```suggestion
                    net_debug!("IPv6 fragmentation support is unimplemented. Dropping.");
```

whitequark (commented):
Looks good to me other than the comment!

src/iface/interface/mod.rs
Easyoakland: Changed.

whitequark:
Thanks! If you squash the commits I'll hit merge.

Easyoakland:
Done.

whitequark (approved):

whitequark:
Thanks for the PR and the prompt response!

PR #1039, opened 2025-02-05
slaac: add initial stateless address autoconfiguration (SLAAC) implementation

bergzand:
This PR adds
- Functions to derive ipv6 addresses from a `Cidr` and a LL address
- An slaac implementation
- Glue to hook it into the `Interface`

This works by having a separate module to track the slaac state machine. The Interface module passes the received router advertisements to this module and polls this to query whether router solicitations must be send and whether the addresses and routes on the interface need to be synced with the slaac state.

Compared to the dhcpv4, this module modifies the address and route state on the `Interface` within the module itself without any outside glue or socket required. 

Currently there's no separate feature for slaac and it is pulled in when `proto-ipv6` together with `medium-ieee802154` or `medium-ethernet` is enabled. Done this way because I consider SLAAC to be an essential part of IPv6. However, I'm not opposed to adding a separate feature for slaac that depends on `proto-ipv6`  

An example is provided that prints the addresses and routes on the interface every second for testing with router advertisement daemons such as radvd

<details><summary>Example output</summary>

```
Addresses:
  - fe80::1/64
Routes:

Addresses:
  - fe80::1/64
Routes:

Addresses:
  - fe80::1/64
  - 2001:db8::ff:fe00:1/64
Routes:
  - ::/0 via fe80::d83c:a5ff:fe9c:3a40

Addresses:
  - fe80::1/64
  - 2001:db8::ff:fe00:1/64
Routes:
  - ::/0 via fe80::d83c:a5ff:fe9c:3a40

Addresses:
  - fe80::1/64
  - 2001:db8::ff:fe00:1/64
Routes:
  - ::/0 via fe80::d83c:a5ff:fe9c:3a40

Addresses:
  - fe80::1/64
  - 2001:db8::ff:fe00:1/64
Routes:

Addresses:
  - fe80::1/64
  - 2001:db8::ff:fe00:1/64
Routes:
```

</details>

Alternative to #948 

I can split out the initial commits of this PR into separate PRs if needed to ease reviewing.


src/iface/mod.rs
```diff
@@ -23,4 +28,9 @@ pub use self::interface::{
 };
 
 pub use self::route::{Route, RouteTableFull, Routes};
+#[cfg(all(
+    feature = "proto-ipv6",
+    any(feature = "medium-ethernet", feature = "medium-ieee802154")
```
Dominaezzz: SLAAC is a general IPv6 thing, why restrict it to any particular medium(s)?

src/wire/ndiscoption.rs
```diff
@@ -402,6 +402,16 @@ pub struct PrefixInformation {
     pub prefix: Ipv6Address,
 }
 
+impl PrefixInformation {
+    /// Validates the prefix information option against check a, b, c in
+    /// https://www.rfc-editor.org/rfc/rfc4862#section-5.5.3
+    pub fn valid_prefix_info(&self) -> bool {
```
Dominaezzz: This probably reads better?
```suggestion
    pub fn is_valid_prefix_info(&self) -> bool {
```

Dominaezzz (commented):
Looks good!

My main comment is here is "Why limit SLAAC to any specific mediums?"

src/iface/mod.rs
Dirbaio: SLAAC generates an IP addr based on the MAC addr. With IP medium there's no MAC addr. How is it supposed to work?

bergzand: as @Dirbaio mentioned, there needs to be support for each link layer with SLAAC, see for example [rfc 2464](https://www.rfc-editor.org/rfc/rfc2464.html#section-4) and [rfc 4944](https://www.rfc-editor.org/rfc/rfc4944#section-6).

Dominaezzz: Sure for link-local address you want to use the MAC address (or equivalent) to generate it, but that's not quite the same for global or temporary addresses where you might want to hide your MAC address. I think in this case the user should be allowed to specify an interface identifier (which would be different from the MAC).

If Linux can do SLAAC for TUN devices, smoltcp can definitely do it for IP mediums.

EDIT: I'm looking at https://datatracker.ietf.org/doc/html/rfc4862#section-5.5

bergzand: > Sure for link-local address you want to use the MAC address (or equivalent) to generate it, but that's not quite the same for global or temporary addresses where you might want to hide your MAC address. I think in this case the user should be allowed to specify an interface identifier (which would be different from the MAC).

I prefer to leave the additional logic required to switch the generation of interface identifiers to a follow up PR as this PR is already complex enough.

> If Linux can do SLAAC for TUN devices, smoltcp can definitely do it for IP mediums.

If there's a way to add an interface identifier to a `Medium::Ip` Interface, it can be hooked in somehow. The whole `process_ndisc` where this code hooks into is skipped on `Medium::Ip` interfaces, so that would need to be modified. Not impossible, but also not sure if that must go in this PR.

Do you know if there's some special handling for SLAAC on Linux tun devices? Or does it always add a privacy address to that interface for slaac?

Dominaezzz: > Or does it always add a privacy address to that interface for slaac?

Yes

> I prefer to leave the additional logic required to switch the generation of interface identifiers to a follow up PR as this PR is already complex enough.

Sure the additional logic isn't necessary for this PR, I'm just saying the `any(feature = "medium-ethernet", feature = "medium-ieee802154")` isn't needed.

bergzand: Sorry for the long silence on this. Still planning on pushing this forward.

> Sure the additional logic isn't necessary for this PR, I'm just saying the `any(feature = "medium-ethernet", feature = "medium-ieee802154")` isn't needed.

I gave this a shot as I agree it *should* be possible to run slaac over IP mediums. 

It's not as easy as just removing these conditional compilation here. The slaac code here depends on existing parsing code in `wire` [that](https://github.com/smoltcp-rs/smoltcp/blob/main/src/wire/mod.rs#L256-L264) is already guarded with `any(feature = "medium-ethernet", feature = "medium-ieee802154")`. Removing the restriction with slaac means also having to remove the conditional compilation on `wire::ndisc` etc, something I consider out of scope for this PR. The ndisc processing can be made optional in the slaac code, but as mentioned above I prefer not to add the extra complexity in this PR.

Furthermore, the whole neighbor discovery processing this ties into has the same conditional compilation guard in the `iface::ipv6` code, also something I don't want to touch in this PR.

vaknin:
Hey @bergzand & @Dominaezzz
What's blocking integrating this?

Dominaezzz:
Fwiw, I neither have the power to approve nor merge this PR.

Dirbaio:
rebased, gated behind `proto-ipv6-slaac` cargo feautre.

PR #1041, opened 2025-02-08
Fix compilation errors when socket-dns is enabled but socket-udp isn't

### LINKED ISSUE #1040: Compilation errors when `socket-dns` is enabled but `socket-udp` isn't
This crate enables the `Frame::Udp` variant if either the `socket-dns` or the `socket-udp` feature is enabled:

https://github.com/smoltcp-rs/smoltcp/blob/3e3afb62b292cc027c8baeee40bce7b644392115/src/iface/packet.rs#L218-L219

Unfortunately, in at least one match statement the `Udp` variant is gated behind the `socket-udp` feature without considering `socket-dns` at all. See for instance here:

https://github.com/smoltcp-rs/smoltcp/blob/3e3afb62b292cc027c8baeee40bce7b644392115/src/iface/packet.rs#L242-L243

This results in a compilation error due to in-exhaustive match.

Would it make sense to have the `socket-dns` feature enable `socket-udp` transitively, rather than having them separately activate the `Udp` frame variant?

scootermon:
Fixes: #1040

I added a test case that would've caught this specific issue, but it likely won't catch any issues in the future. Since I couldn't come up with a better method of testing it more generally, I just left it at that.


whitequark (approved):
Thanks!

PR #1042, opened 2025-02-17
feat: Add initial `serde:{Deserialize, Serialize}` implementation for `Ipv[4|6]Cidr`

AnthonyGrondin:
This PR adds an initial implementation for `serde:{Deserialize, Serialize}`, in preparation for implementing those traits into `embassy-net` types. I've added the minimal required in order to be able to implement `serde:{Deserialize, Serialize}` for `embassy_net::Config`.

Since IP types are now in `core::net`, most already implement serde, and only `Ipv[4|6]Cidr` were missing an implementation. I've also added the derive for `RetryConfig` in order to work with `embassy_net::DhcpConfig`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ bitflags = { version = "1.0", default-features = false }
 defmt = { version = "0.3.8", optional = true, features = ["ip_in_core"] }
 cfg-if = "1.0.0"
 heapless = "0.8"
+serde = { version = "1.0.217", default-features = false, features = ["derive"], optional = true }
 
 [dev-dependencies]
 env_logger = "0.10"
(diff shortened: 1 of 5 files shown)


whitequark (commented):
I would prefer to not have a dependency on serde at all.

AnthonyGrondin:
Closing as [serde](https://github.com/serde-rs/serde) can be quite heavy and users should rather use application specific types for their use-cases. 

For more information: https://github.com/embassy-rs/embassy/pull/3897#issuecomment-2663947596

jordens:
> I would prefer to not have a dependency on serde at all.

Out of curiosity: why?
It would be optional and there is precedent (e.g. `defmt`).

whitequark:
> Out of curiosity: why?

I don't think this issue tracker at this time is the best place to go into it in depth, but it relates both to the way it is implemented and to its author.

PR #1059, opened 2025-05-08
fix: exhaust egress socket state on `poll`

thomaseizinger:
At present, `smoltcp` only does a single pass of advancing the egress socket state. This means the layer above has to call `poll` in a loop to ensure sufficient progress has been made.

We already exhaustively progress the ingress state further up so doing the same thing for the egress state seems appropriate.


thvdveld:
@whitequark @Dirbaio I don't see any reason why not to do this.

whitequark:
I thought @Dirbaio had a case where this caused an issue?

thvdveld:
I thought that as well and was digging in old issues and PRs, but I didn't find anything related to the egress. Most of the issues were with ingress resulting in a DoS.

whitequark (approved):

whitequark:
@thomaseizinger Could you re-run/fix the test please? We'll merge it after.

thomaseizinger:
> @thomaseizinger Could you re-run/fix the test please? We'll merge it after.

I just accepted the new snapshot but I don't really know what it is testing.

whitequark:
Thanks!

PR #1061, opened 2025-05-10
Reuse server DHCPOFFER transaction id in DHCPREQUEST

simonborje:
Hello! From my understanding of the DHCP specification [rfc2131](https://datatracker.ietf.org/doc/html/rfc2131) smoltcp currently does not follow the specification when it comes to the `transaction id` used for DHCPREQUEST messages. I initially saw this mentioned in a discussion here https://github.com/esp-rs/esp-hal/discussions/2345#discussioncomment-11179188.

If I understand the spec. correctly, when a DHCPOFFER is received from a server, the follow-up DHCPREQUEST should reuse the `transaction id` from the DHCPOFFER message. Currently a new random `transaction id`is used for the request instead. 

The relevant parts of the RFC is Table 5 and the text just after it (page 36-37).

I have only been able to test this on my home setup so far but it seems to work as expected there at least. I'm not sure how to add a proper test for this or if the change might interact with retries in some way that needs additional handling.

### DIFF
--- a/src/socket/dhcpv4.rs
+++ b/src/socket/dhcpv4.rs
@@ -72,6 +72,8 @@ struct RequestState {
     server: ServerInfo,
     /// IP address that we're trying to request.
     requested_ip: Ipv4Address,
+    /// Transaction ID from server's `Offer` message
+    offer_transaction_id: u32,
 }
 
 #[derive(Debug)]
@@ -379,6 +381,7 @@ impl<'a> Socket<'a> {
                         identifier: server_identifier,
                     },
                     requested_ip: dhcp_repr.your_ip, // use the offered ip
+                    offer_transaction_id: dhcp_repr.transaction_id,
                 });
             }
             (ClientState::Requesting(state), DhcpMessageType::Ack) => {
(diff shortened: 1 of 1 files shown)


astro (approved):

simonborje:
Thanks for the review!
Is there anything I should do before this can be merged? I see that some checks are marked as failed.

src/socket/dhcpv4.rs
```diff
@@ -72,6 +72,8 @@ struct RequestState {
     server: ServerInfo,
     /// IP address that we're trying to request.
     requested_ip: Ipv4Address,
+    /// Transaction ID from server's `Offer` message
+    offer_transaction_id: u32,
```
Dirbaio: this field is redundant. We know the xid of DHCPOFFER is the same as the xid of DHCPDISCOVER, which is stored in the `Socket.transaction_id` field.

simonborje:
This seems to work as well - I see that the DHCP Discover and Request use the same transaction id.

Dirbaio:
looks good. I got a few more questions tho:

- Should the xid of the DHCPREQUEST in renew/rebind also match the original xid of the DHCPOFFER? The current code is generating them randomly. I've read through the RFC but I don't see an obvious answer.
- The code is always calling `random_transaction_id` but then overwriting it in the Requesting branch. Maybe it'd be better to only call `random_transaction_id` in the branches where it's actually needed to avoid unnecessary RNG calls.

simonborje:
I can't find a clear answer regarding DHCPREQUEST xid:s at renew/rebind in the RFC. However, both Windows `ipconfig /renew`, and linux `dhclient enp1s0` seem to use new xid:s in the DHCPREQUEST when renewing.

Dirbaio (approved):
thanks for checking. looks good to me then 🚀

PR #1063, opened 2025-06-10
Allow withholding the SYN|ACK packet by user code

cvengler:
We at [The Tor Project, Inc.](https://torproject.org) have been using this trivial patch in a [downstream fork](https://gitlab.torproject.org/tpo/core/smoltcp) of smoltcp for almost 1.5 years. We never bothered to upstream it, because the assumed the feature was too much of a niche one. However, given the [recent release of oniux](https://blog.torproject.org/introducing-oniux-tor-isolation-using-linux-namespaces/), which uses the downstream fork of smoltcp extensively, we thought it is a good idea to upstream it, given that many distributions already bundle oniux themselves.

<hr>

In certain use cases, it's desirable to not send a SYN|ACK packet immediately after receiving a SYN -- for example, a TCP proxy that doesn't want to do so until it's connected to the end destination, because the outgoing connection might get refused. (Currently, you have to send a SYN|ACK and then reset the connection afterwards in this case.)

To fix this, add a simple `synack_paused` flag, controllable by user code, that withholds SYN|ACK packets in `SynReceived` state until it is unset.


cvengler:
I assume the CI failures are related to outdated versions in the main branch? This PR is at least not modifying any parts of Cargo.toml et, al.

cvengler:
My bad, there was already a PR for it in #863.

I do understand the argument made there but given that this is more or less already deployed in production systems I think it is worth iterating over it again.

whitequark:
@Dirbaio How about adding this feature behind a feature flag?

Dirbaio:
Yes, I'd prefer it to be gated behind a cargo feature so it doesn't increase code size when you don't need it (yes it's *very little code size* but for embedded use cases i'll take every byte I can get :sweat_smile: )

Also, could you add a test?

cvengler:
Sure thing!
Do you have suggestions for a name of that feature flag?
I would just go with `pause_synack`?

Dirbaio:
probably `socket-tcp-pause-synack` for consistency with the other flags that enable tcp features.

cvengler:
placed it behind a feature flag and added a unit test 🙂

src/socket/tcp.rs
```diff
@@ -3332,6 +3354,17 @@ mod test {
...
 
+    #[cfg(feature = "socket-tcp-pause-synack")]
+    #[test]
+    fn test_syn_paused_ack() {
+        let mut s = socket_syn_received();
+
+        s.pause_synack(true);
+        recv_nothing!(s);
+        assert_eq!(s.state, State::SynReceived);
+        sanity!(s, socket_syn_received());
```
whitequark: Probably should test the unpausing too.

cvengler: done

whitequark (approved):

cvengler:
What is the current state of this? 🙂

whitequark:
Waiting for someone to fix CI.

cvengler:
> Waiting for someone to fix CI.

It seems to be unrelated to the changes made in the PR though, right?

whitequark:
Yes, sorry, let me clarify. Your changes are fine. I was hoping that one of the other smoltcp maintainers would fix the CI (hence "someone", as in "not you"), but I might have to do it.

whitequark:
Actually, do you think you could rebase on top of main?

cvengler:
> Actually, do you think you could rebase on top of main?

Sure

cvengler:
rebased.

whitequark:
Thanks!

PR #1067, opened 2025-06-24
raw: optional IP version and next_header

### LINKED ISSUE #1064: Remove IP version and next header checks for raw sockets
`RawSocket` will currently only send and receive packets which match the given `ip_version` and `ip_protocol`. I think the current behaviour tries to match linux sockets. It could be useful to accept all packets, independent of `IpVersion` and `IpProtocol` via the same socket however.

Currently if I want to receive Ipv4 and Ipv6 packets for TCP, UDP and ICMP the only way I can see to do this is to create 6 different sockets with 6x the number of buffers (please let me know if I'm missing something).

Perhaps the `ip_version` and `ip_protocol` fields could be made `Option`s, or another `ExtraRaw` socket created if we don't want to change existing behaviour.

KingCol13:
Fixes #1064 

Makes `ip_version` and `ip_protocol` fields of `RawSocket` into options, when `None` the socket will not check this field when sending or receiving.

### DIFF
--- a/examples/multicast.rs
+++ b/examples/multicast.rs
@@ -66,8 +66,8 @@ fn main() {
     // Will not send IGMP
     let raw_tx_buffer = raw::PacketBuffer::new(vec![], vec![]);
     let raw_socket = raw::Socket::new(
-        IpVersion::Ipv4,
-        IpProtocol::Igmp,
+        Some(IpVersion::Ipv4),
+        Some(IpProtocol::Igmp),
         raw_rx_buffer,
         raw_tx_buffer,
     );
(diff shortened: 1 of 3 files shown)


whitequark:
I have two concerns.
1. This is a breaking change. @Dirbaio, do we just increment the minor? I'm not sure what the current policy is.
2. Should these be two individual options, or one option with a tuple of (IpVersion, IpProtocol)?

KingCol13:
>     1. This is a breaking change. @Dirbaio, do we just increment the minor? I'm not sure what the current policy is.

Following Rust's modified semantic versioning we'd bump to `0.13.0`. I don't know if that should be done in this PR or just before release though. Let me know if you want me to add to the changelog.

>     2. Should these be two individual options, or one option with a tuple of (IpVersion, IpProtocol)?

Two options gives users more flexibility, e.g. if I want to accept all Ipv4 packets without filtering on the protocol.

whitequark:
> Two options gives users more flexibility, e.g. if I want to accept all Ipv4 packets without filtering on the protocol.

Yes, good point. I'm fine with two options.

I haven't done release engineering on smoltcp in a while so I think we can just merge this, noting that the next release should be a minor bump.

whitequark (approved):

PR #1069, opened 2025-07-01
Do not send TCP reset when using raw socket which accepts TCP

### LINKED ISSUE #1068: Do not reset TCP connections when using a RawSocket
When the socket set only contains a raw socket, all TCP connections will be reset by the code here:
https://github.com/smoltcp-rs/smoltcp/blob/e000f15ebc8b695c481ad8e12bef1536077a422f/src/iface/interface/tcp.rs#L40-L41

However, when using a raw socket which can accept TCP, this is unlikely to be desired behaviour.

This code should probably be modified to check if any raw sockets are able to accept the packet before deciding whether to send an RST.

I'm happy to have a go at fixing this, let me know if this sounds unreasonable.

KingCol13:
Fixes #1068

### DIFF
--- a/src/iface/interface/ipv4.rs
+++ b/src/iface/interface/ipv4.rs
@@ -227,7 +227,9 @@ impl InterfaceInner {
             }
 
             #[cfg(feature = "socket-tcp")]
-            IpProtocol::Tcp => self.process_tcp(sockets, ip_repr, ip_payload),
+            IpProtocol::Tcp => {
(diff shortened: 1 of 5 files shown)


KingCol13:
Sorry for the initial false start, this now matches `process_udp` in `src/iface/interface/udp.rs` behaviour when a raw socket has handled the packet. The coverage change seems to be just because `rustfmt` added a new line. I think this is ready.

thvdveld:
Thank you! I indeed think that's the intended behaviour.

PR #1072, opened 2025-07-10
Single frame ingress to make custom Layer 2 protocols possible

dscso:
This is an idea how to make it simpler for users to handle non-IP packets externally, without passing though Smoltcp. This is just an Idea, how I implemented a quick and dirty fix for my problem.

https://github.com/smoltcp-rs/smoltcp/issues/942

```rust
pub fn poll(/*...*/) {
    // instead of using
    // iface.poll(now(), &mut &mut device, &mut sockets);
    // Only IP packets are getting passed to the TCP/IP stack

    let mut f = || {
        let Some((rx_token, tx_token)) = device.receive(now()) else {
            return PollIngressSingleResult::None;
        };

        let rx_meta = rx_token.meta();
        rx_token.consume(|frame| {
            if frame[12..12 + 2] == [0x88, 0xf7] { // filter out L2 traffic I want
                defmt::info!("PACKET {:x}", frame);
                return PollIngressSingleResult::PacketProcessed;
            }
           iface
                .poll_ingress_single_frame(&mut sockets, rx_meta, frame, tx_token)
        })
    };

    loop {
        match f() {
            PollIngressSingleResult::None => break,
            PollIngressSingleResult::PacketProcessed => {}
            PollIngressSingleResult::SocketStateChanged => {}
        }
    }
    // Process egress.
    match iface.poll_egress(now(), &mut &mut device, &mut sockets)
    {
        PollResult::None => {}
        PollResult::SocketStateChanged => {}
    }
}
```

Another possibility could be to add a closure for each frame received:
```rust
iface.poll_egress_frame(now(), &mut dma, &mut sockets, |frame| {
   if frame[12..12 + 2] == [0x88, 0xf7] {
      defmt::info!("PACKET {:x}", frame);
      return PollIngressSingleResult::PacketProcessed;
   }
   PollIngressSingleResult::None
})
```

It should also be possible to write a wrapper `impl Device` type, that makes L2 protocols possible, although I don't see an elegant way of implementing this without copying. 

Yet another option would be to outsource this problem to the drivers, and add `peek` support to the drivers.

What do you think?

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -572,6 +572,74 @@ impl Interface {
         }
     }
 
+    /// Allows user to just ingress a single frame
+    /// This can be useful if Smoltcp is being used in combination with Layer 2 protocols
+    pub fn poll_ingress_single_frame<'frame>(
+        &mut self,
+        sockets: &mut SocketSet,
+        rx_meta: crate::phy::PacketMeta,
+        frame: &'frame [u8],
+        tx_token: impl TxToken,
+    ) -> PollIngressSingleResult {
+        if frame.is_empty() {
+            return PollIngressSingleResult::PacketProcessed;
+        }
+
+        match self.inner.caps.medium {
+            #[cfg(feature = "medium-ethernet")]
+            Medium::Ethernet => {
+                if let Some(packet) =
+                    self.inner
+                        .process_ethernet(sockets, rx_meta, frame, &mut self.fragments)
+                {
+                    if let Err(err) = self.inner.dispatch(tx_token, packet, &mut self.fragmenter) {
+                        net_debug!("Failed to send response: {:?}", err);
+                    }
+                }
+            }
+            #[cfg(feature = "medium-ip")]
+            Medium::Ip => {
+                if let Some(packet) =
+                    self.inner
+                        .process_ip(sockets, rx_meta, frame, &mut self.fragments)
+                {
+                    if let Err(err) = self.inner.dispatch_ip(
+                        tx_token,
+                        PacketMeta::default(),
+                        packet,
+                        &mut self.fragmenter,
+                    ) {
+                        net_debug!("Failed to send response: {:?}", err);
+                    }
+                }
+            }
+            #[cfg(feature = "medium-ieee802154")]
+            Medium::Ieee802154 => {
+                if let Some(packet) =
+                    self.inner
+                        .process_ieee802154(sockets, rx_meta, frame, &mut self.fragments)
+                {
+                    if let Err(err) = self.inner.dispatch_ip(
+                        tx_token,
+                        PacketMeta::default(),
+                        packet,
+                        &mut self.fragmenter,
+                    ) {
+                        net_debug!("Failed to send response: {:?}", err);
+                    }
+                }
+            }
+        }
+
+        // TODO: Propagate the PollIngressSingleResult from deeper.
+        // There's many received packets that we process but can't cause sockets
+        // to change state. For example IP fragments, multicast stuff, ICMP pings
+        // if they dont't match any raw socket...
+        // We should return `PacketProcessed` for these to save the user from
+        // doing useless socket polls.
+        PollIngressSingleResult::SocketStateChanged
+    }
+
     fn socket_ingress(
         &mut self,
         device: &mut (impl Device + ?Sized),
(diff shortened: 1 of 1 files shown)


whitequark:
This feature would complicate an already complex API. What is the use case and what are you using currently?

dscso:
Thanks for the fast answer! And yea, I agree, I am also not very happy with this implementation :)

I am trying to implement a Layer 2 protocol (gPTP), but Smoltcp drops all non-IP traffic. Therefore, I was thinking to sort BEFORE passing the Ethernet frames into the TCP/IP stack. This is because the `impl Device` Trait is not very easy to work with, if you are having two consumers (TCP/IP stack and the Layer 2 protocol).

I seem to not be alone with this like (https://github.com/smoltcp-rs/smoltcp/issues/942). Do you have an idea how to solve this issue idiomatically?

I am using the [stm32-eth](https://github.com/stm32-rs/stm32-eth) as the `Device` btw.

Ahh and using RawSockets does not work, since RawSockets only support IP traffic.

whitequark:
> This is because the `impl Device` Trait is not very easy to work with, if you are having two consumers (TCP/IP stack and the Layer 2 protocol).

Can you tell me about the issues you encounter implementing an "interposer" `impl Device` struct? That would be my first choice for this use case.

whitequark:
> Ahh and using RawSockets does not work, since RawSockets only support IP traffic.

Since PTP is involved, should I assume that you also have additional requirements for the Ethernet packets; low/predictable handling latency, additional metadata, etc?

dscso:
> Can you tell me about the issues you encounter implementing an "interposer" impl Device struct? That would be my first choice for this use case.

I have tried this, but didn't got too far with it. Tomorrow I will try again, but  I would have to copy data, which I'd prefer to avoid. Might be a skill issue tho :D

> Since PTP is involved, should I assume that you also have additional requirements for the Ethernet packets; low/predictable handling latency, additional metadata, etc?

Yep exactly. Mainly the packet ID so I can "ask" the Ethernet driver for the Timestamp when the package arrived. But I think this should not be too much of a problem.

whitequark:
> Yep exactly. Mainly the packet ID so I can "ask" the Ethernet driver for the Timestamp when the package arrived. But I think this should not be too much of a problem.

That rather sounds like you'd have to add an interposer anyway...

dscso:
**What do I want to achieve:**
I want to intercept layer 2 traffic, which is not of interest for the TCP/IP stack. Furthermore I'd like to tag traffic with the [IEEE 802.1Q](https://de.wikipedia.org/wiki/IEEE_802.1Q) header. So it would be quite handy if I could intercept the packets between the TCP/IP stack and the [network driver](github.com/stm32-rs/stm32-eth).


So I have been trying to build an imposer, unfortunately however I am struggling with a simple pass though solution. Another option would be to copy the implementation of the `Device` trait implementation into my application, although this would also not be the best way to do it.

I totally agree, that my PR suggestion is a pretty bad solution to this problem, maybe someone has a better idea how to solve this problem?

```rust
use core::marker::PhantomData;
use smoltcp::phy::{Device, DeviceCapabilities, RxToken, TxToken};
use smoltcp::time::Instant;

pub struct EthDMA<'a, D>
where
    D: Device<RxToken<'a>: RxToken, TxToken<'a>: TxToken> + ?Sized,
{
    device: &'a mut D,
}

impl<'a, D> Device for EthDMA<'a, D>
where
    D: Device<RxToken<'a>: RxToken, TxToken<'a>: TxToken> + ?Sized,
{
    type RxToken<'token>
        = <D as smoltcp::phy::Device>::RxToken<'a>
    where
        Self: 'token;
    type TxToken<'token>
        = <D as smoltcp::phy::Device>::TxToken<'a>
    where
        Self: 'token;

    fn receive(&mut self, timestamp: Instant) -> Option<(Self::RxToken<'_>, Self::TxToken<'_>)> {
        let res = self.device.receive(timestamp);
        // do stuff with obj

        let Some((rx, tx)) = res else { return None };
        let rx = WrapEthRxToken {
            inner: rx,
            phantom: PhantomData,
        };
        let tx = WrapEthTxToken {
            inner: tx,
            phantom: PhantomData,
        };
        return Some((rx, tx));
    }

    fn transmit(&mut self, timestamp: Instant) -> Option<Self::TxToken<'_>> {
        todo!()
    }

    fn capabilities(&self) -> DeviceCapabilities {
        todo!()
    }
}

pub struct WrapEthRxToken<'rx, T>
where
    T: RxToken,
{
    inner: T,
    phantom: PhantomData<&'rx T>,
}

impl<'rx, T> RxToken for WrapEthRxToken<'rx, T>
where
    T: RxToken,
{
    fn consume<R, F>(self, f: F) -> R
    where
        F: FnOnce(&[u8]) -> R,
    {
        self.inner.consume(f)
    }
}

pub struct WrapEthTxToken<'tx, T>
where
    T: TxToken,
{
    inner: T,
    phantom: PhantomData<&'tx T>,
}

impl<'tx, T> TxToken for WrapEthTxToken<'tx, T>
where
    T: TxToken,
{
    fn consume<R, F>(self, len: usize, f: F) -> R
    where
        F: FnOnce(&mut [u8]) -> R,
    {
        self.inner.consume(len, f)
    }

    fn set_meta(&mut self, meta: smoltcp::phy::PacketMeta) {}
}

```
I've tried quite a few things but I am really struggling with the generic implementation + lifetimes. This is where I gave up :D
```
  --> src/device.rs:38:22
   |
38 |         return Some((rx, tx));
   |                      ^^ expected associated type, found `WrapEthRxToken<'_, ...>`
   |
   = note: expected associated type `<D as Device>::RxToken<'a>`
                       found struct `WrapEthRxToken<'_, <D as Device>::RxToken<'_>>`
```

dscso:
I think there must be another way to build something like this. Until now I think the only way I can think of is to write a custom `Device` implementation, for each hardware separately.

The way outlined in this PR is pretty inflexible, since it would not allow modifying tx packets to e.g. encapsulate them in a 801.2q header.

It might be interesting to think about another way to get packets in and out of the TCP/IP stack. Maybe iterators? 

@whitequark is right, this is not the way. Thanks for the help tho!

PR #1074, opened 2025-07-15
Improve IPv4 source address selection for multi-subnet interfaces

maxslarsson:
### Problem

When an interface has multiple IPv4 addresses in different subnets, the current implementation always returns the first IPv4 address as the source address, regardless of the destination. This causes issues in multi-homed configurations where the interface serves multiple networks.

I encountered this issue with a setup where I registered an interface like this:
```rust
let first_ip = Ipv4Cidr::new(Ipv4Addr::new(172, 18, 1, 2), 24);
let second_ip = Ipv4Cidr::new(Ipv4Addr::new(172, 24, 24, 14), 24);

let mut interface = Interface::new(interface_config, &mut device, now);
interface.update_ip_addrs(|a| {
    a.push(first_ip.into()).unwrap();
    a.push(second_ip.into()).unwrap();
});
```

When a smoltcp socket attempts to send a packet to a device on the `172.24.24.0/24` network, the initial ARP packets that were sent were being sent with the wrong source IP (`172.18.1.2` instead of `172.24.24.14`). This caused the remote device to ignore the ARP requests since they appeared to come from a different network, causing the packet to never be sent.

### Solution

This PR enhances the IPv4 source address selection algorithm to be subnet-aware. The new implementation:

1. First attempts to find an interface address that is in the same subnet as the destination address
2. Falls back to the first IPv4 address if no matching subnet is found (preserving backward compatibility)

### Changes

- Modified `get_source_address_ipv4()` in `src/iface/interface/ipv4.rs` to iterate through all interface addresses and check if they're in the same subnet as the destination
- Updated documentation to reflect the new subnet-aware behavior
- The change maintains backward compatibility - if no subnet match is found, it returns the first IPv4 address as before

Now when communicating with a device at `172.24.24.100`, the source address will correctly be `172.24.24.14` instead of `172.18.1.2`.

### DIFF
--- a/src/iface/interface/ipv4.rs
+++ b/src/iface/interface/ipv4.rs
@@ -38,17 +38,27 @@ impl InterfaceInner {
 
     /// Get an IPv4 source address based on a destination address.
     ///
-    /// **NOTE**: unlike for IPv6, no specific selection algorithm is implemented. The first IPv4
-    /// address from the interface is returned.
+    /// This function tries to find the first IPv4 address from the interface
+    /// that is in the same subnet as the destination address. If no such
+    /// address is found, the first IPv4 address from the interface is returned.
     #[allow(unused)]
-    pub(crate) fn get_source_address_ipv4(&self, _dst_addr: &Ipv4Address) -> Option<Ipv4Address> {
+    pub(crate) fn get_source_address_ipv4(&self, dst_addr: &Ipv4Address) -> Option<Ipv4Address> {
+        let mut first_ipv4 = None;
         for cidr in self.ip_addrs.iter() {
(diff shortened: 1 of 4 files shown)


Dirbaio:
Looks good!

- Can you do the same change for IPv6?
- Can you add tests?

maxslarsson:
- Just looked into it and it seems that it is automatically handled for IPv6 by rule 8 in RFC 6724: use longest matching prefix. I added a test case for it to verify that that is the case.
- I added test cases for both IPv4 and IPv6, following the existing standards and style.

Please let me know if there's anything else!

thvdveld:
Looks good to me, thank you!

PR #1076, opened 2025-07-19
tracer: update public API for smoltcp::phy::Tracer to allow custom inspection and printing of packet

tomkris:
Currently `smoltcp::phy::tracer::Packet` is not exported as public type, which leaves very limited options for using `Tracer` device: traced packets only can be directly printed "as-is" and only in limited context (inside the closure).

I'm updating API of `Tracer` device to provide access to internal fields of traced packet; this will enable implementation of custom printing or packet inspection functions

I renamed `smoltcp::phy::tracer::Packet` to `smoltcp::phy::tracer::TracerPacket`, following prefixing convention used by types declared in `smoltcp::phy::pcap_writer`

### DIFF
--- a/src/phy/tracer.rs
+++ b/src/phy/tracer.rs
@@ -11,12 +11,12 @@ use crate::wire::pretty_print::{PrettyIndent, PrettyPrint};
 /// device.
 pub struct Tracer<D: Device> {
     inner: D,
-    writer: fn(Instant, Packet),
+    writer: fn(Instant, TracerPacket),
 }
 
 impl<D: Device> Tracer<D> {
     /// Create a tracer device.
-    pub fn new(inner: D, writer: fn(timestamp: Instant, packet: Packet)) -> Tracer<D> {
+    pub fn new(inner: D, writer: fn(timestamp: Instant, packet: TracerPacket)) -> Tracer<D> {
         Tracer { inner, writer }
     }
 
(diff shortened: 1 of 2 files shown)


src/phy/tracer.rs
```diff
@@ -148,15 +147,56 @@ impl<Tx: phy::TxToken> phy::TxToken for TxToken<Tx> {
     }
 }
 
-pub struct Packet<'a> {
+/// Packet which is being traced by [Tracer](struct.Tracer.html) device.
+pub struct TracerPacket<'a> {
     buffer: &'a [u8],
     medium: Medium,
-    prefix: &'static str,
+    direction: TracerDirection,
```
whitequark: I think all of these can be `pub` fields. Also, the struct as a whole should have `#[derive(Debug, Clone)]` at least (probably `Copy` too)

whitequark (changes requested):

whitequark:
It looks like there's no tests, can you add some?

src/phy/tracer.rs
tomkris: Updated interface of `TracerPkt` to expose all fields publicly, removed now redundant constructor and accessor fields. Added `Debug/Clone/Copy` auto-derives.

tomkris:
Thank you for taking a look Catherine,

Added unit test for the `Tracer` device implementation and for `Display` implementation of `TracerPacket`.
Also added standard derives to `TracerDirection`

whitequark (approved):
LGTM, thanks!

PR #1079, opened 2025-08-11
tcp: Reject bytes outside the receive window

### LINKED ISSUE #1048: TcpSocket panicked on "attempt to subtract sequence numbers with underflow"
```
thread 'smoltcp-poll' panicked at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/wire/tcp.rs:81:13:
attempt to subtract sequence numbers with underflow
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library/core/src/panicking.rs:75:14
   2: <smoltcp::wire::tcp::SeqNumber as core::ops::arith::Sub>::sub
             at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/wire/tcp.rs:81:13
   3: smoltcp::socket::tcp::Socket::last_scaled_window
             at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/socket/tcp.rs:698:33
   4: smoltcp::socket::tcp::Socket::window_to_update
             at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/socket/tcp.rs:2240:41
   5: smoltcp::socket::tcp::Socket::dispatch
             at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/socket/tcp.rs:2311:19
   6: smoltcp::iface::interface::Interface::socket_egress
             at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/iface/interface/mod.rs:717:40
   7: smoltcp::iface::interface::Interface::poll_egress
             at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/iface/interface/mod.rs:497:9
   8: smoltcp::iface::interface::Interface::poll
             at /Users/zonyitoo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/iface/interface/mod.rs:459:15
   9: shadowsocks_service::local::tun::tcp::TcpTun::new::{{closure}}
             at ./crates/shadowsocks-service/src/local/tun/tcp.rs:329:65
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Why the sequence number underflowed? <del>It only happens when building with `--release`.</del>

### LINKED ISSUE #1051: Crashed @ smoltcp::socket::tcp::Socket::window_to_update
I found the case crashed at ~smoltcp::socket::tcp::Socket::window_to_update~ below,
and just protect this case,

How to fix this bug correct

# Crash Info
message is 'attempt to subtract sequence numbers with underflow'

stack is 

> #00  pc 0x000000000004e178  /apex/com.android.runtime/lib64/bionic/libc.so (abort+168)
> #01  pc 0x00000000006af6e0  xxx.so (panic_abort::__rust_start_panic::abort::h40146cc021174310) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #02  pc 0x00000000006af5b0  xxx.so (__rust_start_panic) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #03  pc 0x00000000008cc970  xxx.so (rust_panic) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #04  pc 0x00000000008cc028  xxx.so (std::panicking::rust_panic_with_hook::hd549f2da7a7179ed) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #05  pc 0x00000000008fafe0  xxx.so (std::panicking::begin_panic_handler::_$u7b$$u7b$closure$u7d$$u7d$::h4a032a406e84da32) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #06  pc 0x00000000008faf80  xxx.so (std::sys::backtrace::__rust_end_short_backtrace::h412ee0edea7f5db4) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #07  pc 0x00000000008fc580  xxx.so (rust_begin_unwind) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #08  pc 0x000000000058dc78  xxx.so (core::panicking::panic_fmt::h84eec9528e2245d0) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #09  pc 0x00000000008c4b6c  xxx.so (smoltcp::socket::tcp::Socket::window_to_update::heaa8a81aaeac8d2f) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #10  pc 0x0000000000847438  xxx.so (smoltcp::iface::interface::Interface::poll::h0975ea8b4eec6382) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #11  pc 0x000000000084d07c  xxx.so (std::sys::backtrace::__rust_begin_short_backtrace::hf376d5be559f9429) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #12  pc 0x00000000008be9ec  xxx.so (core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h43936b4f411de16c) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #13  pc 0x00000000008fd750  xxx.so (std::sys::pal::unix::thread::Thread::new::thread_start::ha53eba763406fe33) (BuildId: 00cd83064dfe907c2ec57d775915ca8e)
> #14  pc 0x00000000000af97c  /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_start(void*)+64)
> #15  pc 0x00000000000500d0  /apex/com.android.runtime/lib64/bionic/libc.so (__start_thread+64)

# Origin code
```rust
    #[inline]
    fn last_scaled_window(&self) -> Option<u16> {
        let last_ack = self.remote_last_ack?;
        let next_ack = self.remote_seq_no + self.rx_buffer.len();

        let last_win = (self.remote_last_win as usize) << self.remote_win_shift;
        let last_win_adjusted = ilast_ack + last_win - next_ack;

        Some(u16::try_from(last_win_adjusted >> self.remote_win_shift).unwrap_or(u16::MAX))
    }
```

# Protected code
```rust
    #[inline]
    fn last_scaled_window(&self) -> Option<u16> {
        let last_ack = self.remote_last_ack?;
        let next_ack = self.remote_seq_no + self.rx_buffer.len();

        let last_win = (self.remote_last_win as usize) << self.remote_win_shift;
        let last_win_adjusted = if last_ack + last_win >= next_ack {
            last_ack + last_win - next_ack
        } else {
            0
        };

        Some(u16::try_from(last_win_adjusted >> self.remote_win_shift).unwrap_or(u16::MAX))
    }
```

lrh2000:
In the current implementation, we accept incoming bytes as long as there is space in the receive buffer. This occurs even if we never advertise its availability. For example, a previous ACK packet may be delayed due to the ACK delay mechanism.
https://github.com/smoltcp-rs/smoltcp/blob/d2d647090d544b1e7c142571da9d55f7280f664b/src/socket/tcp.rs#L1598-L1599

We may not want to do this because it may mess up our internal states by breaking some invariants, e.g., `last_ack + last_win - next_ack` should always work without overflows/underflows:
https://github.com/smoltcp-rs/smoltcp/blob/d2d647090d544b1e7c142571da9d55f7280f664b/src/socket/tcp.rs#L698

As a PoC, if you test the newly added `test_recv_out_of_recv_win` test in the main branch, you'll see the following:
```
---- socket::tcp::test::test_recv_out_of_recv_win stdout ----
send: TCP src=49500 dst=80 psh seq=4294957296 ack=10001 win=256 len=32
rx buffer: receiving 32 octets at offset 0
rx buffer: enqueueing 32 octets (now 32)
starting delayed ack timer
send: TCP src=49500 dst=80 psh seq=4294957328 ack=10001 win=256 len=1
rx buffer: receiving 1 octets at offset 0
rx buffer: enqueueing 1 octets (now 33)
delayed ack timer already started, forcing expiry
outgoing segment will acknowledge
sending ACK
recv: TCP src=80 dst=49500 seq=10001 ack=4294957329 win=31 len=0
stop delayed ack timer (was force-expired)
rx buffer: dequeueing 1 octets (now 32)
send: TCP src=49500 dst=80 psh seq=4294957329 ack=10001 win=256 len=32
rx buffer: receiving 32 octets at offset 0
rx buffer: enqueueing 32 octets (now 64)
starting delayed ack timer
thread 'socket::tcp::test::test_recv_out_of_recv_win' panicked at src/wire/tcp.rs:81:13:
attempt to subtract sequence numbers with underflow
```

The test will pass after changes in this PR.

I think this is the root cause of #1048 and #1051. So this PR may fix #1048 and fix #1051.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -6794,6 +6794,74 @@ mod test {
     }
 
     #[test]
+    fn test_recv_out_of_recv_win() {
+        let mut s = socket_established();
+        s.set_ack_delay(Some(ACK_DELAY_DEFAULT));
+        s.remote_mss = 32;
+
+        // No ACKs are sent due to the ACK delay.
+        send!(
+            s,
+            TcpRepr {
+                control: TcpControl::Psh,
+                seq_number: REMOTE_SEQ + 1,
+                ack_number: Some(LOCAL_SEQ + 1),
+                payload: &[0; 32],
+                ..SEND_TEMPL
+            }
+        );
+        recv_nothing!(s);
+
+        // RMSS+1 bytes of data has been received, so ACK is sent without delay.
+        send!(
+            s,
+            TcpRepr {
+                control: TcpControl::Psh,
+                seq_number: REMOTE_SEQ + 33,
+                ack_number: Some(LOCAL_SEQ + 1),
+                payload: &[0; 1],
+                ..SEND_TEMPL
+            }
+        );
+        recv!(
+            s,
+            Ok(TcpRepr {
+                seq_number: LOCAL_SEQ + 1,
+                ack_number: Some(REMOTE_SEQ + 34),
+                window_len: 31,
+                ..RECV_TEMPL
+            })
+        );
+
+        // This frees up a byte in the receive buffer. However, the remote shouldn't be aware of
+        // this since no ACKs are sent.
+        s.recv_slice(&mut [0; 1]).unwrap();
+        recv_nothing!(s);
+
+        // Now, if the remote wants to send one byte outside of the receive window that we
+        // previously advertised, it should not succeed.
+        send!(
+            s,
+            TcpRepr {
+                control: TcpControl::Psh,
+                seq_number: REMOTE_SEQ + 34,
+                ack_number: Some(LOCAL_SEQ + 1),
+                payload: &[0; 32],
+                ..SEND_TEMPL
+            }
+        );
+        recv!(
+            s,
+            Ok(TcpRepr {
+                seq_number: LOCAL_SEQ + 1,
+                ack_number: Some(REMOTE_SEQ + 65),
+                window_len: 1, // The last byte isn't accepted.
+                ..RECV_TEMPL
+            })
+        );
+    }
+
+    #[test]
     fn test_close_wait_no_window_update() {
         let mut s = socket_established();
         send!(
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1677,7 +1677,11 @@ impl<'a> Socket<'a> {
         }
 
         let window_start = self.remote_seq_no + self.rx_buffer.len();
-        let window_end = self.remote_seq_no + self.rx_buffer.capacity();
+        let window_end = if let Some(last_ack) = self.remote_last_ack {
+            last_ack + ((self.remote_last_win as usize) << self.remote_win_shift)
+        } else {
+            window_start
+        };
         let segment_start = repr.seq_number;
         let segment_end = repr.seq_number + repr.payload.len();
 


whitequark (approved):
Thanks, good catch!

dnadlinger:
Related RFC 9293 paragraph?

https://datatracker.ietf.org/doc/html/rfc9293#section-3.10.7.4-2.1.2.8.2.6.1

> If an incoming segment is not acceptable, an acknowledgment should be sent in reply (unless the RST bit is set, if so drop the segment and return): <SEQ=SND.NXT><ACK=RCV.NXT><CTL=ACK>

> After sending the acknowledgment, drop the unacceptable segment and return.

dnadlinger:
(I guess in the definition of [receive window](https://datatracker.ietf.org/doc/html/rfc9293#section-4-1.62), only segments entirely outside of the window are called outright invalid, so receiving it partially still is probably the way to go.)

lrh2000:
[ RFC9293: 3.4. Sequence Numbers ](https://datatracker.ietf.org/doc/html/rfc9293#section-3.4)
>> Determining that an incoming segment contains sequence numbers that are expected (i.e., that the segment **"overlaps"** the receive window).
>> [..]
>> The first part of this test checks to see if the beginning of the segment falls in the window, the second part of the test checks to see if the end of the segment falls in the window; **if the segment passes either part of the test, it contains data in the window.**

So I think the implementation in this PR is correct.

> only segments entirely outside of the window are called outright invalid

Yes. Although I failed to find a sentence that explicitly states it, the quoted wording above does imply this case.

dnadlinger:
I couldn't find an explicit statement of that either, but yes, it matches my interpretation too.

ibigbug:
would like to see this checked in 👍🏻

PR #1087, opened 2025-09-27
Add generic IPv4/IPv6 packet parsing support to `IpRepr`

jeff-moon:
- Add generic Packet type for handling both IPv4 and IPv6 packets
- Add parsing functionality for protocol-agnostic IP packet handling
- Add tests for new parsing functionality

This also adds `ChecksumCapabilities` into the generic parse, since it's needed by IPv4, which is something that I am (and probably you will be) iffy about. This would clean up a lot of our code that generically handles IPv4/IPv6 code.

### DIFF
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -533,6 +533,60 @@ impl From<Ipv6Repr> for Repr {
     }
 }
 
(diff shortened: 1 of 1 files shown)


thvdveld:
I don't think it's _that_ iffy. Looks fine to me!

PR #1088, opened 2025-10-02
switch AtomicUsize to AtomicU64 in example now that it's stable

JPDye:
Minor thing, just getting my feet wet. Switch to AtomicU64 for `Clock` in example and fuzzing, inline with the comments left in the code.


thvdveld:
Thank you!

Issue #1071, opened 2025-07-09
ICMP errors (such as destination unreachable) should reset TCP socket


thomaseizinger:
ICMP errors contain at least 64bit of the original packet in their payload. This is enough to identify TCP connections. When receiving an ICMP error for one of its sockets, `smoltcp` should reset that socket state.

thvdveld:
I think this is something we let the user decide on. You can make an ICMP socket bound to the same endpoint as your UDP socket (TCP is not implemented yet, but should be trivial). In your application, you should check if this ICMP socket received an ICMP message. Currently, this socket accepts Destination Unreachable and Time Exceeded errors. When receiving such a message in the socket, the user can decide how to react to it.

PR #1089, opened 2025-10-07
Support binding ICMP socket to TCP port

thvdveld:
This allows to bind the ICMP socket to a TCP port as well, in order to receive ICMP errors associated to TCP packets sent from that port.

Using the ICMP socket, the TCP socket could be reset when receiving an IMCP error related to that socket (#1071).

### DIFF
--- a/src/socket/icmp.rs
+++ b/src/socket/icmp.rs
@@ -14,6 +14,7 @@ use crate::wire::{Icmpv4Packet, Icmpv4Repr, Ipv4Repr};
 #[cfg(feature = "proto-ipv6")]
 use crate::wire::{Icmpv6Packet, Icmpv6Repr, Ipv6Repr};
 use crate::wire::{IpAddress, IpListenEndpoint, IpProtocol, IpRepr};
+use crate::wire::{TcpPacket, TcpRepr};
 use crate::wire::{UdpPacket, UdpRepr};
 
 /// Error returned by [`Socket::bind`]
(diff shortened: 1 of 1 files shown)


Dirbaio:
I think it'd be better to make smoltcp itself handle these ICMP messages and reset the socket. Possibly behind a Cargo feature.

Requiring the user to set up the ICMP socket themselves is quite burdensome:
- It makes the user set up the socket, bind it, manage its state, poll it for packets, then parse the packet themselves and act accordingly. It's quite a bit of code that will be essentially the same for all projects that want this.
- you need one ICMP socket for each TCP socket you have, and each ICMP socket needs enough buffer space to hold one packet. It's quite a bit of wasted RAM.

thvdveld:
I was thinking about doing it internally in smoltcp, but just quickly added this now such that _if_ someone wants to do it using the ICMP socket, which we already allow for UDP, then they can do so with this PR.

When doing it internally, we shouldn't do it with a feature flag. It would result in a conflict when two crates depend on smoltcp, one using the feature flag and one not using the feature flag, and the latter actually depending on smoltcp *not* to handle the error and reset the socket.

Dirbaio:
> the latter actually depending on smoltcp not to handle the error and reset the socket.

This seems a bit unrealistic? in what use case would you want it to *actively not* handle the ICMP errors in a way that breaks if you do?

anyway, I was thinking of the Cargo feature mostly to avoid increased code size when you don't need it. We can make a Cargo feature *and* a flag you set in each socket indivudually.

thvdveld:
A custom protocol might handle errors differently. Using a flag in the socket itself solves this. In any case, I totally agree we should do it internally.

I think this PR already provides some way of handling ICMP errors for TCP, not ideal, but allowing to do the same as for UDP. Doing it internally could be a follow up PR.

PR #1099, opened 2025-11-07
Fix #1095 preserve all fixed ipv4 fields

### LINKED ISSUE #1095: Sending a raw IP packet drops some fields from original IP header
When sending an IP packet using a raw socket, the `dispatch` function in `src/socket/raw.rs` is called. This function parses the outgoing IP header into an `Ipv4Repr` to ensure its a valid packet, and then rebuilds that header using `emit`. Since `Ipv4Repr` is missing some fields (for example, ECN), any values that were set in the original header are dropped when the `Ipv4Repr` is rebuilt into the outgoing header. I imagine similar behavior exists for `Ipv6Repr`, but haven't investigated further.

aaronddavies:
This PR preserves the IPv4 fixed length fields that were being dropped. They are now parsed directly into the `Ipv4Repr` struct. 

The exception is header length, which is always 20 bytes, until options support is implemented. Options is a variable length field that will be implemented in a separate PR.

Bench tests

```
develop @ 39cd44e
test wire::bench_emit_ipv4 ... bench:           8.07 ns/iter (+/- 0.18)
test wire::bench_emit_ipv6 ... bench:           1.12 ns/iter (+/- 0.01)
test wire::bench_emit_tcp  ... bench:          30.81 ns/iter (+/- 0.85)
test wire::bench_emit_udp  ... bench:          26.34 ns/iter (+/- 11.85)

1095-preserve-all-fixed-ipv4-fields @ 9c2b2c3
test wire::bench_emit_ipv4 ... bench:           8.52 ns/iter (+/- 0.19)
test wire::bench_emit_ipv6 ... bench:           1.13 ns/iter (+/- 0.03)
test wire::bench_emit_tcp  ... bench:          31.88 ns/iter (+/- 0.73)
test wire::bench_emit_udp  ... bench:          26.35 ns/iter (+/- 1.32)
```


Dirbaio:
i'm going to close this in favor of #1108 since that one seems more complete.

PR #1103, opened 2025-11-16
Simplify tcp::Socket::can_recv

smaeul:
This function does not actually care about the socket state, so there is no need to check it. The ability to immediately receive data only depends on there being something in the receive buffer.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1185,7 +1185,7 @@ impl<'a> Socket<'a> {
             // we still can receive indefinitely.
             State::FinWait1 | State::FinWait2 => true,
             // If we have something in the receive buffer, we can receive that.
-            _ if !self.rx_buffer.is_empty() => true,
+            _ if self.can_recv() => true,
             _ => false,
         }
     }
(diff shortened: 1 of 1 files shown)


Dirbaio:
This is changing the behavior of `can_recv`. Could you explain why do you think the new behavior is better?

smaeul:
It doesn't change the behavior. If you inline the current `may_recv`, the expression for `can_recv` expands to:

```rust
(
    matches!(self.state, State::Established | State::FinWait1 | State::FinWait2) ||
    !self.rx_buffer.is_empty()
) && !self.rx_buffer.is_empty()
```

The value of the `matches!` subexpression has no effect on the value of the whole expression. The outer expression is either "true && true" or "anything && false", so it always takes the value of the right side (`!self.rx_buffer.is_empty()`).

It took me a bit to parse this originally, so I wanted to make it simpler for future readers.

Dirbaio:
it's possible for some data to arrive, and then receive a FIN before the user reads the data out of the buffer. The FIN immediately changes the socket to CLOSE_WAIT but the data is still in the buffer available for being read. In this case the old `can_recv` would return `false` and the new would return `true`.

smaeul:
Wouldn't that example hit the `_ if !self.rx_buffer.is_empty() => true,` case in the old `may_recv` and thus return true from `can_recv`?

Dirbaio:
ah damn you're right, I was reading the expression wrong.

PR #1108, opened 2025-12-08
[feature] Fix #1095 Implement full IPv4 header and fragmentation

### LINKED ISSUE #1095: Sending a raw IP packet drops some fields from original IP header

aaronddavies:
Fixes #1095.

Changes to operational code:
* **Add the missing IPv4 fixed length fields**. They are now a part of the `Ipv4Repr` struct so that they are no longer dropped when message traffic goes through smoltcp. (**TODO** - this is being updated to pass the bytes to raw sockets only, and not augment the `Repr`.)
* **Add support for** `options`. This implemented as a fixed size byte array to the `IPv4Repr` struct. The header length field determines how many bytes of the options array are to be included when emitting from the struct. Bench tests showed that a the raw array had better performance than using `Option<[u8]>`. (**TODO** - see above.)
* **Filter options across fragments** for outgoing packets.
  1. Options are persisted across fragments if the copy bit in the type octet is set.
  2. Option lengths are preserved according to the length octet following the type octet, with the exception of padding octets.
* **Capture the options** from an incoming packet. The options of the first incoming fragment of a packet stream are captured as the representative set of options for the final assembled packet, which is to be delivered to the sockets. NOTE: An options capture error in this process will cause the entire packet to be dropped.

Changes to unit tests:
* Update all test IPv4 repr definitions to include the init of the rest of the header fields. This boilerplate is a significant portion of the line changes in this PR. (**TODO** -  see above, this will no longer be necessary.)
* Fix unit tests that don't require fragmentation to not specify fragmentation in their test packet headers
* Use a config-if block for choosing the right `IpvX` struct repr (**TODO** -  see above, this will no longer be necessary.)
* Fix some places to only import related modules if features require them
* Add unit tests for the new changes listed above

Reference:
1. [RFC 791](https://datatracker.ietf.org/doc/html/rfc791)
2. [IPv4 Parameters - Options](https://www.iana.org/assignments/ip-parameters/ip-parameters.xhtml#ip-parameters-1)

Bench tests before & after

```
current HEAD main (39cd44e)
test wire::bench_emit_ipv4 ... bench:           8.27 ns/iter (+/- 0.52)
test wire::bench_emit_ipv6 ... bench:           1.11 ns/iter (+/- 0.01)
test wire::bench_emit_tcp  ... bench:          29.89 ns/iter (+/- 0.20)
test wire::bench_emit_udp  ... bench:          25.04 ns/iter (+/- 0.18)


this branch (1bce638)
test wire::bench_emit_ipv4 ... bench:           9.44 ns/iter (+/- 0.54)
test wire::bench_emit_ipv6 ... bench:           1.11 ns/iter (+/- 0.01)
test wire::bench_emit_tcp  ... bench:          29.90 ns/iter (+/- 0.40)
test wire::bench_emit_udp  ... bench:          25.06 ns/iter (+/- 0.44)
```

### DIFF
--- a/src/iface/fragmentation.rs
+++ b/src/iface/fragmentation.rs
@@ -9,7 +9,15 @@ use crate::storage::Assembler;
 use crate::time::{Duration, Instant};
 use crate::wire::*;
 
+use crate::iface::interface::DispatchError;
+use crate::phy::ChecksumCapabilities;
+#[cfg(feature = "proto-ipv4")]
+use crate::wire::ipv4::{ALIGNMENT_32_BITS, HEADER_LEN, MAX_OPTIONS_SIZE, Packet, Repr};
 use core::result::Result;
+
+// Special option type octets.
+const OPTION_TYPE_PADDING: u8 = 0x00;
+const OPTION_TYPE_NO_OPERATION: u8 = 0x01;
 
 #[cfg(feature = "alloc")]
 type Buffer = alloc::vec::Vec<u8>;
@@ -299,6 +307,10 @@ pub(crate) struct Ipv4Fragmenter {
     pub frag_offset: u16,
     /// The identifier of the stream.
     pub ident: u16,
+    /// The header options.
+    pub options_buffer: [u8; MAX_OPTIONS_SIZE],
+    /// Actual length of the options.
+    pub options_len: usize,
 }
 
 #[cfg(feature = "proto-sixlowpan-fragmentation")]
@@ -339,6 +351,8 @@ impl Fragmenter {
                 dst_hardware_addr: EthernetAddress::default(),
                 frag_offset: 0,
                 ident: 0,
+                options_buffer: [0u8; MAX_OPTIONS_SIZE],
+                options_len: 0,
             },
 
             #[cfg(feature = "proto-sixlowpan-fragmentation")]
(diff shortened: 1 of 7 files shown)


Dirbaio:
Thanks for the PR! 

I'm concerned adding all these fields and options to `Ipv4Repr` will increase memory usage. `Ipv4Repr` intentionally includes only the fields smoltcp cares about when emitting packets. Do you think it's possible to fix this by avoiding roundtrips through `Ipv4Repr`? ie avoiding parsing+emitting in the first place, just copying bytes. 

> Fragment packet data in multiples of 8 octets according to RFC 791. The fragment offset for packet fragments is measured in units of 8 octets, therefore payloads must be fragmented in multiples of 8. This adds logic in two places: 1) where the first fragment is sent immediately, and 2) where each successive fragment is sent from the queue.

This is a nice bug fix. could you send a PR with just this change?

aaronddavies:
@Dirbaio thank you for the review and feedback! That's a good point and I will get the bugfix into a separate PR. 

I will also look at carrying around the original bytes without augmenting the repr.

aaronddavies:
> > Fragment packet data in multiples of 8 octets according to RFC 791. The fragment offset for packet fragments is measured in units of 8 octets, therefore payloads must be fragmented in multiples of 8. This adds logic in two places: 1) where the first fragment is sent immediately, and 2) where each successive fragment is sent from the queue.
> 
> This is a nice bug fix. could you send a PR with just this change?

Opened https://github.com/smoltcp-rs/smoltcp/pull/1116.

PR #1111, opened 2026-01-26
BSD-like  Silly Window  Syndrome Avoidance

JPDye:
Implementing SWS avoidance using the BSD approach (also copied by Linux).

- [X] Receiver-side SWS avoidance
- [X] Fix failing receiver-side tests
- [X] Sender-side SWS avoidance
- [X] Override timer for sender-side SWS avoidance
- [X] Fix failing sender-side tests
- [ ] Remove duplicate calculations from `seq_to_transmit()` and `send_is_blocked()`
- [ ] `SWSOverrideTimer` based on RTT?
- [ ] More tests

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -2321,12 +2321,29 @@ impl<'a> Socket<'a> {
         }
     }
 
-    /// Return whether we should send ACK immediately due to significant window updates.
+    /// Return whether we should send an ACK to update our advertised window.
     ///
-    /// ACKs with significant window updates should be sent immediately to let the sender know that
-    /// more data can be sent. According to the Linux kernel implementation, "significant" means
-    /// doubling the receive window. The Linux kernel implementation can be found at
-    /// <https://elixir.bootlin.com/linux/v6.9.9/source/net/ipv4/tcp.c#L1472>.
+    /// Implements BSD-style Silly Window Syndrome avoidance to prevent advertising
+    /// tiny windows and to reduce unnecessary window update traffic.
+    ///
+    /// The algorithm sends a window update when EITHER:
+    ///
+    /// 1. The window increased by at least 2 * MSS and one of the following is true:
+    ///    - The increase is >= 25% of buffer capacity (significant growth)
+    ///    - The window is <= 12.5% of buffer capacity (nearly closed)
+    ///    - The MSS is >= 12.5% of buffer capacity (small buffer)
+    ///
+    /// 2. The window increased by at least 50% of buffer capacity (always significant)
+    ///
+    /// Case 1 handles constrained scenarios where aggressive updates reduce unnecessary latency.
+    /// Case 2 handles normal operation where rarely changing the window size helps header prediction.
+    ///
+    /// Updates are NEVER sent when a delayed ACK is pending as it will piggyback on that ACK. This
+    /// is enforced in `dispatch()` and `poll_at()`. Not true anymore. FIX.
+    ///
+    /// For details, see:
+    /// - <https://elixir.bootlin.com/linux/v6.9.9/source/net/ipv4/tcp.c#L190>
+    /// - <https://cgit.freebsd.org/src/tree/sys/netinet/tcp_output.c?h=stable/15#n627>
     fn window_to_update(&self) -> bool {
         match self.state {
             State::SynSent
@@ -2334,9 +2351,17 @@ impl<'a> Socket<'a> {
             | State::Established
             | State::FinWait1
             | State::FinWait2 => {
-                let new_win = self.scaled_window();
                 if let Some(last_win) = self.last_scaled_window() {
-                    new_win > 0 && new_win / 2 >= last_win
+                    let window_len = self.rx_buffer.window();
+                    let window_capacity = self.rx_buffer.capacity();
+                    let last_window_len = (last_win as usize) << self.remote_win_shift;
+                    let window_increase = window_len.saturating_sub(last_window_len);
+
+                    (window_increase >= 2 * self.remote_mss
+                        && (window_increase >= window_capacity / 4
+                            || window_len <= window_capacity / 8
+                            || self.remote_mss >= window_capacity / 8))
+                        || window_increase >= window_capacity / 2
                 } else {
                     false
                 }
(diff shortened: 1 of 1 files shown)


JPDye:
Not sure about the `recv_out_of_recv_win()` test. Previously the final ACK was sent immediately, now it's delayed. Probably need to fix that.

JPDye:
Obviously need to fix the tests.

More difficult is the approach to take with the override timeout that [RFC 9293 requires](https://www.rfc-editor.org/rfc/rfc9293.html#section-3.8.6.2.1-8.4). It says it can be combined with the ZWP timer but it's clearer to do it as a seperate timer I guess. Issue is where to do the setting of the timer, could do it by letting `seq_to_transmit()` take `&mut self` but I don't think that's very clean/inline with existing approach.

JPDye:
Need to add additional tests for the SWS behaviour.

Should SWS avoidance be toggle-able like nagles?

Edit: Now that I've _finally_ got the netsim test working and see the throughput numbers, it would make sense to make SWS optional.

Dirbaio:
> Edit: Now that I've finally got the netsim test working and see the throughput numbers, it would make sense to make SWS optional.

why? it shouldn't hurt throughput much

JPDye:
> why? it shouldn't hurt throughput much

```
╭─────┬─────────┬─────────┬─────────┬─────────┬──────────┬──────────┬──────────┬──────────┬──────────╮
│   # │   buf   │  0.000  │  0.001  │  0.010  │  0.020   │  0.050   │  0.100   │  0.200   │  0.300   │
├─────┼─────────┼─────────┼─────────┼─────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
│   0 │     128 │ -9.1%   │ -8.9%   │ -7.5%   │ -6.4%    │ -3.8%    │ -1.6%    │ -0.2%    │ +0%      │
│   1 │     256 │ -9.1%   │ -8.9%   │ -7.5%   │ -6.4%    │ -3.8%    │ -1.6%    │ -0.2%    │ +0%      │
│   2 │     512 │ -9.1%   │ -8.9%   │ -7.5%   │ -6.3%    │ -3.8%    │ -1.6%    │ -0.2%    │ +-0%     │
│   3 │    1024 │ -9.1%   │ -8.9%   │ -7.5%   │ -6.4%    │ -3.6%    │ -1.6%    │ -0.2%    │ +-0%     │
│   4 │    2048 │ -4.8%   │ -4.8%   │ -2.8%   │ -3.3%    │ +1%      │ +7.3%    │ +0.8%    │ +9.4%    │
│   5 │    4096 │ +0%     │ +0%     │ -0.3%   │ +0.2%    │ +4.9%    │ -1.2%    │ +2.4%    │ +13.4%   │
│   6 │    8192 │ +0%     │ +1.6%   │ +7.5%   │ +10.7%   │ +14.5%   │ +3.3%    │ +10.6%   │ -3.8%    │
│   7 │   16384 │ +0%     │ +0.2%   │ +1.8%   │ +1.9%    │ +23.7%   │ -7.3%    │ -6.8%    │ -16.7%   │
│   8 │   32768 │ +0%     │ +0.4%   │ +1.5%   │ -7.9%    │ +20.8%   │ -24.4%   │ -5%      │ -12.4%   │
╰─────┴─────────┴─────────┴─────────┴─────────┴──────────┴──────────┴──────────┴──────────┴──────────
```
If you know the network you're on and the hosts you're communicating with, SWS avoidance perhaps isn't necessary. Especially in an embedded context where the window is always "silly" (i.e. tiny buffers). TBF, I know nothing about embedded use-cases so if you're telling me it's not needed then that's cool.

Probably worth playing with the thresholds in `window_to_update()`, I mostly took them from the BSD source and so they're probably not be optimal for an embedded stack. 

Also going to test if window updates need to be blocked by delayed ACKs. SWS means we'd only ever announce an update if it was significant, so delaying it to piggyback on a delayed ACK might not be optimal for throughput.

EDIT:
```
╭─────┬─────────┬─────────┬─────────┬─────────┬──────────┬──────────┬──────────┬──────────┬──────────╮
│   # │   buf   │  0.000  │  0.001  │  0.010  │  0.020   │  0.050   │  0.100   │  0.200   │  0.300   │
├─────┼─────────┼─────────┼─────────┼─────────┼──────────┼──────────┼──────────┼──────────┼──────────┤
│   0 │     128 │ +0%     │ +0%     │ +0%     │ +0%      │ +0%      │ +0%      │ +0%      │ +0%      │
│   1 │     256 │ +0%     │ +0%     │ +0%     │ +0%      │ +0%      │ +0%      │ +0%      │ +0%      │
│   2 │     512 │ +0%     │ +0%     │ +0%     │ +0%      │ +0%      │ +0%      │ +0%      │ +0%      │
│   3 │    1024 │ +0%     │ +0%     │ +0%     │ +0%      │ +0%      │ +0%      │ +0%      │ +0%      │
│   4 │    2048 │ +0%     │ +0%     │ +0%     │ +0%      │ +0%      │ +0%      │ +-0%     │ +0%      │
│   5 │    4096 │ +0%     │ +0%     │ +0%     │ +0%      │ +0%      │ +-0%     │ +-0%     │ +-0%     │
│   6 │    8192 │ +0%     │ +2.9%   │ +9%     │ +10.4%   │ +11.9%   │ +12.5%   │ +5.1%    │ +3.3%    │
│   7 │   16384 │ +0%     │ +0.1%   │ +1.7%   │ +0%      │ +6.1%    │ -1.9%    │ -8.4%    │ +14.9%   │
│   8 │   32768 │ +0%     │ +0.7%   │ +3.5%   │ -8.7%    │ +15.9%   │ +23.5%   │ +91.5%   │ -19.2%   │
╰─────┴─────────┴─────────┴─────────┴─────────┴──────────┴──────────┴──────────┴──────────┴──────────╯
```

All green if I never delay window updates to piggyback on the delayed ACK. The BSD implementation that I'm kinda copying is very explicit about not doing this so need to see how it affects number of segments sent/average segment size/etc to decide if it's worth it.

JPDye:
Pretty happy with it as it is. Sending the immediate window update (even when delayed ACK is pending) seems to give much nicer performance on the netsim and I didn't notice anything significant in terms of small segments, etc....

Need to write some SWS specific tests and refactor `seq_to_transmit()` and `send_is_blocked()` to reuse their calculated values.

PR #1113, opened 2026-01-30
tcp: close socket if local ip is no longer ours.

### LINKED ISSUE #466: TCP sockets can transmit packets with wrong source IP

Dirbaio:
Fixes #466

Alternatiev to #970

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -814,6 +814,12 @@ impl InterfaceInner {
         self.now = now
     }
 
+    #[cfg(test)]
+    #[allow(unused)] // unused depending on which sockets are enabled
+    pub(crate) fn set_ip_addrs(&mut self, addrs: Vec<IpCidr, IFACE_MAX_ADDR_COUNT>) {
+        self.ip_addrs = addrs;
+    }
+
     #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
     fn check_hardware_addr(addr: &HardwareAddress) {
         if !addr.is_unicast() {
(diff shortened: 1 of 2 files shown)


cvengler:
Unfortunately, this broke our (the Tor VPN project) setup: https://gitlab.torproject.org/tpo/core/onionmasq/-/issues/175

(It is a bit unfortunate that we are using the git main branch but we really need a few features that are in there, that are not in 0.12)

I am currently investigating this behavior as I am unsure on whether it stems from this patch or an incorrect API usage from our end.

cvengler:
Our relevant part in the code is as follows:

```rust
        let mut iface = Interface::new(iface_config, &mut device, Instant::now());

        iface.set_any_ip(true);

        // FIXME(eta): I think we just need to insert these to make packets flow, and they aren't
        //             actually relevant for anything. This assumption should probably be checked.
        #[allow(clippy::expect_used)]
        iface
            .routes_mut()
            .add_default_ipv4_route(Ipv4Address::new(0, 0, 0, 1))
            .expect("smoltcp routing table full?");
        iface.update_ip_addrs(|ip_addrs| {
            #[allow(clippy::expect_used)]
            ip_addrs
                .push(IpCidr::new(IpAddress::v4(0, 0, 0, 1), 0))
                .expect("smoltcp IP address vector not big enough?")
        });
```

I assume having `set_any_ip` to true should allow this.

Issue #1104, opened 2025-11-18
Fragment payload size should be a multiple of eight octets


aaronddavies:
**Expected behavior:**

Per RFC 791 3.1, the fragment offset for packet fragments is measured in units of 8 octets (64 bits). Therefore payloads must be fragmented in multiples of 8 (see also the discussion in RFC 791 3.2).

**Actual behavior:**

Fragments are not aligned to 8-octets, but are as large as the maximum transmission unit allows for the packet. This can cause problems when trying to determine where a received packet's payload actually begins and ends.

aaronddavies:
Done per #1116.

PR #1116, opened 2026-01-31
bugfix: Fix fragment payload sizes to be a multiple of eight octets (#1104)

aaronddavies:
This PR forces the fragment packet data size to be a multiple of 8 octets according to RFC 791. The fragment offset for the packet fragments is measured in units of 8 octets, therefore payloads must be fragmented in multiples of 8. This adds logic in two places: 1) where the first fragment is sent immediately, and 2) where each successive fragment is sent from the queue. The existing unit test is enhanced to check both of these paths.

This fix is isolated from its original inclusion in https://github.com/smoltcp-rs/smoltcp/pull/1108.


Dirbaio:
thanks!

PR #1119, opened 2026-02-03
iface: Honor any_ip when checking local address

cvengler:
#1113 introduced a fix to discard packets from interfaces with non-matching source addresses.  This is a totally valid behavior in most circumstances.  However, when the `any_ip` feature is enabled, an interface should not bother with checking the source IP addresses, as the feature is described as follows:

> AnyIP allowins packets to be received locally on IP addresses other
> than the interface’s configured ip_addrs.

See #466
See <https://gitlab.torproject.org/tpo/core/onionmasq/-/issues/175>

### DIFF
--- a/src/iface/interface/ipv4.rs
+++ b/src/iface/interface/ipv4.rs
@@ -187,12 +187,6 @@ impl InterfaceInner {
             && !self.is_broadcast_v4(ipv4_repr.dst_addr)
         {
             // Ignore IP packets not directed at us, or broadcast, or any of the multicast groups.
-            // If AnyIP is enabled, also check if the packet is routed locally.
-
-            if !self.any_ip {
-                net_trace!("Rejecting IPv4 packet; any_ip=false");
-                return None;
-            }
 
             if !ipv4_repr.dst_addr.x_is_unicast() {
                 net_trace!(
@@ -211,6 +205,9 @@ impl InterfaceInner {
 
                 return None;
             }
+
+            net_trace!("Rejecting IPv4 packet; no assigned address");
+            return None;
         }
 
         #[cfg(feature = "medium-ethernet")]
@@ -272,7 +269,7 @@ impl InterfaceInner {
                 ..
             } => {
                 // Only process ARP packets for us.
-                if !self.has_ip_addr(target_protocol_addr) && !self.any_ip {
+                if !self.has_ip_addr(target_protocol_addr) {
                     return None;
                 }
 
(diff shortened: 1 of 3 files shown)


whitequark (approved):
Thanks!

whitequark:
I think these checks are now made redundant:

https://github.com/smoltcp-rs/smoltcp/blob/2aa413d2f26ed4023d1f8672d7f9055167f973e7/src/iface/interface/ipv4.rs#L192-L195

https://github.com/smoltcp-rs/smoltcp/blob/2aa413d2f26ed4023d1f8672d7f9055167f973e7/src/iface/interface/ipv4.rs#L274-L277

https://github.com/smoltcp-rs/smoltcp/blob/2aa413d2f26ed4023d1f8672d7f9055167f973e7/src/iface/interface/ipv6.rs#L217-L220

whitequark:
In other words, your PR entirely obviates the need for checking `any_ip` elsewhere, which I really like. Please remove the checks.

cvengler:
> In other words, your PR entirely obviates the need for checking `any_ip` elsewhere, which I really like. Please remove the checks.

done.

cvengler:
Given the CI failures, I don't think they are redundant, especially because they appear for ARP, IPv4 and IPv6

whitequark:
Looking into it.

src/iface/interface/ipv4.rs
```diff
@@ -271,11 +265,6 @@ impl InterfaceInner {
                 target_protocol_addr,
                 ..
             } => {
-                // Only process ARP packets for us.
-                if !self.has_ip_addr(target_protocol_addr) && !self.any_ip {
-                    return None;
-                }
```
whitequark, line 277: Here, just the `&& !self.any_ip` needs to be removed.

src/iface/interface/ipv4.rs
whitequark, line 207:
```suggestion
            }

        net_trace!("Rejecting IPv4 packet; no assigned address");
        return None;
```

src/iface/interface/ipv6.rs
whitequark, line 233:
```suggestion
            }

        net_trace!("Rejecting IPv6 packet; no assigned address");
        return None;
```

whitequark (changes requested):
Sorry, "delete the checks" was too broad. I've specified what I think is the right fix in the review comments.

src/iface/interface/ipv4.rs
cvengler: Oh indeed. Was a bit quirky there.

cvengler:
@whitequark fixed

PR #1120, opened 2026-02-09
Rely on TCP SYN retransmission to handle a larger number of incoming connections than the number of available sockets

ivmarkov:
`smoltcp` (and `embassy-net`) often operate with just a handful of TCP sockets due to memory constraints.

This poses a problem when implementing something like [an HTTP server on top](https://github.com/sysgrok/edge-net/tree/master/edge-http) in that - typically - there would be only a few allocated sockets and - most often than not - **all** these sockets might be busy serving already accepted HTTP connections.

In that hypothetical scenario, what happens when a new TCP SYN packet arrives is that it is immediately refused with a TCP RST request, so the other peer (a browser, or curl or whatever) ends up with an error.

However, this is not ideal because there is a high likelihood that the "busy" sockets will - sooner rather than later - become available again for accepting new incoming connections.

One way to resolve this issue is to just have much more sockets than HTTP connection handlers, but this does not come free in terms of resources, as each socket needs a TX + RX buffer (granted, these could be small-ish, but still!). I.e. maintaining a user-level SYN acceptance queue of sorts.

Another way - which is what this PR does is to just **not** answer the incoming SYN request **at all** and rely on the other peer to re-transmit (with a back-off!) the SYN request. Where the assumption is that _eventually_ some of the busy sockets will become free during the retransmission, and the SYN packet would get accepted.

So in a way, we emulate a SYN acceptance queue by (ab)using TCP retransmission logic.

This specifically solves this issue in edge-http: https://github.com/sysgrok/edge-net/issues/80

### DIFF
--- a/src/iface/interface/tcp.rs
+++ b/src/iface/interface/tcp.rs
@@ -39,6 +39,19 @@ impl InterfaceInner {
             // Never send a TCP RST packet with unspecified addresses.
             // Never send a TCP RST when packet has been handled by raw socket.
             None
+        } else if tcp_repr.control == TcpControl::Syn {
+            // The packet is trying to open a new connection, but we don't have a socket listening on the
+            // destination port. In this case, we should just ignore the packet instead of sending a TCP RST packet.
+            //
+            // The idea is to keep the other peer re-sending the SYN packet until eventually we get a new socket
+            // listening on the destination port, at which point the connection can be established.
+            //
+            // This allows for emulating a SYN backlog, without actually having to allocate any resources for it.
+            // A typical example would be an HTTP server that is operating off from a limited number of sockets
+            // and is currently busy in that all sockets are handling requests, but the server would _still_ want
+            //  to allow new connections to be established as soon as a socket becomes available, rather than
+            // those being dropped on the floor because the SYN packet was responded to with a TCP RST packet.
+            None
         } else {
             // The packet wasn't handled by a socket, send a TCP RST packet.
             let (ip, tcp) = tcp::Socket::rst_reply(&ip_repr, &tcp_repr);


ivmarkov:
NOTE: Tests might need fixing, but before/while I'm doing this, the question remains whether this is a legit approach, or there could be issues with it.

whitequark (commented):
What happens if there is no socket planning to listen on the port?

ivmarkov:
> What happens if there is no socket planning to listen on the port?

I guess the other party would eventually give up re-transmitting the SYN packet after some time.

It is a bit of a "what is the lesser evil" question I think:
- Actively denying SYN packets and thus having to emulate a heavyweight acceptance queue in the HTTP server
- Or never responding to a SYN packet (EDIT: if there are no sockets accepting it) and thus (maybe?) not following strictly the TCP protocol spec, though I have to double-check what other TCP implementations do in that case either

ivmarkov:
Maybe a relevant article: https://blog.cloudflare.com/syn-packet-handling-in-the-wild/

To quote from the article (emphasis mine):
"
What happens if the application can't keep up with calling accept() fast enough?

This is when the magic happens! When the Accept Queue gets full (is of a size of backlog+1) then:

**Inbound SYN packets to the SYN Queue are dropped.**

**Inbound ACK packets to the SYN Queue are dropped.**

The TcpExtListenOverflows / LINUX_MIB_LISTENOVERFLOWS counter is incremented.

The TcpExtListenDrops / LINUX_MIB_LISTENDROPS counter is incremented.

**There is a strong rationale for dropping inbound packets: it's a push-back mechanism. The other party will sooner or later resend the SYN or ACK packets by which point, the hope is, the slow application will have recovered.**

This is a desirable behavior for almost all servers. For completeness: it can be adjusted with the global net.ipv4.tcp_abort_on_overflow toggle, but better not touch it.
"

So what I'm (trying to do) here is basically the same, except not having an explicit SYN queue at all.

ivmarkov:
> What happens if there is no socket planning to listen on the port?

To address your question in more detail, I think nothing stops us from modeling an "I plan to listen on port XYZ even if I don't have an active socket on it right now" registration/deregistration metaphor.

But this would only make sense IF the idea of not answering a SYN packet when there is no socket to accept it is legid in the first place.

If you say "yes", but "you need to model the "I plan to listen on THAT port" thing" then I guess I can model that too. Though that might be quite some lift and shift and maybe API changes which are not worth it?

whitequark:
Your observations are correct. I think the direction is promising but only if we don't break the existing behavior, and modeling the new state would be necessary.

ivmarkov:
> Your observations are correct. I think the direction is promising but only if we don't break the existing behavior, and modeling the new state would be necessary.

OK (and sorry if I'm a bit annoying), but to be completely clear so that I don't jump the guns without fully understanding:
- If I implement the "I plan to listen on port XYZ" registration/deregistration API metaphor, THEN not answering SYN packets for THAT port if there are currently no active sockets in the socket set on THAT port would be accepted as "OK"

Right?

whitequark:
Yep that's correct. (Aand thank you for double checking, I appreciate it!)

I *think* it could be as easy as "a flag on a socket that says whether it is a part of a 'flexible SYN queue' on this port", which causes the behavior to change from RST to drop.

Dirbaio:
I think the better solution for this is to officially have a concept of "TCP accept queue" or "TCP listener". See https://github.com/smoltcp-rs/smoltcp/pull/1118#issuecomment-3841213993

The accept queue would be able to buffer many more SYNs than you have sockets, so clients would get the ACK immediately when a socket becomes available and wouldn't have to rely on retransmisison.

Only if the queue gets full it would drop the SYNs (and wouldn't reply with RST) so it degrades to rely on retransmission. We'd still reply with RST if there's no TCP listener on that port.

whitequark:
Yeah, you are right. It is more invasive but in the long term much better.

ivmarkov:
> I think the better solution for this is to officially have a concept of "TCP accept queue" or "TCP listener". See [#1118 (comment)](https://github.com/smoltcp-rs/smoltcp/pull/1118#issuecomment-3841213993)
> 
> The accept queue would be able to buffer many more SYNs than you have sockets, so clients would get the ACK immediately when a socket becomes available and wouldn't have to rely on retransmisison.
> 
> Only if the queue gets full it would drop the SYNs (and wouldn't reply with RST) so it degrades to rely on retransmission. We'd still reply with RST if there's no TCP listener on that port.

The good thing is, the API requested here **anyway** requires some notion of a "socket listener". See above where I talk about the "I plan to listen on port XYZ" registration/deregistration API metaphor.

So in a way, we are implementing a socket listener no matter what. It is just that **why** do we really need a listener with a SYN queue > 0?

What would that solve? Are we so worried that clients have to re-transmit? Well, they should, smoltcp is a resource-constrainted API after all.

I'm also worried a bit about the "how do we teach this stuff?" aspect of having a fixed-size queue with SYN packets, where the user has to "size" the queue. This will bubble-through all abstraction layers (smoltcp, embassy-net, edge-nal-embassy) and reach users who are already struggling with allocating them damn socket buffers.

Dirbaio:
> It is just that why do we really need a listener with a SYN queue > 0? What would that solve? Are we so worried that clients have to re-transmit? Well, they should, smoltcp is a resource-constrainted API after all.

Because this is the way TCP listening is supposed to be implemented, and the way it works in all OSs. You buffer TCP SYN packets which are just a handful of integers, and the fat socket buffers are allocated when you actually `accept()` a connection.

It gives much better latency than dropping SYNs. If an older connection becomes available in 100ms, the accept loop will pick up the next one in the queue, immediately sending the SYN+ACK back to the client. If we don't buffer SYNs the client will have to wait for at least one retransmission which is typically 1s or more.

> I'm also worried a bit about the "how do we teach this stuff?" aspect of having a fixed-size queue with SYN packets, where the user has to "size" the queue. This will bubble-through all abstraction layers (smoltcp, embassy-net, edge-nal-embassy) and reach users who are already struggling with allocating them damn socket buffers.

TCP listening on Linux works exactly like this. You create a listener, then you loop accepting incoming connections. This change would make smoltcp's API more similar to Linux, therefore easier to learn (?). 

It's our current API that's weird. I regularly get people asking how to accept mutiple TCP conns on the same port, because the answer "create multiple sockets listening on the same port" is quite non-obvious (it would give an error on Linux!).

Also if the abstractions can deal with this API shape on top of Linux surely the can deal with it on top of smoltcp.

ivmarkov:
> > I'm also worried a bit about the "how do we teach this stuff?" aspect of having a fixed-size queue with SYN packets, where the user has to "size" the queue. This will bubble-through all abstraction layers (smoltcp, embassy-net, edge-nal-embassy) and reach users who are already struggling with allocating them damn socket buffers.
> 
> TCP listening on Linux works exactly like this. You create a listener, then you loop accepting incoming connections. This change would make smoltcp's API more similar to Linux, therefore easier to learn (?).
> 
> It's our current API that's weird. I regularly get people asking how to accept mutiple TCP conns on the same port, because the answer "create multiple sockets listening on the same port" is quite non-obvious (it would give an error on Linux!).
> 
> Also if the abstractions can deal with this API shape on top of Linux surely the can deal with it on top of smoltcp.

I'm not worried about the introduction of the notion of a "TCP Listener" per se. After all, this is what folks have in Rust STD so I'm with you that this is what folks expect and the current behavior is indeed non-obvious. After all, **I** was one of those folks asking you how to  accept multiple TCP conns with smoltcp ~ 1-2yrs ago.

I am only worried about another pool (this time, for SYN packets) that needs to be sized by the user - perhaps on the upcoming "TCP Listener" thing. That's all. Because that means you need to explain to folks like me what is a SYN packet in the first place. Not that we don't have this complexity already with e.g. the UDP "metadata" buffers.

Perhaps we can call the SYN queue "an incoming connections queue" and that's all they need to know. "Improves latency, but you can set it to a small size / zero".

====

Regardless. "TCP listener" is the way to go. I need to introduce that guy first anyway. 

Once we have this, perhaps **next step** is me trying to introduce the SYN queue. 

Introducing it is not so trivial, as now `smoltcp` needs to process this queue (btw, it might be one or multiple - per tcp socket listener) and re-transmit SYN+ACKs for the SYN until it gets the final "ACK" from the other peer. A state management which is likely currently in the socket impl itself but probably has to be duplicated outside it.

whitequark:
The state duplication worries me.

ivmarkov:
> The state duplication worries me.

T.b.h. I don't know the smoltcp codebase well enough to assess how much that would be. I operate on "feelings" currently. Perhaps, indeed:
- First a queue-less TCP listener
- Next step, the SYN queue. At that step I would know the real size of the (code/state duplication) issue. In the worst case we just don't do it and live with the re-transmissions latency. I'm a lazy guy anyway.

whitequark:
Yeah, I'm happy to review a proposal.

tomkris:
I just want to point out that code change that is proposed in the PR breaks the existing behavior of the smoltcp TCP protocol in observable way. Currently TCP packets that don't hit any socket are replied with RST. With proposed change they will be blackholed. This affects how clients can observe and handle the failures (either through fast RST handling or slow timeouts).

For comparison, current smoltcp behavior matches Linux TCP/IP stack behavior in default configuration that replies with RST to incoming TCP packets that don't match any TCP socket.

```
% time curl http://127.0.0.1:12345
curl: (7) Failed to connect to 127.0.0.1 port 12345 after 0 ms: Couldn't connect to server
curl http://127.0.0.1:12345  0.00s user 0.00s system 95% cpu 0.005 total

% sudo tcpdump -i any -nn port 12345
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on any, link-type LINUX_SLL (Linux cooked), capture size 262144 bytes
23:39:03.475586 IP 127.0.0.1.46040 > 127.0.0.1.12345: Flags [S], seq 3426861327, win 65495, options [mss 65495,sackOK,TS val 953522722 ecr 0,nop,wscale 7], length 0
23:39:03.475598 IP 127.0.0.1.12345 > 127.0.0.1.46040: Flags [R.], seq 0, ack 3426861328, win 0, length 0
```

PR #1121, opened 2026-02-09
Add raw Ethernet socket support and example

xyzzyz:
Introduce a new raw Ethernet socket type behind the socket-raw-ethernet feature, with EtherType filtering and verbatim Ethernet II frame RX/TX.

Integrate raw Ethernet sockets into interface ingress/egress paths, add Ethernet-focused tests, update feature documentation, and include a TAP-based raw_ethernet example with usage guidance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,7 @@ defmt = ["dep:defmt", "heapless/defmt"]
 
 "socket" = []
 "socket-raw" = ["socket"]
+"socket-raw-ethernet" = ["socket", "medium-ethernet"]
 "socket-udp" = ["socket"]
 "socket-tcp" = ["socket"]
 "socket-tcp-pause-synack" = ["socket-tcp"]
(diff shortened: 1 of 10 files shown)


Dirbaio:
Thanks for the PR!

`socket/raw_ethernet.rs` has a lot of overlap with `socket/raw.rs`, it's the exact same public API for sending/receiving, the exact same packet buffers. It would be better if we added support for the existing raw socket to send/receive raw Ethernet frames in addition to IP frames.. This is similar to how raw sockets work on Linux etc: depending on the flags you create them with you send/receive ethernet or IP.

One problem is the current code of `socket/raw.rs` uses `IpRepr` in `accepts()`, `process()`, `dispatch()`. However, this is a design issue we need to move away from, see https://github.com/smoltcp-rs/smoltcp/pull/1108#issuecomment-3825692616

So the steps would look like:

- Change `RawSocket` so it deals with `[u8]` slices instead of parsed `IpRepr`s in `accepts()`, `process()`, `dispatch()`. `Interface` should copy those raw bytes to/from the packets on the wire, instead of parsing+reemitting.
- Change `RawSocket` so it accepts an enum of either `Ethernet` or `Ip` when binding it. `Ethernet` would contain the ethertype, `Ip` would contain the ip version + protocol.
- Change `Interface` to check whether raw sockets are IP or Eth to dispatch packets to/from them at the right layer.

xyzzyz:
That makes perfect sense; I'll talk to Aaron to coordinate work on that.

aaronddavies:
Got it - I will adjust my PR https://github.com/smoltcp-rs/smoltcp/pull/1108 according to this advice.

PR #1126, opened 2026-03-05
tcp: send challenge ACK for duplicate ACK in LAST-ACK

### LINKED ISSUE #1060: Unexpected behavior in LAST-ACK state when receiving ACK with incorrect acknowledgment numbers
I found that when the TCP socket is in the `LAST-ACK` state, smoltcp behaves incorrectly when handling ACK segments whose acknowledgement numbers do not match the expected value:

- If the client sends an ACK larger than the expected acknowledgement number, smoltcp behaves correctly by sending an ACK and then retransmitting the `FIN+ACK`.
- However, if the client sends an ACK smaller than the expected acknowledgement number, smoltcp immediately replies with a `FIN+ACK`, instead of first sending an ACK.

This seems to violate the RFC 9293. According to Section 3.10.7.4:

> "All incoming segments whose ACK value doesn't satisfy the above condition MUST be discarded and an ACK sent back." 

In the LAST-ACK state, only an ACK value that acknowledges the FIN (i.e., ACK equals SND.NXT) is valid. Therefore, in the latter case, it might be expected to discard the segment and respond with an ACK, before retransmitting the `FIN+ACK`.

HueCodes:
Per RFC 9293 §3.10.7.4, the only acceptable segment in LAST-ACK is an
acknowledgement of the local FIN. When a duplicate or stale ACK arrived
(ack_len == 0, ack_of_fin == false), smoltcp returned None with no
response, leaving the remote without feedback on the current sequence
state. The asymmetry was visible: ACKs exceeding SND.NXT already
triggered a challenge ACK via challenge_ack_reply, but ACKs below
SND.NXT did not.

Add an else-if branch in the (State::LastAck, TcpControl::None) arm
that returns a challenge ACK when ack_len == 0 and ack_of_fin == false.
A partial ACK (ack_len > 0) is left on the normal path so SND.UNA
advances and the FIN is retransmitted by the timer.

Fixes #1060

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1979,7 +1979,12 @@ impl<'a> Socket<'a> {
                     // Clear the remote endpoint, or we'll send an RST there.
                     self.set_state(State::Closed);
                     self.tuple = None;
+                } else if ack_len == 0 {
+                    // Duplicate ACK; our FIN has not been acknowledged.
+                    // Per RFC 9293 (3.10.7.4), send a challenge ACK.
+                    return self.challenge_ack_reply(cx, ip_repr, repr);
                 }
+                // Partial ACK: fall through to advance SND.UNA normally.
             }
 
             _ => {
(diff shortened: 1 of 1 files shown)


whitequark (approved):
LGTM, thanks!

PR #1129, opened 2026-03-09
feat: add default gateway check and retrieval functions

bergzand:
This extends:
- The route with methods to check if the route is a default route.
- The routing table with methods to get the default route if one is in the table.


whitequark (approved):
LGTM, thanks!

PR #1130, opened 2026-03-09
feat: add slaac_updated_at method

bergzand:
This adds a bit of functionality to know when / if the slaac state was last changed.

I'm open to alternative ways to detect if the address and route state was changed, this is the best I could come up with over the weekend.

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -697,8 +697,14 @@ impl Interface {
                 }
             });
         }
-
+        self.inner.slaac_updated = timestamp;
         self.inner.slaac.update_slaac_state(timestamp);
+    }
+
+    /// Retrieve the timestamp at which the slaac state was last updated.
+    #[cfg(feature = "proto-ipv6-slaac")]
+    pub fn slaac_updated_at(&self) -> Instant {
+        self.inner.slaac_updated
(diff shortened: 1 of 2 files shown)


whitequark:
> I'm open to alternative ways to detect if the address and route state was changed, this is the best I could come up with over the weekend.

Maybe add a "SLAAC version" that's incremented whenever it's updated? Not sure if that's more or less convenient for debugging, but this is the solution I would normally reach out for.

bergzand:
As extra datapoint this is how I'm using the timestamp with embassy-net: https://github.com/bergzand/embassy/blob/9525a1ae4117bdf19507001b5c01a5eba724b41c/embassy-net/src/lib.rs#L930. Which is pretty simple *if* the timestamp passed into the earlier `poll` call is still around.

A version would also work for me, it's slightly less convenient when the timestamp is still around, but still an relative easy comparison and I can add a wrapper for users, e.g.: `pub fn slaac_update_happened(version: u32) -> u32`

whitequark:
I don't have an argument against the timestamp (I can't think of any real-world issues with it) and I don't want to introduce a different mechanism if the end user will need a workaround for it, so let's just merge this.

whitequark (approved):
LGTM

PR #1133, opened 2026-03-31
tcp: add first-class TcpListener with syn/accept queues

Sam-nankai:
## Motivation

Closes / relates to #1118.

`tcp::Socket` handles one connection at a time.  Accepting N concurrent connections requires N pre-allocated sockets each with their own rx/tx buffers, even before any handshake completes.  There is also no real backlog: a SYN that arrives when all sockets are busy gets an RST.

This PR implements a `tcp_listener::Socket` that separates the listening phase from data transfer, following the direction discussed in #1118/#1120.

## Design

`tcp_listener::Socket` separates the listening phase from data transfer:

- **`syn_queue`** — fixed-size array of lightweight `HalfOpen` entries  (one per in-progress handshake, SYN_RECEIVED state)
- **`accept_queue`** — fixed-size array of `PendingConnection` tokens  (completed handshakes waiting for `accept()`)
- On completion, `PendingConnection` is passed to `tcp::Socket::accept()`,  which transitions directly to ESTABLISHED
- No heap allocation; both queues are caller-provided slices

Key behaviours:
- SYN-ACK retransmit with 1 s fixed interval
- Half-open entries expire after 75 s
- Backlog full → silent drop (no RST), client retries
- Accept queue full → handshake parked, promoted without extra RTT once space opens

## Testing

Six integration tests covering: basic handshake, SYN-ACK retransmit, backlog overflow, timeout/slot reuse, accept-queue-full parking.


src/socket/mod.rs
```diff
@@ -159,6 +165,8 @@ from_socket!(icmp::Socket<'a>, Icmp);
 from_socket!(udp::Socket<'a>, Udp);
 #[cfg(feature = "socket-tcp")]
 from_socket!(tcp::Socket<'a>, Tcp);
+#[cfg(feature = "socket-tcp")]
+from_socket!(tcp_listener::Socket<'a>, TcpListener);
```
whitequark: Please put it in the TCP namespace, something like `tcp::Listener<'a>`.

src/socket/tcp_listener.rs
```diff
@@ -0,0 +1,468 @@
...
+    remote_mss: u16,
+    remote_win_scale: Option<u8>,
+    remote_win_len: u16,
+    remote_has_sack: bool,
+    /// MSS we advertised in our SYN-ACK (cached for retransmit on
+    /// duplicate SYN).
+    our_mss: u16,
+    /// Creation timestamp (for expiry).
+    created_at: Instant,
+}
```
whitequark: I think you can make the half-open array smaller by implementing `Default`? I have no strong feelings about this one way or another.

src/socket/tcp_listener.rs
```diff
@@ -0,0 +1,468 @@
...
+///     tcp.accept(pending).unwrap();
+///     // tcp is now ESTABLISHED
+/// }
+/// ```
+pub struct Socket<'a> {
+    listen_endpoint: IpListenEndpoint,
+    /// Queue of half-open (SYN_RECEIVED) connections.
+    syn_queue: &'a mut [Option<HalfOpen>],
+    /// Queue of completed connections waiting to be accepted.
+    accept_queue: &'a mut [Option<PendingConnection>],
```
whitequark: Would it be possible to unify the queues? This seems like a UX improvement. The options in them are nearly the same anyways, and I think unifying them would give some more flexibility in terms of capacity.

whitequark:
I've done a first pass of a review, looking for potential UX issues for the interface and glaring problems. (I didn't find any of the latter.) I haven't yet reviwed the TCP state machine logic closely.

src/socket/tcp_listener.rs
```diff
@@ -0,0 +1,577 @@
...
+impl<'a> Socket<'a> {
+    /// Create a new TCP listen socket.
+    ///
+    /// * `syn_queue` – storage for half-open connections (SYN_RECEIVED).
+    ///   Its length limits the number of concurrent in-progress handshakes.
+    /// * `accept_queue` – storage for completed connections waiting to be
+    ///   accepted.  Its length is the backlog.
+    pub fn new(
+        syn_queue: &'a mut [Option<HalfOpen>],
+        accept_queue: &'a mut [Option<PendingConnection>],
```
whitequark: These should not be bare `&mut slices`. With the current design of smoltcp, these should be `managed::Slice`s (and if the two are unified this still holds).

Dirbaio:
I think this can be simplified more. The current design does
- Receive SYN
- Send SYN+ACK
- Receive ACK
- User accepts the connection and sets up a TcpSocket for it
- TcpSocket handles sending/receiving data

An alternative design would be:
- Receive SYN, queue it
- User accepts the connection and sets up a TcpSocket for it
- TcpSocket handles sending SYN+ACK, receiving ACK same as without using TcpListener.
- TcpSocket handles sending/receiving data

Advantages:

- Simpler, less code duplication with the state machine in TcpSocket. Only one queue.
- The first ACK can already carry data. How's that handled currently? It seems to me it's just dropped and it relies on the peer retransmitting it before it does land in the TcpSocket buffer? This alternative design handles it without retransmissions

Also, can you add an example? It's useful both for users as documentation, and for PR review so we can see how the end-user API looks like in practice.

src/socket/mod.rs
sunhaosheng: Done

src/socket/tcp_listener.rs
sunhaosheng: Done. I unified the two queues into a single `backlog: ManagedSlice<'a, Option<BacklogEntry>>` where `BacklogEntry` is a `HalfOpen | Completed` enum. `process_ack` now promotes entries in-place.

src/socket/tcp_listener.rs
sunhaosheng: Done, the unified backlog now uses `ManagedSlice<'a, Option<BacklogEntry>>`.

src/socket/tcp_listener.rs
sunhaosheng: Thanks for the suggestion! This is now moot since `HalfOpen` was absorbed into the `BacklogEntry::HalfOpen { .. }` variant after unifying the two queues. As a side note, `IpAddress` doesn't implement `Default` (its variants are feature-gated), so a `Default` impl would have required `#[cfg]` guards anyway — another reason the unification worked out nicely here.

sunhaosheng:
> I think this can be simplified more. The current design does
> 
>     * Receive SYN
> 
>     * Send SYN+ACK
> 
>     * Receive ACK
> 
>     * User accepts the connection and sets up a TcpSocket for it
> 
>     * TcpSocket handles sending/receiving data
> 
> 
> An alternative design would be:
> 
>     * Receive SYN, queue it
> 
>     * User accepts the connection and sets up a TcpSocket for it
> 
>     * TcpSocket handles sending SYN+ACK, receiving ACK same as without using TcpListener.
> 
>     * TcpSocket handles sending/receiving data
> 
> 
> Advantages:
> 
>     * Simpler, less code duplication with the state machine in TcpSocket. Only one queue.
> 
>     * The first ACK can already carry data. How's that handled currently? It seems to me it's just dropped and it relies on the peer retransmitting it before it does land in the TcpSocket buffer? This alternative design handles it without retransmissions
> 
> 
> Also, can you add an example? It's useful both for users as documentation, and for PR review so we can see how the end-user API looks like in practice.

Thanks for the review and the nice suggestion. My initial implementation kept the server-side passive-open handshake inside `TcpListener`, and only handed an already-established connection over to `TcpSocket` on `accept()`. However, I agree with your points—it led to unnecessary state machine duplication, and the issue with dropping data on the third ACK is definitely a flaw in my previous approach.

Therefore, I've just pushed an update that implements your idea:

- Refactored `Listener`: It now solely queues incoming `SYN` packets and extracts their metadata without sending replies.
- Updated `TcpSocket::accept`: It now takes the queued connection from the listener, enters the `SynReceived` state, and natively handles sending the `SYN+ACK` and receiving the `ACK` (preserving any piggybacked data).
- Simplified the backlog: Removed the split between "half-open" and "completed" states; it's now just a single unified queue of pending SYNs.
- Added an example: Included tcp_listener.rs to demonstrate how a user can poll the listener and hand off queued connections to a pre-allocated pool of `TcpSocket`s.

yanshay:
Any updates on this? 
I also need this, currently have to create multiple buffers to avoid failures from parallel requests arriving from clients and low on memory on ESP32 device.
Would be happy to see this merged.

Dirbaio:
i'm going to close this as slop. 

@sunhaosheng it's clear you're just pasting my review comments into an LLM and pasting back the LLM response. Please do not do this, it is incredibly rude, wastes maintainer time and provides no value to the project.

PR #1135, opened 2026-04-01
ipv4: implement RFC 3514 evil bit

pxseu:
This PR implements [RFC 3514](https://datatracker.ietf.org/doc/html/rfc3514) ("The Security Flag in the IPv4 Header"), closing what I can only describe as a critical gap in smoltcp's security posture.

With this great addition, I believe users will finally be able to build significantly more secure and maintainable projects.

Happy April 1st

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ defmt = ["dep:defmt", "heapless/defmt"]
 
 "proto-ipv4" = []
 "proto-ipv4-fragmentation" = ["proto-ipv4", "_proto-fragmentation"]
+"proto-evil" = ["proto-ipv4"]
 "proto-dhcpv4" = ["proto-ipv4"]
 "proto-ipv6" = []
 "proto-ipv6-hbh" = ["proto-ipv6"]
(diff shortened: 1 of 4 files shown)


whitequark:
I think you missed a few `#[cfg(feature = "proto-evil")]` :p

pxseu:
Right, good catch. Also made a test that should make the codecov not so upset :)

whitequark (approved):

Dirbaio:
As the worldwide threat model evolves, evil-er actors might discover they can set the evil bit to false to evade the filters. Smoltcp strives for world-class security and safety as evidenced by its strict no-unsafe policy, so we should have a better filter than simply dropping packets with the evil bit set. I suggest feeding all packets through an LLM on ingress. If the LLM says the packet is evil, we drop it. This way updates for the ever-changing security landscape are just a prompt update away!

whitequark:
Personally, I propose a common middle ground: introducing a new IP option where you could place a cryptocurrency private key. If the packet seems a little too evil we grab that (as a bribe) and process it anyway.

Dirbaio:
With the advent of AI the evil bit is considered deprecated, you can just run a hexdump of your traffic through an LLM. Of course larger LLMs are better at detecting exploits so it's a tradeoff between security and destroying your bank account but hey, such is the price of progress.

PR #1136, opened 2026-04-03
fix: Prevent SACK overflow panic from SeqNumber i32-to-usize sign-extension

eddo26:
Casting SeqNumber `s.0` (`i32`) directly to `usize` via `as usize` sign-extends negative values on 64-bit platforms, producing near-`usize::MAX` offsets. When the assembler iterator adds these to buffer-relative positions, the result overflows and panics.

Without this fix, any remote peer that completes a TCP handshake with SACK Permitted and an ISN in the upper half of the 32-bit sequence space (~50% of all connections) can cause a crash.  In release builds without overflow checks, the wrap-around silently produces incorrect SACK ranges, degrading TCP performance.

Changes
- `src/socket/tcp.rs` -- Cast through `u32` before widening to `usize` (`s.0 as u32 as usize`) at both `iter_data` call sites in `ack_reply()`, preventing sign-extension.
- `src/storage/assembler.rs` -- Use `wrapping_add` for the offset addition in `AssemblerIter::next()`, which is correct for modular-2^32 TCP sequence number arithmetic and prevents overflow on 32-bit targets.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -4410,6 +4410,40 @@ mod test {
     }
 
     #[test]
+    fn test_established_sack_no_overflow_on_near_max_seqnumber() {
+        let mut s = socket_established();
+        s.remote_has_sack = true;
+        s.remote_seq_no = TcpSeqNumber(-4);
+        s.remote_last_ack = Some(TcpSeqNumber(-4));
+
+        // Send an out-of-order segment 10 bytes past the expected sequence,
+        // creating a 10-byte hole at the front of the assembler.
+        send!(
+            s,
+            TcpRepr {
+                seq_number: TcpSeqNumber(-4 + 10),
+                ack_number: Some(LOCAL_SEQ + 1),
+                payload: &b"AAAAAAAAAA"[..],
+                ..SEND_TEMPL
+            },
+            Some(TcpRepr {
+                seq_number: LOCAL_SEQ + 1,
+                ack_number: Some(TcpSeqNumber(-4)),
+                window_len: 64,
+                sack_ranges: [
+                    Some((
+                        (-4_i32 + 10) as u32,
+                        (-4_i32 + 20) as u32,
+                    )),
+                    None,
+                    None,
+                ],
+                ..RECV_TEMPL
+            })
+        );
+    }
+
+    #[test]
     fn test_established_sliding_window_recv() {
         let mut s = socket_established();
         // Update our scaling parameters for a TCP with a scaled buffer.
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -1478,7 +1478,7 @@ impl<'a> Socket<'a> {
             if let Some(last_seg_seq) = self.local_rx_last_seq.map(|s| s.0 as u32) {
                 reply_repr.sack_ranges[0] = self
                     .assembler
-                    .iter_data(reply_repr.ack_number.map(|s| s.0 as usize).unwrap_or(0))
+                    .iter_data(reply_repr.ack_number.map(|s| s.0 as u32 as usize).unwrap_or(0))
                     .map(|(left, right)| (left as u32, right as u32))
                     .find(|(left, right)| *left <= last_seg_seq && *right >= last_seg_seq);
             }
@@ -1493,7 +1493,7 @@ impl<'a> Socket<'a> {
                 // most quickly advance the acknowledgement number.
                 reply_repr.sack_ranges[0] = self
                     .assembler
-                    .iter_data(reply_repr.ack_number.map(|s| s.0 as usize).unwrap_or(0))
+                    .iter_data(reply_repr.ack_number.map(|s| s.0 as u32 as usize).unwrap_or(0))
                     .map(|(left, right)| (left as u32, right as u32))
                     .next();
             }
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -4431,10 +4431,7 @@ mod test {
                 ack_number: Some(TcpSeqNumber(-4)),
                 window_len: 64,
                 sack_ranges: [
-                    Some((
-                        (-4_i32 + 10) as u32,
-                        (-4_i32 + 20) as u32,
-                    )),
+                    Some(((-4_i32 + 10) as u32, (-4_i32 + 20) as u32,)),
                     None,
                     None,
                 ],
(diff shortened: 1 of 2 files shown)


Dirbaio:
Please add tests showcasing the crashes.

eddo26:
@Dirbaio 

Please refer to the example in this branch [here](https://github.com/eddo26/smoltcp/tree/example/sack-seqnum-overflow-crash)

To run the example:
`RUST_BACKTRACE=1 cargo run --example crash_sack_overflow"`

The example crash output is here:
```
=== SACK Integer Overflow Reproducer ===

Attacker ISN: 0xFFFFFFF0 (i32: -16)
After SYN:    remote_seq_no = 0xFFFFFFF1 (i32: -15)

[1] Injecting SYN (seq=0xFFFFFFF0) with SACK Permitted option...
       Server responded with SYN-ACK (server ISN=0x77368C50)
[2] Injecting ACK (seq=0xFFFFFFF1, ack=0x77368C51) to complete handshake...
       Socket state: Established
[3] Injecting out-of-order data (seq=0x00000055, 100 byte gap, 10 bytes payload)...

    This triggers ack_reply() -> iter_data() -> AssemblerIter::next()
    where the sign-extended offset causes the overflow.


thread 'main' (3054524) panicked at src/storage/assembler.rs:361:35:
attempt to add with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/7e46c5f6fb87f8cf4353e058479cef15d1d952b4/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
             at /rustc/7e46c5f6fb87f8cf4353e058479cef15d1d952b4/library/core/src/panicking.rs:80:14
   2: core::panicking::panic_const::panic_const_add_overflow
             at /rustc/7e46c5f6fb87f8cf4353e058479cef15d1d952b4/library/core/src/panicking.rs:175:17
   3: <smoltcp::storage::assembler::AssemblerIter as core::iter::traits::iterator::Iterator>::next
             at ./src/storage/assembler.rs:361:35
   4: <smoltcp::storage::assembler::AssemblerIter as core::iter::traits::iterator::Iterator>::try_fold::<(), core::iter::adapters::map::map_try_fold<(usize, usize), (u32, u32), (), core::ops::control_flow::ControlFlow<(u32, u32)>, <smoltcp::socket::tcp::Socket>::ack_reply::{closure#3}, core::iter::traits::iterator::Iterator::find::check<(u32, u32), <smoltcp::socket::tcp::Socket>::ack_reply::{closure#4}>::{closure#0}>::{closure#0}, core::ops::control_flow::ControlFlow<(u32, u32)>>
             at /rustc/7e46c5f6fb87f8cf4353e058479cef15d1d952b4/library/core/src/iter/traits/iterator.rs:2508:34
   5: <core::iter::adapters::map::Map<smoltcp::storage::assembler::AssemblerIter, <smoltcp::socket::tcp::Socket>::ack_reply::{closure#3}> as core::iter::traits::iterator::Iterator>::try_fold::<(), core::iter::traits::iterator::Iterator::find::check<(u32, u32), <smoltcp::socket::tcp::Socket>::ack_reply::{closure#4}>::{closure#0}, core::ops::control_flow::ControlFlow<(u32, u32)>>
             at /rustc/7e46c5f6fb87f8cf4353e058479cef15d1d952b4/library/core/src/iter/adapters/map.rs:121:19
   6: <core::iter::adapters::map::Map<smoltcp::storage::assembler::AssemblerIter, <smoltcp::socket::tcp::Socket>::ack_reply::{closure#3}> as core::iter::traits::iterator::Iterator>::find::<<smoltcp::socket::tcp::Socket>::ack_reply::{closure#4}>
             at /rustc/7e46c5f6fb87f8cf4353e058479cef15d1d952b4/library/core/src/iter/traits/iterator.rs:2989:14
   7: <smoltcp::socket::tcp::Socket>::ack_reply
             at ./src/socket/tcp.rs:1483:22
   8: <smoltcp::socket::tcp::Socket>::process
             at ./src/socket/tcp.rs:2210:23
   9: <smoltcp::iface::interface::InterfaceInner>::process_tcp
             at ./src/iface/interface/tcp.rs:28:22
  10: <smoltcp::iface::interface::InterfaceInner>::process_ipv4
             at ./src/iface/interface/ipv4.rs:235:22
  11: <smoltcp::iface::interface::InterfaceInner>::process_ip
             at ./src/iface/interface/mod.rs:931:22
  12: <smoltcp::iface::interface::Interface>::socket_ingress::<crash_sack_overflow::CapturingDevice>::{closure#0}
             at ./src/iface/interface/mod.rs:662:30
  13: <crash_sack_overflow::InjectorRxToken as smoltcp::phy::RxToken>::consume::<smoltcp::iface::interface::PollIngressSingleResult, <smoltcp::iface::interface::Interface>::socket_ingress<crash_sack_overflow::CapturingDevice>::{closure#0}>
             at ./examples/crash_sack_overflow.rs:57:9
  14: <smoltcp::iface::interface::Interface>::socket_ingress::<crash_sack_overflow::CapturingDevice>
             at ./src/iface/interface/mod.rs:641:18
  15: <smoltcp::iface::interface::Interface>::poll::<crash_sack_overflow::CapturingDevice>
             at ./src/iface/interface/mod.rs:482:24
  16: crash_sack_overflow::main
             at ./examples/crash_sack_overflow.rs:304:11
  17: <fn() as core::ops::function::FnOnce<()>>::call_once
             at /rustc/7e46c5f6fb87f8cf4353e058479cef15d1d952b4/library/core/src/ops/function.rs:250:5
```

Dirbaio:
I mean add tests to the smoltcp codebase that fail with the bug present and pass with the fix applied. This way we have the guarantee that the bug won't be reintroduced in the future in e.g. a refactor.

eddo26:
@Dirbaio should be good to go now, let me know if you need anything else

Dirbaio:
Thanks for the tests!

The Assembler is supposed to deal only with offsets and lengths, never with sequence numbers modulo 2^32. It's logically a sequence of "hole, data, hole, data..." where the start of the first hole lines up with the start of the rx buffer (ie the next byte the application is supposed to recv). The numbers here can't get larger than 2^32 as long as the tcp socket buffer size doesn't exceed 2^32 which is a reasonable assumption.

The assembler `iter_data` takes a `first_offset: usize` to add to the returned tuples. The problem is it's *supposed* to be an offset, but the TCP code passes a sequence number, expecting the assembler to do sequence number math, instead of integer math.

IMO a cleaner solution is to remove `first_offset`, and let the TCP code do the addition *with sequence number math*. This keeps the assembler free of sequence number math.

did that in https://github.com/smoltcp-rs/smoltcp/pull/1138

PR #1140, opened 2026-04-08
iface: make neighbor discovery silent time configurable

lasiotus:
At the moment ARP requests are throttled to one/sec. In many situations this is too slow.

Allow the "discovery silent time" delay to be configurable per interface, with the default remaining at 1000ms.

### DIFF
--- a/src/iface/interface/mod.rs
+++ b/src/iface/interface/mod.rs
@@ -155,6 +155,8 @@ pub struct InterfaceInner {
     routes: Routes,
     #[cfg(feature = "multicast")]
     multicast: multicast::State,
+
+    discovery_silent_time: Duration,
 }
 
 /// Configuration structure used for creating a network interface.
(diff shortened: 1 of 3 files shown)


whitequark (approved):

Dirbaio:
I would prefer things to "just work" if possible instead of adding a configuration knob. 

Maybe we can ratelimit ARP to 1 per second per socket, or per destination IP, instead of 1 per second for the whole interface? I think this is what Linux does, it ARPs multiple IPs in parallel.

whitequark:
1 per second per socket is doable but won't help for some UDP workloads. 1 per IP would require a new data structure, but maybe that could be a fixed width one by default.

whitequark:
Here's an idea: what if we embed the ARP timestamp into the socket for TCP and into the packet buffer for UDP &c?

Dirbaio:
Closing since the approach is not ideal and OP is not responding.

For anyone reading this (including OP): we'd accept a PR that makes the ratelimit more granular.

PR #1142, opened 2026-04-16
Avoid immediate repoll loops when no transmit tokens are available

bugadani:
If the device's buffer is full, smoltcp requests an immediate re-poll hoping the buffer clears fast enough. This causes an issue: if the device cannot hand out tokens because the network link is down (cable is unplugged, wi-fi access point is not in range, etc.), smoltcp will eat up all CPU time by polling constantly. This PR adds a small delay to prevent busy-looping forever, if an immediate re-poll can't make progress.

### DIFF
--- a/src/iface/interface/ipv4.rs
+++ b/src/iface/interface/ipv4.rs
@@ -18,11 +18,14 @@ impl Interface {
         }
 
         let pkt = &self.fragmenter;
-        if pkt.packet_len > pkt.sent_bytes
-            && let Some(tx_token) = device.transmit(self.inner.now)
-        {
-            self.inner
-                .dispatch_ipv4_frag(tx_token, &mut self.fragmenter);
+        if pkt.packet_len > pkt.sent_bytes {
+            if let Some(tx_token) = device.transmit(self.inner.now) {
+                self.inner
+                    .dispatch_ipv4_frag(tx_token, &mut self.fragmenter);
+                self.inner.clear_device_exhausted();
+            } else {
+                self.inner.mark_device_exhausted();
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


Dirbaio:
I'm not sure if this is the best fix. If tx is exhausted we ideally want to retry immediately when it gets un-exhausted, not wait an arbitrary amount of time (10ms) which adds latency.

Smoltcp's phy trait doesn't have a mechanism for notifying un-exhaustion neither for rx nor tx. I'd say it's the user's responsibility to handle this when they write their own main loop. They can track tx exhaustion and block polling until tx gets unexhausted. Something like:

- Have a flag `tx_exhausted`.
- if smoltcp tries to tx and the driver is exhausted, set `tx_exhausted=true`.
- Normally poll every time one of these happens:
   - rx'd packet
   - after touching a socket
   - smoltcp poll_at timer expires
- except if `tx_exhausted=true`, poll every time one of these happens:
   - rx'd packet
   - after touching a socket
- When the driver reports tx is unexhausted, clear `tx_exhausted` and poll once.

The problem with this is it's somewhat complex, it's not great that every user has to implement this. OTOH smoltcp intentionally doesn't want to own the main loop, and I don't see how this can be fixed in an easier way if we don't own the main loop.

bugadani:
I've put this back to draft because I'm not even sure this fixes our issue, or if our analysis is even correct...

Dirbaio:
The bug might be in embassy-net, because it doesn't do this "ignore poll_at timer if tx exhausted" thing.

whitequark:
> OTOH smoltcp intentionally doesn't want to own the main loop, and I don't see how this can be fixed in an easier way if we don't own the main loop.

Make smoltcp pass through a future returned by the phy? I'm not sure that would be usable, nor that there's a better way.

PR #1148, opened 2026-05-05
 feat(phy): add TCP segmentation support

cagatay-y:
The Windows and Linux documentations for the same feature were used as references. It was more difficult to find information on how devices support the feature, so the interface is mostly based on how virtio-net devices expect to be driven. The code was [tested](https://github.com/hermit-os/kernel/pull/2308) on the Hermit kernel with a virtio-net device, but only with IPv4, as Hermit does not support IPv6.

## Design rationale
- The segmentation capabilities are specified separately for TCP on IPv4 and IPv6. This follows the example of [Linux](https://docs.kernel.org/networking/netdev-features.html#:~:text=TCPv4%20%28when%20NETIF%5FF%5FTSO%20is%20enabled%29%20or%20TCPv6%20%28NETIF%5FF%5FTSO6%29%2E) and [Windows](https://learn.microsoft.com/en-us/windows-hardware/drivers/network/offloading-the-segmentation-of-large-tcp-packets#:~:text=Support%20IPv4%20or%20IPv6%2C%20or%20both%20IPv4%20and%20IPv6.). This should provide flexibility for devices that support only one of the IP versions and should not be a problem for devices that support both as a single feature.
- As it is necessary to communicate to the device what the target segment size is, a field is added to [PacketMeta](https://docs.rs/smoltcp/latest/smoltcp/phy/struct.PacketMeta.html). To keep the compiler able to optimise out the structure if the features associated with the fields, the segmentation offload feature is optional, following the example of  [`packetmeta-id`](https://docs.rs/smoltcp/latest/smoltcp/phy/struct.PacketMeta.html#:~:text=If%20no%20field%20is%20enabled%2C%20this%20struct%20becomes%20zero%2Dsized%2C%20which%20allows%20the%20compiler%20to%20optimize%20it%20out%20as%20if%20the%20packet%20metadata%20mechanism%20didn%E2%80%99t%20exist%20at%20all.).
- For pre-segmentation IP packets whose size exceeds the maximum that is allowed by the length field of the IP header, we fall back to zero. An alternative would be to set the field to zero unconditionally, and this [seems](https://www.intel.com/content/dam/doc/manual/pci-pci-x-family-gbe-controllers-software-dev-manual.pdf#page=73) to be a "should" level expectation for at least for some Intel cards. The fallback approach has the advantage of keeping code that uses the IP header field (for example for [constructing `TcpPacket`s](https://docs.rs/smoltcp/latest/smoltcp/wire/struct.TcpPacket.html#method.new_checked) to set the partial checksum in our case) functional for sufficiently small unsegmented packets. On the other hand, it can also be argued that causing a failure even for smaller packets can help uncover the error case with the larger packets earlier in the development of the device drivers.

## Comparison to  #830
- The segmentation capabilities struct in the older PR follows the structure of [`ChecksumCapabilities`](https://docs.rs/smoltcp/latest/smoltcp/phy/struct.ChecksumCapabilities.html). This PR does not do so for the following reasons:
  - The segmentation capability for a given protocol is not binary. Different devices may support large (i.e. unsegmented) packets of different maximum sizes ([references for various devices](https://github.com/hermit-os/kernel/issues/1861#:~:text=Different,10%5D%2E))
  - On the receive side the related feature seems to be called "coalescing." It may be somewhat confusing to represent information for the receive side in a structure whose name refers to "segmentation."
- The capability structure is named `SegmentationCapabilities` rather than the TCP specific name `TsoCapabilities`. I believe the current structure can be expanded to support UDP, though I currently do not have a prototype for that.
- The current PR has support for communicating the target segment size to the device.
- This PR is missing an equivalent to [this line](https://github.com/smoltcp-rs/smoltcp/pull/830/changes#diff-9797218dceaf8c0bdfeddfab917f17b8d1028fd1e817a5b3de18bc4814fd894eR1742). The addition of the `set_meta` calls seems correct to me but because it is not segmentation offload specific and I was not able to test its necessity because of the lack of IPv6 support on our test platform, I did not include it in this PR.

### DIFF
--- a/src/phy/mod.rs
+++ b/src/phy/mod.rs
@@ -89,6 +89,8 @@ impl<'a> phy::TxToken for StmPhyTxToken<'a> {
 )]
 
 use crate::time::Instant;
+#[cfg(feature = "segmentation-offload")]
+use core::num::{NonZeroU16, NonZeroUsize};
 
 #[cfg(all(
     any(feature = "phy-raw_socket", feature = "phy-tuntap_interface"),
@@ -147,7 +149,7 @@ pub const IPV4_FRAGMENT_PAYLOAD_ALIGNMENT: usize = 8;
 /// struct becomes zero-sized, which allows the compiler to optimize it out as if
 /// the packet metadata mechanism didn't exist at all.
 ///
-/// Currently only UDP sockets allow setting/retrieving packet metadata. The metadata
+/// Currently only TCP and UDP sockets allow setting/retrieving packet metadata. The metadata
 /// for packets emitted with other sockets will be all default values.
 ///
 /// This struct is marked as `#[non_exhaustive]`. This means it is not possible to
@@ -168,6 +170,8 @@ pub const IPV4_FRAGMENT_PAYLOAD_ALIGNMENT: usize = 8;
 pub struct PacketMeta {
     #[cfg(feature = "packetmeta-id")]
     pub id: u32,
+    #[cfg(feature = "segmentation-offload")]
+    pub segmentation_offload_size: Option<NonZeroU16>,
 }
 
 /// A description of checksum behavior for a particular protocol.
@@ -233,6 +237,28 @@ impl ChecksumCapabilities {
     }
 }
 
+/// The maximum buffer size for a particular protocol or protocol pair that
+/// can be offloaded to the device for segmentation, or [None] if segmentation
+/// offload is not supported.
+///
+/// For Ethernet devices, this includes the Ethernet header (14 octets), but
+/// *not* the Ethernet FCS (4 octets).
+///
+/// If the device supports unsegmented IP packets with (depending on the IP
+/// version, total or payload) lengths greater than [u16::MAX], it should not
+/// rely on the length field in the IP header, as the actual length cannot be
+/// represented there. The value will be 0 instead.
+#[derive(Debug, Clone, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+#[cfg(feature = "segmentation-offload")]
+pub struct SegmentationCapabilities {
+    #[cfg(all(feature = "socket-tcp", feature = "proto-ipv4"))]
+    pub tcpv4: Option<NonZeroUsize>,
+    #[cfg(all(feature = "socket-tcp", feature = "proto-ipv6"))]
+    pub tcpv6: Option<NonZeroUsize>,
+}
+
 /// A description of device capabilities.
 ///
 /// Higher-level protocols may achieve higher throughput or lower latency if they consider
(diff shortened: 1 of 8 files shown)


src/socket/tcp.rs
```diff
@@ -2536,17 +2546,54 @@ impl<'a> Socket<'a> {
...
+                let device_limit = {
+                    let segmentation_caps = cx.segmentation_caps();
+                    match ip_repr.version() {
+                        #[cfg(feature = "proto-ipv4")]
+                        crate::wire::IpVersion::Ipv4 => segmentation_caps.tcpv4,
+                        #[cfg(feature = "proto-ipv6")]
+                        crate::wire::IpVersion::Ipv6 => segmentation_caps.tcpv6,
+                    }
+                    .map(|buf_size| {
+                        #[cfg(feature = "medium-ethernet")]
```
Dirbaio: checking `feature = "medium-ethernet"` here is not enough, you have to check the actual medium in the phy capabilities.

It's possible to enable `medium-ethernet` and `medium-ip` at the same time, so even if `medium-ethernet` is enabled you may have an IP medium phy.

src/phy/mod.rs
```diff
@@ -168,6 +170,8 @@ pub const IPV4_FRAGMENT_PAYLOAD_ALIGNMENT: usize = 8;
 pub struct PacketMeta {
     #[cfg(feature = "packetmeta-id")]
     pub id: u32,
+    #[cfg(feature = "segmentation-offload")]
+    pub segmentation_offload_size: Option<NonZeroU16>,
```
Dirbaio, line 183: Can you document what this does?

src/socket/tcp.rs
```diff
@@ -2536,17 +2546,54 @@ impl<'a> Socket<'a> {
...
+                        let ip_mtu = buf_size.get() - crate::wire::ETHERNET_HEADER_LEN;
+                        #[cfg(not(feature = "medium-ethernet"))]
+                        let ip_mtu = buf_size.get();
+                        ip_mtu - ip_repr.header_len() - TCP_HEADER_LEN
+                    })
+                    .unwrap_or(segment_size)
+                };
```
Dirbaio: can you add tests for the logic here? e.g. that if segmentation offload is enabled oversized packets are emitted up to the limit supported by the phy, with the right MSS value in the meta.

cagatay-y: I have added a test and a new testing device but I could not find any TCP tests to use as an example that use the network device, so the test does not actually use the testing device implementation. I am not sure if it is better to have an untested testing device or replace the method bodies with `unimplemented!()`.

mkroening:
@whitequark, would you mind taking a look at this? :)

Dirbaio:
@cagatay-y since you explicitly requested a review from me: i'm no longer maintaining smoltcp, i'm maintaining a fork here https://github.com/embassy-rs/xarxa . This PR would be welcome there, it looks good to me now after the last round of changes.

whitequark:
@mkroening I'll do that later today.

whitequark (approved):
Looks good, thank you!

mkroening:
Awesome, thanks! :)

Do you have any idea when we can expect a new release? Are there any specific things you want to get in before publishing a new one?

whitequark:
I'll take a look tonight once it's below 30C ambient.

whitequark:
Sorry for the delay, the temperature only just got bearable. I've released v0.14.0, enjoy!

PR #1150, opened 2026-05-19
Enqueue PacketBuffer metadata before closure call

KingCol13:
This is at least a small optimisation that avoids calling the closure to write the packet before we are sure that we can enqueue it.

I found it while debugging [this embassy issue](https://github.com/embassy-rs/embassy/issues/6143), as this change would avoid that issue. I think there's a chance that we could guarantee that `f` does not get called until we are sure that `enqueue_with_infallible` will succeed. In that case I should also document that and add a test.

### DIFF
--- a/src/storage/packet_buffer.rs
+++ b/src/storage/packet_buffer.rs
@@ -154,11 +154,14 @@ impl<'a, H> PacketBuffer<'a, H> {
             }
         }
 
+        let metadata_slot = self.metadata_ring.enqueue_one()?;
+
+        // Only call f once we know that we will succeed
         let (size, _) = self
             .payload_ring
             .enqueue_many_with(|data| (f(&mut data[..max_size]), ()));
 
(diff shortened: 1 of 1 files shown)


robamu:
would it be okay to update the function contract and guarantee that f is only called once if we pre-check like you suggested? Because otherwise, the calling embassy-net must assume f is called more than once, and the error handling needs to be updated

Dirbaio:
the "call multiple times" comes from a loop in embassy-net, not smoltcp.

The guarantee smoltcp should give is: enqueue either
- calls the closure and returns Ok
- doesn't call the closure and returns Err

(and the bug is there's today a case where it does call the closure and a later check fails which returns Err)

Changing the docs to explicitly guarantee this sounds good to me, even though this is arguably "common sense".

Dirbaio:
thanks!

PR #1151, opened 2026-05-25
Add DNS SRV/PTR request/response support

Qix-:
This PR adds DNS `SRV` and `PTR` record retrieval via DNS sockets with support for mDNS. This was a big painpoint in making mDNS-enabled devices under embassy; it'll require a code change to `embassy-net`, which I'm reviewing now and can PR if/when this is merged. It's already been tested on a real STM32.

This *is* a breaking change to users of the DNS subsystem, but not by too much.

In order to parse nontrivial responses (currently, smoltcp only resolved IPs from A/AAAA records) I turned the API into an iterator over the result. This was due to two reasons:

1. The response size was so large that Clippy rightfully complained about the enum variant size mismatches, and
2. Text-based answers (i.e. in PTR/SRV/CNAME responses) can have [compression](https://deepwiki.com/cilium/dns/2.3-domain-name-compression) markers _as well as_ literal `.` in labels (which is distinct from the typical FQDN notation `foo.bar`) which need to be preserved for full compliance with the spec, especially when chaining DNS requests (as one does with a PTR -> SRV chain for SD-style resolution). Parsing needed to be able to refer to the entire packet for resolving compressed labels, which under the old API was very memory intensive.

Thus, the API got a bit of a makeover, and I hope that's okay. The returned result is thus an iterator over an otherwise unparsed response. Responses are stored as 512 byte blobs that are held onto and parsed iteratively instead of shoving them all into a compile-time-specified `Vec`.

While not a _specific_ goal of the API, an unintended benefit is that now callers can decide the "maximum response count" per-callsite, instead of defining it globally, by `.collect()`-ing into a container of their choice (e.g. `heapless::Vec<_, 4>`).

The 512 byte blob size is really the only number that makes sense as per the spec given that most of the DNS handling is unicast resolution without EDNS support or TCP fallback, which as far as I understand dictates 512 byte maximum payload sizes. In the future, if EDNS/TCP fallback are implemented, a new config feature to change the size can probably be introduced (I added a comment in-situ mentioning this).

Further, I had to update a bit of the invalid packet logic since the current implementation assumed a point-to-point legacy implementation of mDNS for trivial A/AAAA resolution, which allows for a transaction ID and expected a question to be echoed in the response. mDNS does not dictate as far as I researched, and is typically not the case with non-A/-AAAA resolution, so the packet rejection logic had to be updated a bit.

If someone smarter than I knows better about some of these changes please let me know as I have had to deep dive into DNS internals for a few weeks to get this far. I knew DNS was hairy, but not this hairy.

Also welcome to feedback about the API changes.


src/socket/dns.rs
```diff
@@ -127,7 +134,21 @@ pub enum MulticastDns {
...
+pub type SrvQueryResult = WireSrvRecord<String<253>>;
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[cfg_attr(feature = "proto-dns-srv", expect(clippy::large_enum_variant))]
+/// A DNS answer returned by [`Socket::get_query_result`].
+pub enum QueryResult {
+    Address(IpAddress),
+    #[cfg(feature = "proto-dns-srv")]
+    Srv(SrvQueryResult),
```
mkj: `#[cfg(feature)]` on public enum variants tends to be awkward. If a library crate is written ignoring `proto-dns-srv` feature, it'll break if someone tries to use it in an application that enables `proto-dns-srv`.

Would it make more sense to have a separate `get_query_result_srv()`?

Qix-: Eh yeah, that's a good point, though breaking it out into separate methods is kind of painful too given that a request can (and often does) have multiple questions in it, so naming one `_srv` doesn't really make much sense.

I also want to add mDNS multi-response support in a later PR (performing a DNS request to the multicast group and waiting for a time period for multiple responses and accumulating them). Doing multiple disparate requests for each question rather than batching them will make it difficult to reason about discovered services on the network.

Would it be sufficient instead to mark the enum as `#[non_exhaustive]`? Then if a library that doesn't have it enabled uses the enum it'll be forced to handle `Srv` if the feature _is_ ever enabled?

Qix-: Or, alternatively, if there's another way to sort of "box" the SRV record instead of having it inline (not sure what the best practice would be here, maybe a statically allocated array + passing in a reference to the array as part of the SRV question) then the feature can go away entirely, I suppose, as the enum wouldn't be large.

Or maybe instead of an array reference it's a `&mut heapless::Vec<SrvRecord>` where it can push records onto the vec and store just the index on the returned enum variant.

Open to suggestions.

Qix-:
I'm going to refactor this a bit to try to remove the need for the feature flag (since I've learned a bit more about DNS formats in the meantime) and add PTR requests since they're also necessary for dns-sd style resolution to take place.

Mstrodl:
Thanks for doing this! Hoping to use it to discover mDNS services in my project.

Qix-:
Alright this is finally ready. @mkj made me go back and realize that the first attempt at this wasn't enough to handle e.g. compressed labels, which under the old everything-is-owned API was causing a huge memory inflation, so I had to do a bit more of an invasive change. Welcome to feedback.

PR #1153, opened 2026-05-26
multiflow netsim test

JPDye:
Created a multi-flow netsim test and refactored the existing netsim test to run using the same framework. Existing netsim results are unchanged.

Can set per-flow RTT, loss, bandwidth and can simulate a bottleneck link with a fixed bandwidth and buffer size (causing tail drop loss). This is important for congestion control changes coming soon (#1149) and SACK implementation after that.


JPDye:
Changed the starting values for the netsim. Previous values were capping congestion control algorithms at around `cwnd = 2` which didn't let the algorithms really show what they can do. Have maybe gone the other way now - getting some silly numbers in the CC vs no CC tests (+5000%...). 

Chime in if there's better starting values to use.

PR #1161, opened 2026-06-08
tcp: fix bug with effective MSS not considering the length of the TCP options

JPDye:
When timestamps are enabled, every data segment contains a 12 byte option that consumes space the payload would have otherwise taken up. The effective-mss/payload-size calculations (min of remote MSS and local MSS) did not previously take this into account.

My fix simply subtracts the options header length from whatever is calculated as the effective MSS. In `dispatch()` this uses the available `TcpRepr::header_len()`. In `seq_to_transmit` I branch based on the enabling of timestamping which, in the long run when other options are added, is a route for divergence and bugs to be introduced. That's a problem for another PR though.

As for why this bug hasn't previously been found, I assume it's because common ways of using `smoltcp` (`embassy-net`, `tokio-smoltcp` and `netstack-smoltcp`) don't enable timestamps by default _and_ don't provide a way to enable them. I only discovered this issue after forking `tokio-smoltcp` to allow me to set the timestamps.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -551,6 +551,15 @@ pub struct Socket<'a> {
 
 const DEFAULT_MSS: usize = 536;
 
+/// Minimum MSS we accept from the remote, same value as Linux's `TCP_MIN_SND_MSS`.
+/// Without it, a peer advertising a tiny MSS could force segments to carry little
+/// or no payload once the TCP options length is subtracted from the effective MSS,
+/// stalling the connection in an endless stream of empty segments.
+///
+/// Must exceed the maximum possible length of the TCP options (currently 12, for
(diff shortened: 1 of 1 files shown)


Dirbaio:
Fix looks good. 

There's one detail: if effective MSS becomes 0 smoltcp enters an infinite loop of sending empty packets. Before this PR it was possible only if the remote set MSS=0 in the SYN, which was disallowed by 43697943a8c7ec103d536280a7f17ed03e942b0d . but with this PR it can happen again if remote MSS is in 1..12. 

I've pushed a commit that fixes it by copying what Linux does (set a `remote_mss` floor of 48).

JPDye:
> There's one detail... 

Thanks for catching that!

PR #1162, opened 2026-06-11
fix: don't rate-limit ACKs for out-of-window data segments.

venezia:
If our ACK for a data segment is lost, the remote retransmits the segment, which is now behind our receive window. Same story for zero window probes. Per RFC 9293 §3.5.2 these should get an ACK back, but the challenge ACK rate limit drops all but one reply per second, so a retransmitting remote gets silence, backs off its RTO, retransmits, gets silence again... On a lossy link this turns a brief stall into a transfer that crawls at one exchange per second instead of recovering. Seen in a Raspberry Pi Pico 2W (cyw43) pulling images over WiFi with occasional ACK loss.  It seemed to happen in conjunction with istio/envoy, but I think the fix works for other situations as well.

Only data-carrying segments are exempted: the rate limit exists to break infinite ACK loops between desynced peers (f6a7380), and a data segment can't be part of such a loop, since the remote only sends it from its retransmission/persist timer, never in direct response to an ACK of ours. Pure ACK/RST/SYN handling is unchanged.


Dirbaio:
Looks great! thanks

there's 2 details i'll follow up in a next PR:

- RFC 9293 3.10.7.4 says out-of-window RSTs should be silently dropped. Before this PR smoltcp would send a challenge ACK (was already wrong) but this PR changes this to an unratelimited ACK (so arguably "more wrong"). 
- This PR also removes the ratelimits for SYNs with data. i'm not 100% sure this is correct, will check.

PR #1163, opened 2026-06-11
prevent underflow to tcp sequence numbers

calderonth:
TCP sequence numbers are 32-bit unsigned integers that naturally wrap around. Direct subtraction ( self.remote_last_seq - self.local_seq_no ) fails when sequence numbers wrap, triggering overflows or panic.
  The fix uses  .wrapping_sub()  on the raw sequence values ( .0 ) to safely perform wrapping arithmetic (modulo 2³²) conforming to RFC 793, and then casts it to usize for buffer offset calculations.
  This diff directly prevents:

  1. Panics or connection hangs caused by underflowing the adjusted window calculations.
  2. Buffer indexing errors or panics when sequence numbers wrap around during large transfers.

### DIFF
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -2549,7 +2549,8 @@ impl<'a> Socket<'a> {
                     .min(self.remote_mss)
                     .min(cx.ip_mtu() - ip_repr.header_len() - TCP_HEADER_LEN);
 
-                let offset = (self.remote_last_seq.0.wrapping_sub(self.local_seq_no.0)) as u32 as usize;
+                let offset =
+                    (self.remote_last_seq.0.wrapping_sub(self.local_seq_no.0)) as u32 as usize;
                 repr.payload = self.tx_buffer.get_allocated(offset, size);
 
                 // If we've sent everything we had in the buffer, follow it with the PSH or FIN
--- a/src/socket/tcp.rs
+++ b/src/socket/tcp.rs
@@ -763,7 +763,12 @@ impl<'a> Socket<'a> {
         let next_ack = self.remote_seq_no + self.rx_buffer.len();
 
         let last_win = (self.remote_last_win as usize) << self.remote_win_shift;
-        let last_win_adjusted = last_ack + last_win - next_ack;
+        let last_win_end = last_ack + last_win;
+        let last_win_adjusted = if next_ack > last_win_end {
+            0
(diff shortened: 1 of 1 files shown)


src/socket/tcp.rs
```diff
@@ -763,7 +763,12 @@ impl<'a> Socket<'a> {
         let next_ack = self.remote_seq_no + self.rx_buffer.len();
 
         let last_win = (self.remote_last_win as usize) << self.remote_win_shift;
-        let last_win_adjusted = last_ack + last_win - next_ack;
+        let last_win_end = last_ack + last_win;
```
Dirbaio, line 766: The crash here is impossible to hit AFAICT. The  socket state has the invariant `next_ack <= remote_last_ack + (remote_last_win << shift)` ie we never ACK things beyond the edge of the window we advertise (incoming data is truncated in rx if out of window)

The test that shows the crash is mucking with TCP socket internals directly. If you want to prove this is a bug you'll have to construct a test that reproduces it without touching the socket internals.

src/socket/tcp.rs
```diff
@@ -2544,7 +2549,8 @@ impl<'a> Socket<'a> {
                     .min(self.remote_mss)
                     .min(cx.ip_mtu() - ip_repr.header_len() - TCP_HEADER_LEN);
 
-                let offset = self.remote_last_seq - self.local_seq_no;
+                let offset =
+                    (self.remote_last_seq.0.wrapping_sub(self.local_seq_no.0)) as u32 as usize;
```
Dirbaio, line 2553: this is also an impossible state due to socket invariants. And not even covered by the test. 

Even worse, if the bad state was genuinely possible here this would change it from a panic (which we DO want, because it'd be a bug) into a bogus very high almost 2^32 offse.t

Dirbaio (changes requested):
> TCP sequence numbers are 32-bit unsigned integers that naturally wrap around. Direct subtraction ( self.remote_last_seq - self.local_seq_no ) fails when sequence numbers wrap, triggering overflows or panic.

This is not true. the `impl Sub for SeqNumber` handles this correctly. e.g. `SeqNumber(5) - SeqNumber(0xFFFF_FFF0)` yields 21, no panic)

Other than that, the logic fixes you propose are not fixes, they fix states that are impossible to hit due to socket invariants.

PR #1167, opened 2026-06-26
iface: add inbound packet drop callback

croepha:
Add `Interface::set_packet_drop_callback` to observe inbound packets that the interface does not deliver to a socket: both silent drops (malformed, not-for-us, no route, unhandled ICMP, ...) and well-formed packets it rejects with a reply (TCP RST, ICMP unreachable).

The callback receives a `PacketDropInfo` carrying the `PacketDropReason`, the `PacketDropDisposition` (dropped vs. replied), and — where parsed for free — the offending packet's IP header, transport ports and/or raw bytes (`PacketDropInfo::five_tuple`). An opaque `usize` cookie supplied with the callback is passed back on each invocation, so the capture-less `fn` callback can reach application state without an allocator and without any `unsafe` in the core iface.

Drops are reported across the ethernet, IP-demux, IPv4, IPv6, ARP, TCP, UDP, ICMP, IEEE 802.15.4 and 6LoWPAN receive paths. Malformed-packet drops go through a new `check_report!` macro (a self-aware sibling of `check!`). Reporting is a no-op when no callback is installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-No unreleased changes yet. Please send PRs!
+### Added
+
+- iface: Add an inbound packet drop callback. Use `Interface::set_packet_drop_callback` to be notified whenever the interface does not deliver an inbound packet to a socket — both silent drops (malformed, not-for-us, no route, unhandled ICMP, etc.) and well-formed packets that are rejected with a reply (TCP RST, ICMP unreachable). The callback receives a `PacketDropInfo` describing the `PacketDropReason`, the `PacketDropDisposition`, and, where they were parsed for free, the offending packet's IP header, transport ports and/or raw bytes (see `PacketDropInfo::five_tuple`). An opaque `usize` cookie supplied alongside the callback is passed back on every invocation, so the (capture-less) `fn` callback can reach application state without an allocator.
+
+
 
 ## [0.13.1] - 2026-05-01
 
--- a/src/iface/interface/drop.rs
+++ b/src/iface/interface/drop.rs
@@ -1,0 +1,326 @@
+//! Inbound packet drop / non-delivery notification.
+//!
+//! This module provides a mechanism for the user to observe what the interface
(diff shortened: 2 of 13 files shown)


Dirbaio:
What is the use case for this? It's a very invasive change touching every corner of packet handling.

whitequark:
Seconding what @Dirbaio says. The PR also appears to be LLM-generated.

croepha:
Thanks for the prompt engagement :) 

> What is the use case for this? It's a very invasive change touching every corner of packet handling.

I'm interested in ideas on how to make it less invasive... and I would hope that in the case where the feature wasn't being used it could be compiled out...

So, right now, if I want to log, or collect statistics about traffic that either gets dropped, or is only handled internally to   smoltcp, then I am basically forced to parse the packet myself.  Basically replicating the logic... having a callback seems like the best solution

whitequark:
You still haven't provided a motivation (*why* you are collecting this statistics). This isn't a feature commonly offered by network stacks besides a simple "N dropped packets" counter. For debugging you should already be able to use logs.

Certainly I don't want to be on the hook for maintaining a new public API with a detailed drop reason, so that's out of question.

croepha:
> Certainly I don't want to be on the hook for maintaining a new public API with a detailed drop reason, so that's out of question.

I get it, that's a fair point.  I suppose it is a bit overzealous...  Maybe we just need counters instead...  But I was thinking this would give the usage code, maximal flexibility... Which, to your point, isn't always the best option :-) 

> You still haven't provided a motivation (why you are collecting this statistics).

I'm interested in knowing where packets are coming from and why aren't they getting handled. Like, from a traffic accounting perspective.  Like we see X amount of data coming in but we only see Y coming out, and I'd like to know why... 

> For debugging you should already be able to use logs

Logging is much more expensive than having counters, you can leave counters on generally, in cases where you can't do logging...  

>  "N dropped packets" counters

Essentially yes...  I would probably push that slightly, and say that dropped bytes is also interesting, _and_ a common feature... _And_ by category, is probably the next step...

I'm not really interested in trying to litigate how every existing network stack does accounting, I'm not trying to sell you on it, if you don't want it, then we can drop it...

PR #1168, opened 2026-06-28
Minimum impl of RFC 4861 for RFC 6775 Work

aketchum15:
## RFC 4861 minimum compliance: multi-router default list, SLLAO capture, neighbor cache fix

This is groundwork for RFC 6775 (6LoWPAN Neighbor Discovery), which requires a node to send unicast Neighbor Solicitations carrying an Address Registration Option (ARO) to one or more router on its Default Router List. Three things were missing that blocked that work.

### Capture SLLAO from Router Advertisements into the neighbor cache (`src/iface/interface/ipv6.rs`)

When an RA includes a Source Link-Layer Address Option, the router's link-layer address is now immediately stored in the neighbor cache. Previously `lladdr` was discarded, forcing a broadcast NS/NA exchange before any off-link packet could be sent. On IEEE 802.15.4 multicast is broadcast, so this extra round-trip is especially costly. RFC 6775 ARO dispatch needs to reach the router directly without it.                  

### Multiple default routers in the route table (`src/iface/route.rs`)
`Routes::add_default_ipv6_route` previously called `remove_default_ipv6_route()` before inserting, evicting the existing default router. It now uses upsert semantics keyed on the gateway address: two different gateways coexist, re-adding an existing gateway refreshes it in-place. Adds `remove_default_ipv6_route_via(gateway)` to remove a specific router and `default_ipv6_routers()` to iterate over all live ones.               

### Expose `Interface::default_ipv6_routers()` (`src/iface/interface/mod.rs`)
Thin delegate to `Routes::default_ipv6_routers()`. RFC 6775 ARO dispatch will iterate this to know which routers to register with.

### Conditional neighbor cache flush in `update_ip_addrs` (`src/iface/interface/mod.rs`)
update_ip_addrs previously flushed the neighbor cache unconditionally, even when no addresses changed. SLAAC sync calls update_ip_addrs on every poll cycle, which was silently discarding the SLLAO-captured entries from the fix above. The flush is now gated on the address list actually changing, which preserves neighbor entries across no-op syncs while still flushing correctly when addresses are added or removed.

Minimal IPv4 behavior changes. All new and modified IPv6 functions are `#[cfg(feature = "proto-ipv6")]`. The update_ip_addrs change is strictly narrowing, the existing `test_arp_flush_after_update_ip` test was preserved unchanged and still passes.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/iface/interface/ipv6.rs
+++ b/src/iface/interface/ipv6.rs
@@ -513,7 +513,7 @@ impl InterfaceInner {
                 router_lifetime,
                 reachable_time: _,
                 retrans_time: _,
-                lladdr: _,
+                lladdr,
                 mtu: _,
                 prefix_info,
             } if self.slaac_enabled => {
@@ -522,6 +522,17 @@ impl InterfaceInner {
                         || ip_repr.dst_addr.is_link_local())
                     && ip_repr.hop_limit == 255
                 {
+                    // Capture the router's link-layer address so unicast NS (e.g. for RFC 6775
+                    // ARO) can be sent without a prior broadcast solicitation.
+                    #[cfg(any(feature = "medium-ethernet", feature = "medium-ieee802154"))]
+                    if let Some(lladdr) = lladdr {
+                        if let Ok(hw_addr) = lladdr.parse(self.caps.medium) {
+                            if hw_addr.is_unicast() && ip_repr.src_addr.x_is_unicast() {
+                                self.neighbor_cache
+                                    .fill(ip_repr.src_addr.into(), hw_addr, self.now);
+                            }
+                        }
+                    }
                     self.slaac.process_advertisement(
                         &ip_repr.src_addr,
                         router_lifetime,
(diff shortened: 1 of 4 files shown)


whitequark:
I am strongly opposed to merging any code that was generated with Claude.

aketchum15:
@whitequark I am really only interested in working on RFC 6775 implementation. I did perform a manual code review on all these changes, and tried to limit the scope as much as possible to minimize AI generated code. Is your opposition related to the model as implied by your comment? If I were to go back and manually implement this now, it would look very similar. I completely respect your decision here, on a side note, perhaps adding a contributing.md clarifying LLM usage would be helpful. Thanks!

whitequark:
> perhaps adding a contributing.md clarifying LLM usage would be helpful. Thanks!

I agree! I just added one: https://github.com/smoltcp-rs/smoltcp/pull/1169. This is actually significantly more permissive than my usual approach to LLM use, and the particular policy text is chosen to allow some of the existing contributors to continue using LLMs for code review as well as for vulnerability research.

aketchum15:
Okay, thank you. I will rewrite this PR manually I guess.

whitequark:
Thanks! I will put it higher in my review queue, unless someone else beats me to it.

PR #1171, opened 2026-07-01
Fixing warnings when buidling under a minimal features set.

cirvladimir:
This change is a noop. It only fixes warnings that get generated under a minimal features set, because for example a variable is declared or imported always, but only used under some feature sets.

The warnings can be reproduced on the main branch with this command:

`cargo check --no-default-features --features std,medium-ethernet,proto-ipv4,socket-tcp,socket-dhcpv4`

It also  explicitly declares when imports/variables are only used under certain features.

These are the warnings it fixes:
warning: unused import: `super::fragmentation::PacketAssemblerSet`
  --> smoltcp\src\iface\interface\mod.rs:35:5
   |
35 | use super::fragmentation::PacketAssemblerSet;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default

warning: unused imports: `IFACE_MAX_PREFIX_COUNT` and `IFACE_MAX_SIXLOWPAN_ADDRESS_CONTEXT_COUNT`
  --> smoltcp\src\iface\interface\mod.rs:42:27
   |
42 |     IFACE_MAX_ADDR_COUNT, IFACE_MAX_PREFIX_COUNT, IFACE_MAX_SIXLOWPAN_ADDRESS_CONTEXT_COUNT,
   |                           ^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: variable does not need to be mutable
   --> smoltcp\src\iface\interface\ipv4.rs:103:13
    |
103 |         let mut ipv4_repr = check!(Ipv4Repr::parse(ipv4_packet, &self.caps.checksum));
    |             ----^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` (part of `#[warn(unused)]`) on by default

warning: unused variable: `meta`
  --> smoltcp\src\iface\interface\ipv4.rs:98:9
   |
98 |         meta: PacketMeta,
   |         ^^^^ help: if this is intentional, prefix it with an underscore: `_meta`
   |
   = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default

warning: unused variable: `frag`
   --> smoltcp\src\iface\interface\ipv4.rs:101:9
    |
101 |         frag: &'a mut FragmentsBuffer,
    |         ^^^^ help: if this is intentional, prefix it with an underscore: `_frag`

warning: unused variable: `ip_repr`
   --> smoltcp\src\iface\interface\ipv4.rs:323:9
    |
323 |         ip_repr: Ipv4Repr,
    |         ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_ip_repr`

warning: unused variable: `fragmenter`
    --> smoltcp\src\iface\interface\mod.rs:1127:9
     |
1127 |         fragmenter: &mut Fragmenter,
     |         ^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_fragmenter`

warning: unreachable pattern
    --> smoltcp\src\iface\interface\mod.rs:1324:13
     |
1318 |             Medium::Ethernet => {
     |             ---------------- matches all the relevant values
...
1324 |             _ => (EthernetAddress([0; 6]), tx_token),
     |             ^ no value can reach this
     |
     = note: `#[warn(unreachable_patterns)]` (part of `#[warn(unused)]`) on by default

warning: unreachable pattern
    --> smoltcp\src\iface\interface\mod.rs:1321:21
     |
1320 |                     (HardwareAddress::Ethernet(addr), tx_token) => (addr, tx_token),
     |                     ------------------------------------------- matches all the relevant values
1321 |                     (_, _) => unreachable!(),
     |                     ^^^^^^ no value can reach this

warning: unused variable: `repr`
    --> smoltcp\src\iface\interface\mod.rs:1356:26
     |
1356 |             IpRepr::Ipv4(repr) => {
     |                          ^^^^ help: if this is intentional, prefix it with an underscore: `_repr`

warning: unreachable `else` clause
   --> smoltcp\src\socket\dhcpv4.rs:332:75
    |
332 |         let HardwareAddress::Ethernet(ethernet_addr) = cx.hardware_addr() else {
    |         ----------------------------------------------------------------- ^^^^
    |         |
    |         assigning to binding pattern will always succeed
    |
    = note: this pattern always matches, so the else clause is unreachable
    = note: `#[warn(irrefutable_let_patterns)]` on by default

warning: unreachable `else` clause
   --> smoltcp\src\socket\dhcpv4.rs:569:75
    |
569 |         let HardwareAddress::Ethernet(ethernet_addr) = cx.hardware_addr() else {
    |         ----------------------------------------------------------------- ^^^^
    |         |
    |         assigning to binding pattern will always succeed
    |
    = note: this pattern always matches, so the else clause is unreachable

warning: method `emit_header` is never used
   --> smoltcp\src\wire\udp.rs:275:19
    |
232 | impl Repr {
    | --------- method in this implementation
...
275 |     pub(crate) fn emit_header<T>(&self, packet: &mut Packet<&mut T>, payload_len: usize)
    |                   ^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: `smoltcp` (lib) generated 13 warnings (run `cargo fix --lib -p smoltcp` to apply 8 suggestions)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,8 @@ name: Test
 jobs:
   tests:
     runs-on: ubuntu-22.04
-    needs: [check-msrv, test-msrv, test-stable, clippy, test-netsim]
+    needs:
+      [check-msrv, test-msrv, test-stable, clippy, check-unused, test-netsim]
     steps:
       - name: Done
         run: exit 0
@@ -32,6 +33,13 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run Clippy
         run: ./ci.sh clippy
+
+  check-unused:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check for unused code
+        run: ./ci.sh check_unused
 
   test-stable:
     runs-on: ubuntu-22.04
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ license = "0BSD"
 autoexamples = false
 
 [lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(fuzzing)'] }
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(fuzzing)', 'cfg(ci_forbid_unused)'] }
 
 [dependencies]
 managed = { version = "0.8", default-features = false, features = ["map"] }
--- a/ci.sh
+++ b/ci.sh
@@ -81,6 +81,11 @@ clippy() {
     cargo +$MSRV clippy --tests --examples -- -D warnings
 }
 
+check_unused() {
+    export RUSTFLAGS="--cfg ci_forbid_unused --deny unused --deny irrefutable_let_patterns"
+    cargo check --tests --examples
+}
+
 build_16bit() {
     rustup toolchain install nightly
     rustup +nightly component add rust-src
@@ -128,6 +133,10 @@ fi
 
 if [[ $1 == "clippy" || $1 == "all" ]]; then
     clippy
+fi
+
+if [[ $1 == "check_unused" || $1 == "all" ]]; then
+    check_unused
 fi
 
 if [[ $1 == "build_16bit" || $1 == "all" ]]; then
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,8 @@
 #![cfg_attr(not(any(test, feature = "std")), no_std)]
 #![deny(unsafe_code)]
+// Since we have many different flags that can come on and off, we expect some variables to
+// be unused sometimes, and we allow that.
+#![cfg_attr(not(ci_forbid_unused), allow(unused, irrefutable_let_patterns))]
 
 //! The _smoltcp_ library is built in a layered structure, with the layers corresponding
 //! to the levels of API abstraction. Only the highest layers would be used by a typical


cirvladimir:
@whitequark , this PR has been sitting for a while, just wondering if there's any steps I missed in submitting a PR or if something else is off, let me know, thanks.

whitequark:
Sorry, the email has gotten lost in my inbox. I'm not sure this is a great approach. It adds a lot of visual noise in the source code, is hard to maintain, and it's not clear what the advantages are over just disabling the warning. 

Philosophically, warnings are useful because following them results in better code. I don't think in this case we are seeing better code.

cirvladimir:
Thanks for the reply. The commit does add visual noise, which is not great.

Would you prefer a bunch of `#[allow(unused_variables)]` instead of my `#[cfg]` macros everywhere? That would also add a bunch of visual noise, though maybe less.

My only goal is to get rid of compiler warnings when compiling my code - I'm trying to create a production system and I don't want any compiler warnings. I also don't want to disable all these warnings globally as I think they're actually useful in many cases. If you have any other approach where I can both disable the features and also compile without warnings that would work for me.

whitequark:
I'd be fine adding `#![allow(unused)]` by default, having a cfg flag to override it, and enabling that flag with `RUSTFLAGS="--deny warnings"` in one of the CI jobs.

cirvladimir:
Just to make sure I understan you correctly - I was suggesting adding `#![allow(unused)]` and other allows to every place where the warning is triggered. I'm not sure I quite understand what you meant by overriding it with a cfg flag.

Also I would do it on a case by case basis for each warning.

For example, the first warning I fixed is this:

```
-#[cfg(any(feature = "proto-ipv4", feature = "proto-sixlowpan"))]
+#[cfg(feature = "_proto-fragmentation")]
 use super::fragmentation::PacketAssemblerSet;
```

I think that's actually the correct fix - whether we use PacketAssemblerSet depends on `_proto-fragmentation` in that file, not the other flags above.

One fix was

```
-use crate::config::{
-    IFACE_MAX_ADDR_COUNT, IFACE_MAX_PREFIX_COUNT, IFACE_MAX_SIXLOWPAN_ADDRESS_CONTEXT_COUNT,
-};
+use crate::config::IFACE_MAX_ADDR_COUNT;
+#[cfg(feature = "proto-sixlowpan")]
+use crate::config::IFACE_MAX_SIXLOWPAN_ADDRESS_CONTEXT_COUNT;
```

Alternatively I could have done.

```
+#[allow(unused_imports)]
 use crate::config::{
     IFACE_MAX_ADDR_COUNT, IFACE_MAX_PREFIX_COUNT, IFACE_MAX_SIXLOWPAN_ADDRESS_CONTEXT_COUNT,
 };
```

I'm not sure which version is better. The second one is shorter/less lines, but the first one self-documents why we're doing the `allow`, if we ever remove the usage it will actually tell us to remove it with a warning.

One fix was

```
         dst_addr: &IpAddress,
-        fragmenter: &mut Fragmenter,
+        #[cfg(feature = "proto-ipv6")] fragmenter: &mut Fragmenter,
+        #[cfg(not(feature = "proto-ipv6"))] _fragmenter: &mut Fragmenter,
     ) -> Result<(HardwareAddress, Tx), DispatchError>
```

This one is very verbose, so for this one the `allow` version might be better, though it does trade less lines for less protections from warning.


```
         dst_addr: &IpAddress,
+      // Allowing an unused variable because fragmenter is unused under certain feature sets.
+      #[allow(unused)]
         fragmenter: &mut Fragmenter,
```

Or did you have something else in mind, like disabling the warning for the whole file or completely? I would not recommend that because warnings on unused variables generally make code cleaner, if someone erases the variable usage in the future the warning protects us from forgetting to also erase the variable declaration.

whitequark:
> Or did you have something else in mind, like disabling the warning for the whole file or completely?

For the whole file.

> I would not recommend that because warnings on unused variables generally make code cleaner, if someone erases the variable usage in the future the warning protects us from forgetting to also erase the variable declaration.

That's what the cfg flag is for. Actually, I think this could be implemented in a slightly simpler way: add `#![allow(unused)]` and add a CI job with `--forbid unused` (which if I recall overrides the attribute, where `--deny unused` would not).

Also, this PR is a great illustration of how blindly following the rules does *not* make the code cleaner, so it's not really a good justification.

cirvladimir:
I followed your suggestion of a global allow unused, can you check it here before I switch it over to my current fix?

https://github.com/smoltcp-rs/smoltcp/compare/main...cirvladimir:smoltcp_pub:allow-unused-global

This basically disables all unused checks during development in the whole library, they'll only run during CI. I think that has a few downsides - you might miss some stuff until you run CI - but I do see it's quite a bit simpler and more global and still protected by CI.

Let me know if you think it's a good move for the library and I'll swap it into this PR.

Doing "--forbid unused" gives a bunch of these warnings - so it didn't work: "allow(unused) incompatible with previous forbid"

whitequark:
Thanks for doing the work! I'm happy with your impl.

cirvladimir:
Cool, updated the PR. Didn't check in CI but I did check locally adding an unused variable and running `ci.sh check_unused` successfully fails.

PR #1173, opened 2026-07-08
Fix broken doc links

nzec:
A few of the doc links were broken likely due to reorganization.
This commit fixes them


whitequark (approved):
Thanks!

PR #1175, opened 2026-07-11
Implement core::error::Error for errors

orsinium:
The `Error` trait is now available in the `core` and doesn't require `std`: [core::error::Error](https://doc.rust-lang.org/beta/core/error/trait.Error.html)


whitequark:
Thanks. It looks like core::error::Error is available since 1.81.0, so this doesn't affect our MSRV.

PR #1177, opened 2026-07-14
fix: incorrect conversion from `std::time::Instant` to `Instant`

### LINKED ISSUE #347: The impl From<std::Instant> for smoltcp::Instant seems buggy
https://github.com/smoltcp-rs/smoltcp/blob/master/src/time.rs#L71

While reviewing the crate a bit, I noticed this implementation, which seems a bit suspicious to me. It creates a `smoltcp::time::Instant` from an `std::time::Instant` by looking at how much time has elapsed since said instant was created? This seems super wrong to me.

`smoltcp::time::Instant` is supposed to represent the number of milliseconds passed since an arbitrary point in time (e.g. be monotically increasing, with a frequency of 1000Hz). Calling `smoltcp::Instant::from(std::Instant)` twice with the same argument should always return the same value. This cannot be done with `elapsed`, since it returns the amount of time since the previous instant. See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1311ed777b61850c4ccf82642cd190a6)

I expected `From<Instant>` to be implemented along the lines of:

```rust

#[cfg(feature = "std")]
impl From<::std::time::Instant> for Instant {
    fn from(other: ::std::time::Instant) -> Instant {
        lazy_static! {
            static ref REFERENTIAL: Instant = Instant::now();
        }
        let elapsed = other - *REFERENTIAL;
        Instant::from_millis((elapsed.as_secs() * 1_000) as i64 + (elapsed.subsec_nanos() / 1_000_000) as i64)
    }
}
```

frederic-arr:
Fixes #347.

Converting the same `std::time::Instant` twice produced two different values because the `From` implementation was incorrect.

This PR fixes it by comparing against a single lazily-initialized reference `std::time::Instant`. All conversions are now measured relative to that fixed point instead of "now" and converting the same `std::time::Instant` yields the same result. I've also added a test that fails with the previous implementation and works with the fixed one.

## Notes

- I'm using [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) which is available since Rust 1.80.0 and fits within the MSRV.
- Since the reference `std::time::Instant` gets lazily initialized, it is possible that on the first use, the `std::time::Instant` from which we try to converts happens to be earlier than the reference. In this case, the conversion will be saturated to 0.
- The saturating behavior is the default for [`std::time::Instant::duration_since`](https://doc.rust-lang.org/std/time/struct.Instant.html#method.duration_since) but not a guarantee, so I've used [`std::time::Instant::saturating_duration_since`](https://doc.rust-lang.org/std/time/struct.Instant.html#method.saturating_duration_since) instead.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -105,8 +105,11 @@ impl Instant {
 #[cfg(feature = "std")]
 impl From<::std::time::Instant> for Instant {
     fn from(other: ::std::time::Instant) -> Instant {
-        let elapsed = other.elapsed();
-        Instant::from_micros((elapsed.as_secs() * 1_000000) as i64 + elapsed.subsec_micros() as i64)
+        static REFERENTIAL: ::std::sync::LazyLock<::std::time::Instant> =
+            ::std::sync::LazyLock::new(::std::time::Instant::now);
+
+        let n = other.saturating_duration_since(*REFERENTIAL);
+        Self::from_micros(n.as_secs() as i64 * 1000000 + n.subsec_micros() as i64)
     }
 }
 
--- a/src/time.rs
+++ b/src/time.rs
@@ -397,6 +397,21 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "std")]
+    fn test_instant_conversions_from_std_instant() {
+        let std_now = ::std::time::Instant::now();
+
+        let before = Instant::from(std_now);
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Thanks!

PR #1187, opened 2026-07-24
wire/rpl: avoid option_length underflow when emitting PadN

### LINKED ISSUE #1180: `wire::rpl` option emit: `option_length` underflow on PadN(254/255)
**Component:** `smoltcp::wire::rpl` (feature `proto-rpl`) · **Commit:** `764ef3a`

`RplOptionRepr::emit` computes `let mut option_length = self.buffer_len() as u8;` then, for non-`Pad1` options, `option_length -= 2;`. For `Repr::PadN(size)`, `buffer_len()` is `2 + size`; a `size` of 254/255 makes it 256/257, which `as u8` truncates to 0/1, and the `- 2` underflows. `parse` copies the raw length byte into `PadN(size)`, so a parsed-then-re-emitted PadN option carries `size ∈ {254,255}` into `emit`: a panic under overflow-checks, a corrupt `option_length` on the wire in release.

- emit: https://github.com/smoltcp-rs/smoltcp/blob/764ef3a8cc38d543f407555772f495d4810c8895/src/wire/rpl.rs#L2114-L2123
- taint source: https://github.com/smoltcp-rs/smoltcp/blob/764ef3a8cc38d543f407555772f495d4810c8895/src/wire/rpl.rs#L2040

**Reproduce:** parse `[0x01, 0xFE]` → `PadN(254)`, then `emit` it, panics `attempt to subtract with overflow` (debug) / writes a wrong length.

**Fix:** subtract the 2-byte header in `usize` before the `as u8` cast.

kirk-baird:
Resolves #1180

### DIFF
--- a/src/wire/rpl.rs
+++ b/src/wire/rpl.rs
@@ -2172,13 +2172,14 @@ pub mod options {
         }
 
         pub fn emit<T: AsRef<[u8]> + AsMut<[u8]> + ?Sized>(&self, packet: &mut Packet<&'p mut T>) {
-            let mut option_length = self.buffer_len() as u8;
+            // Subtract the header in usize before the cast to avoid PadN(254/255) underflow.
+            let mut option_length = self.buffer_len();
 
             packet.set_option_type(self.into());
 
             if !matches!(self, Repr::Pad1) {
                 option_length -= 2;
-                packet.set_option_length(option_length);
+                packet.set_option_length(option_length as u8);
             }
 
             match self {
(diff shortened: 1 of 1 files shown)


whitequark (approved):

whitequark:
Could you resolve the merge conflict, please?

whitequark:
Thanks!

kirk-baird:
I put the tests in the same spot on all 4 branches, should of seen this coming :sweat_smile:

whitequark (approved):

PR #1189, opened 2026-07-25
meta: add extra field

xoviat:
Adding this field will allow me to store a timestamp in packetmeta for rx packets, avoiding extra complexity in embassy-net. The complexity added to smoltcp is virtually zero.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,7 @@ defmt = ["dep:defmt", "heapless/defmt"]
 "socket-tcp-reno" = []
 
 "packetmeta-id" = []
+"packetmeta-extra" = []
 
 "async" = []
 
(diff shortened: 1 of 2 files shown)


adamgreig:
Would this update https://github.com/embassy-rs/embassy/pull/6325 so that instead of storing an `id` in metadata and using it to look up a 2x`u32` timestamp, you just store the 2x`u32` directly in the metadata?

xoviat:
The implementation of ptp in embassy-stm32 has way too much generic code. I am adding traits to make this generic over drivers and to reduce the complexity of driver implementation.

> you just store the 2xu32 directly in the metadata?

For the rx packets, yes. The tx packets will be handled by other code but for the rx packets, this makes more sense because we're already storing this information and I just need another u32 to store the timestamp.

PR #1193, opened 2026-08-03
tcp: make RTTE_MIN_RTO and RTTE_INITIAL_RTO configurable at build time

### LINKED ISSUE #1190: TCP: make RTTE_MIN_RTO configurable — the 0.13 change to 1s makes loss recovery ~100x slower on low-RTT links
## Summary

smoltcp 0.13 changed the RTO clamp in socket/tcp.rs from [10 ms, 10 s] (0.12: RTTE_MIN_RTO = 10, RTTE_MAX_RTO = 10000) to RFC 6298's [1 s, 60 s] RTTE_MIN_RTO = 1000, RTTE_MAX_RTO = 60_000, initial RTO 1 s). That is RFC-conformant (6298 §2.4 "SHOULD be at least 1 second"), but on the microsecond-RTT links where smoltcp typically lives — virtio, TAP, local segments, point-to-point radios — it means every timeout-recovered loss now stalls the connection for a full second where 0.12 recovered in ~10 ms. There is currently no knob: the constant is not in build.rs's CONFIGS mechanism and there is no socket-level setter.

## Why timeout recovery matters more than it may look

A tail-drop of the entire in-flight window — the common loss shape when a bulk sender overruns a queue — cannot be recovered by fast retransmit: with no later data arriving at the receiver, the sender never sees the third duplicate ACK. In that case RTO is the only recovery path, so the minimum RTO directly sets the cost of every such loss.

## Measurement

Real capture (Linux TAP, kernel peer, smoltcp 0.13.1, ~µs RTT): a 64 MiB bulk send hit one kernel-side burst drop of the 16 KiB in-flight window. The receiver dup-ACKed twice, then both sides went silent; the retransmission fired exactly 1.0000 s later, after which the remaining 62 MiB moved at ~615 MiB/s. Net effect: 55.7 MiB/s effective for the transfer, ~11x below the loss-free rate, from a single lost burst. Under 0.12's 10 ms floor the same event cost ~10 ms.

For calibration: mainstream stacks do not use the RFC's 1 s in practice — Linux's TCP_RTO_MIN is 200 ms (and is itself configurable per-route with rto_min).

## Ask

Could RTTE_MIN_RTO (and ideally RTTE_INITIAL_RTO) join the existing build-time tunables in build.rs's CONFIGS table, overridable via the SMOLTCP_* env vars / features like IFACE_NEIGHBOR_CACHE_COUNT already is? A runtime socket setter (precedent: set_ack_delay) would be even nicer but the build-time knob alone would solve it. The default staying at the RFC-conformant 1 s is fine.

Happy to send a PR for whichever shape the maintainers prefer.

AsafMesi:
Fixes #1190

## Motivation

The 0.13 change to RFC 6298's [1 s, 60 s] RTO range makes RTO-based loss recovery ~100x slower on low-RTT links (virtio, TAP, local segments). Tail-drop losses — where the entire in-flight window is lost — cannot recover via fast retransmit and depend entirely on the RTO timer, so every such event now costs a full second regardless of the actual path RTT. See #1190 for measurements (a single 16 KiB burst drop in a 64 MiB transfer cut effective throughput 11x). For comparison, Linux's `TCP_RTO_MIN` is 200 ms and is itself configurable per-route.

## Changes

- Adds `RTTE_MIN_RTO` and `RTTE_INITIAL_RTO` (both in milliseconds) to the existing `CONFIGS` build-time mechanism, following the same pattern as the other tunables:
  - Environment variables: `SMOLTCP_RTTE_MIN_RTO`, `SMOLTCP_RTTE_INITIAL_RTO` (any value)
  - Cargo features: `rtte-min-rto-*`, `rtte-initial-rto-*` (power-of-two values from 1 to 32768, plus the 1000 default)
- Defaults are unchanged and remain RFC 6298-conformant at 1000 ms — behavior is identical unless a user opts in.
- `gen_config.py` updated; `Cargo.toml` and `build.rs` sections regenerated with it.
- CHANGELOG entry added.

`RTTE_MAX_RTO` is intentionally left untouched to keep the scope minimal.

## Testing

- `cargo test`: 672 passed, 0 failed (plus doctests).
- Verified `SMOLTCP_RTTE_MIN_RTO=10 SMOLTCP_RTTE_INITIAL_RTO=50 cargo build` emits `RTTE_MIN_RTO = 10` / `RTTE_INITIAL_RTO = 50` in the generated `config.rs`.
- Verified `cargo build --features rtte-min-rto-16` emits `RTTE_MIN_RTO = 16`.


whitequark:
Please be aware that your PR is violating the project's contribution policy. https://github.com/smoltcp-rs/smoltcp/blob/main/CONTRIBUTING.txt#L8-L11

AsafMesi:
This PR is LLM-generated, so I’m closing it. The measurement work in #1190 is my own, and I’d like to finalize that design before I write a replacement.

whitequark:
Thank you!

Dirbaio:
@AsafMesi is a fork that doesn't ban LLMs https://github.com/embassy-rs/xarxa

Issue #256, opened 2018-07-22
Implement TCP selective acknowledgement extension (sACK)


jhwgh1968:
The TCP extension is described in [RFC 2018](https://tools.ietf.org/html/rfc2018).

I am opening a tracking issue for this since I didn't see one. If no one else wants to pick this up, I volunteer.

@whitequark previously made some offhanded comments suggesting it would be easy to send sACK records in acknowledgements based on current gaps tracked by the TCP reassembly code. I tend to agree, based on my understanding of the code.

The more difficult part is actually using information in sACK records sent by the remote TCP. While completely ignoring the remote's sACK records is legal -- the ACK field must not advance past the first gap to prevent mismatching state -- it defeats the extension's purpose, and so is highly discouraged.

whitequark:
> It seems when I write a straightforward case (do a bunch of `send!` calls with frames but one missing from the sequence), the `recv!` call at the end -- expecting an ACK with sACK in the TCP options -- gets `Ok(None)`, which causes an assert failure. And I cannot figure out why.

Where is your SACK code located? Is it just in `TcpSocket::process` or are you also emitting SACKs in `TcpSocket::dispatch`?

whitequark:
@jhwgh1968 You can't call `s.rx_buffer.dequeue*` if you want to continue using the socket afterwards. Take a look at the implementation of `TcpSocket::recv`; you're skipping `TcpSocket::recv_impl`, which means `s.remote_seq_no` doesn't get updated, and the assembler gets confused afterwards.

I think it would be ideal if we replaced all uses of `s.rx_buffer.dequeue*` in tests with `s.recv*`.

jhwgh1968:
I have updated my test branch, and replaced the ``dequeue`` calls with ``recv``. It seems to still return ``Ok(None)``.

What else am I missing?

whitequark:
Great to hear, and thanks for your work and patience!

jhwgh1968:
Since it didn't get linked to this issue, I have a phase 1 implementation in #266.

I will probably work on a phase 1.5 next, where it actually becomes fully "should" compliant with its receiving behavior.

jhwgh1968:
Unfortunately, I have not had time to work on this. If anyone else wants to take this issue, feel free. Otherwise, I may get to it later in the year.

jabedude:
@jhwgh1968 what is left to implement?

jhwgh1968:
[...]

PR #1195, opened 2026-08-07
tcp: sending SACK ranges

JPDye:
A partial implementation of selective acknowledgements (SACK), following on from the original SACK work done in #266. See issue #256 for discussion around that PR and suggestions for future approaches to the work.

Copying from that thread:

> On the sending side, SACK ranges are sent only in response to duplicate data (when process() calls ack_reply()) and only a single SACK range is sent. This range doesn't operate as the RFC suggests and is simply a "here's the earliest missing segment" hint.

> On the receiving side, all SACK ranges are discarded.

This PR implements full data-receiver behaviour ([RFC 2018 - Section 4](https://datatracker.ietf.org/doc/html/rfc2018#section-4)): transmitting SACK ranges on all emitted segments and making full use of the  available SACK blocks that can be sent. I've limited it to three blocks and it would be easy to support four SACK blocks, but I imagine this will nearly always be combined with the timestamp option so there's not much point.

The main two tests from the RFC are implemented, plus a few more checking edge cases and scenarios not discussed in the RFC.

Against Linux TCP through a TAP device, I see these results with SACK enabled and disabled:
```
         ┌────────────┬─────────────┐
         │     throughput delta     |
┌────────┬────────────┬─────────────┐
│ loss % │ 5 ms delay │ 10 ms delay │
├────────┼────────────┼─────────────┤
│ 0      │ +3%        │ +1%         │
├────────┼────────────┼─────────────┤
│ 0.001  │ −0.3%      │ −2%         │
├────────┼────────────┼─────────────┤
│ 0.01   │ −3%        │ −4%         │
├────────┼────────────┼─────────────┤
│ 0.1    │ +0.6%      │ +3%         │
├────────┼────────────┼─────────────┤
│ 1      │ −81%       │ −65%        │
├────────┼────────────┼─────────────┤
│ 2      │ −60%       │ −93%        │
├────────┼────────────┼─────────────┤
│ 5      │ −59%       │ −54%*       │
└────────┴────────────┴─────────────┘

*sack-disabled run timed out so true difference is greater
```

Overall, significant performance gains as loss is introduced. The same tests run against the smoltcp stack show 0 difference in throughput (due to the SACK ranges not being used) so there doesn't seem to be any regressions.

Next steps will be to implement the other half (receiving SACK ranges and acting on them). The work is done, just in need of tidying up. Once this is merged I'll follow up with that, plus DSACK and PLPMTUD too.


whitequark:
It looks like this PR is still a work in progress (there are some failing CI jobs, etc). Once it is ready, please add some explanatory text describing your changes. Thanks!

JPDye:
Didn't mean to make this PR until I got back to the office. It's mostly good now.

Wrote up some notes, fixed a few things. CI still fails in the multiflow-netsim. It's just meaning less noise from SACK ranges introducing extra bytes to be transmitted. Running over 50 different seeds shows no meaningful change in throughput (as I would expect since we never consume these ranges).


Any thoughts would be appreciated.

PR #1198, opened 2026-08-18
feat: gate MLD behind multicast feature

sigvartmh:
For embedded devices it's important to have things configurable and tunable to ensure that you don't use flash space for features you don't use.


whitequark (approved):
Thanks!

PR #1199, opened 2026-08-18
feat: MaybeUnint sockets to save space by putting sockets in .bss

sigvartmh:
### Proposal

So this may be a bit more controversial but sockets currently don't get the `00` bit patteren meaning they can't be in `.bss`  on embedded targets. Maybe there is a better way of not initializing them but this is the solution I found. Feel free to deny or come with feed back on it. This also adds `unsafe` to the code which may not be wanted.

## Commit message: 

Initialise a slice of uninitialised socket storage in place, returning the initialised slice.

This exists to keep socket storage out of `.data` on bare-metal targets.

`SocketStorage` is a niche-filled `Option<Item>`: `None` occupies zero bytes of payload and its bit pattern is chosen by the compiler, so `EMPTY` is *not* all-zero. A

```rust
static mut STORAGE: [SocketStorage; N] = [SocketStorage::EMPTY; N];
```

therefore cannot live in `.bss`; it goes in `.data` and the linker emits an initialiser image of `N * size_of::<SocketStorage>()` bytes into flash, almost all of which is zeros. Declaring the storage `MaybeUninit` instead places it in `.bss`, and this function fills it at startup:

```rust
static mut STORAGE: MaybeUninit<[SocketStorage; N]> = MaybeUninit::uninit();
let sockets = SocketSet::new(SocketStorage::init_slice(unsafe {
    &mut *(core::ptr::addr_of_mut!(STORAGE) as *mut [MaybeUninit<SocketStorage>; N])
}));
```

Measured on thumbv7em-none-eabihf with 7 sockets, `opt-level="z"` and fat LTO: 3616 bytes of `.data` removed, of which 3584 were zeros, at a cost of ~380 bytes of `.rodata` and a short init loop. RAM use is unchanged.

### DIFF
--- a/src/iface/socket_set.rs
+++ b/src/iface/socket_set.rs
@@ -1,4 +1,5 @@
 use core::fmt;
+use core::mem::MaybeUninit;
 use managed::ManagedSlice;
 
 use super::socket_meta::Meta;
@@ -15,6 +16,16 @@ pub struct SocketStorage<'a> {
 
 impl<'a> SocketStorage<'a> {
     pub const EMPTY: Self = Self { inner: None };
+
+    #[allow(unsafe_code)]
+    pub fn init_slice<'s>(storage: &'s mut [MaybeUninit<Self>]) -> &'s mut [Self] {
+        for slot in storage.iter_mut() {
+            slot.write(Self::EMPTY);
+        }
+        // every element of `storage` was initialised by the loop above,
+        // and `MaybeUninit<T>` is guaranteed to have the same layout as `T`.
+        unsafe { &mut *(storage as *mut [MaybeUninit<Self>] as *mut [Self]) }
+    }
 }
 
 /// An item of a socket set.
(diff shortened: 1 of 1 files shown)


whitequark:
I've designed to keep the core of smoltcp (excluding `phy` drivers) free of `unsafe` as an intentional tradeoff. I would like to keep that tradeoff intact. There are quite a few optimizations that can be done if you're OK writing unsafe code, but I think there is a place for a TCP/IP stack which chooses robustness over absolute highest performance (or code size, etc), so smoltcp is such a stack.

Regarding your particular problem: smoltcp has been originally designed to have its structures allocated on stack, perhaps in `main`. Would that solve your problem?

sigvartmh:
> I've designed to keep the core of smoltcp (excluding `phy` drivers) free of `unsafe` as an intentional tradeoff. I would like to keep that tradeoff intact. There are quite a few optimizations that can be done if you're OK writing unsafe code, but I think there is a place for a TCP/IP stack which chooses robustness over absolute highest performance (or code size, etc), so smoltcp is such a stack.
> 
> Regarding your particular problem: smoltcp has been originally designed to have its structures allocated on stack, perhaps in `main`. Would that solve your problem?

Sure if I self allocate them on the stack having thme being allocated/initialized from `.bss`  that's fine thanks for your reply :)

I looked through the code and found a no places of unsafe which is why I assumed this maybe wouldn't go in anyways great work on the stack thanks for making it public :+1:

PR #1203, opened 2026-09-05
feat: allow listen on ANY port 

flaneur2020:
I've been trying to build something like [tun2socks](https://github.com/xjasonlyu/tun2socks) (which based on gVisor's network stack) in Rust using smoltcp to route TUN traffic through a SOCKS proxy. 

However, I've run into a limitation: smoltcp does not yet support listening on any ports. In the context of a TUN device, we really need to be able to catch traffic on any port to forward it to the proxy.

I found @chayleaf's [PR #814](https://github.com/smoltcp-rs/smoltcp/pull/814) which was tackling this, but it hasn't been merged yet. I decide to build on top of the code from PR #814 and incorporated the comment from @raftario.

I think this functionality would be super useful for anyone building VPN-like tools. I've passed the tests in my tun2socks impl with this branch here: https://github.com/flaneur2020/smoltcp-socks/blob/master/src/netstack/actor.rs#L288

Any feedback would be appreciated!

### DIFF
--- a/src/wire/ip.rs
+++ b/src/wire/ip.rs
@@ -425,19 +425,32 @@ impl<T: Into<Address>> From<(T, u16)> for Endpoint {
 /// An internet endpoint address for listening.
 ///
 /// In contrast with [`Endpoint`], `ListenEndpoint` allows not specifying the address,
-/// in order to listen on a given port at all our addresses.
+/// in order to listen on a given port at all our addresses, and/or not specifying the
+/// port, in order to listen on all ports at a given address.
+///
+/// A `None` address means "any address", and a `None` port means "any port".
 ///
 /// An endpoint can be constructed from a port, in which case the address is unspecified.
 #[derive(Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
 pub struct ListenEndpoint {
     pub addr: Option<Address>,
-    pub port: u16,
+    pub port: Option<u16>,
 }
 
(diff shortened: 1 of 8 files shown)


whitequark:
I think it's a great feature! I'll give it a closer look during the workweek.

src/wire/ip.rs
```diff
@@ -466,39 +489,43 @@ impl From<::core::net::SocketAddrV6> for ListenEndpoint {
...
+            (None, None) => write!(f, "*:*"),
         }
     }
 }
 
 #[cfg(feature = "defmt")]
 impl defmt::Format for ListenEndpoint {
     fn format(&self, f: defmt::Formatter) {
-        defmt::write!(f, "{:?}:{=u16}", self.addr, self.port);
+        defmt::write!(f, "{:?}:{:?}", self.addr, self.port);
```
sigvartmh: This will print `Some(<num>)` in `defmt` won't it? 
It's not unreadable it will just be a bit different than before. So `<num>.<num>.<num>.<num>:Some(<num>)`  now I could be wrong so feel free to ignore it.

flaneur2020: yeah, it really doesn't look great. i think it would look better if it followed the same format as the `Display` above. i've fixed it in 0e05e3b, PTAL. thanks for pointing this out!

(though I suspect it was originally showing up as `Some(<num>.<num>.<num>.<num>):<num>`, since `self.addr` is also an `Option` 😲...)

PR #1207, opened 2026-09-19
iface: IPv6 fragment reassembly (RFC 8200, 5722, 6946, 7112)

254CARBON:
## The problem

`proto-ipv6-fragmentation` does not do anything, and it does not compile.

It turns on `_proto-fragmentation` and nothing reads it on the IPv6 receive path, so an IPv6 Fragment header falls through to the bottom of `process_nxt_hdr` and draws an "unrecognized next header" Parameter Problem. The datagram is never reassembled.

On `main` today, enabling it on its own is also a build error:

```console
$ cargo check --no-default-features --features \
    std,medium-ethernet,medium-ip,proto-ipv6,proto-ipv6-fragmentation,multicast,\
socket-raw,socket-udp,socket-tcp,socket-icmp,auto-icmp-echo-reply
error[E0433]: failed to resolve: use of undeclared type `PacketAssemblerSet`
```

`FragmentsBuffer`'s `assembler` field is gated on the `_proto-fragmentation` umbrella while its `use` is gated on `any(proto-ipv4, proto-sixlowpan)`, and `proto-ipv6-fragmentation` is the only way to turn the umbrella on without either of those. No row in `ci.sh` enables the feature, which is why this has gone unnoticed.

## What this does

Five commits. **0001 stands alone and is worth taking on its own merits even if the rest is unwanted.**

| | |
|---|---|
| **0001** | `PacketAssembler::{add, add_with}` called `Assembler::add` for its side effect and discarded its `Result`. `Assembler::add` returns `Err(TooManyHolesError)` once `ASSEMBLER_MAX_SEGMENT_COUNT` (default **4**) discontiguous ranges are recorded — so on refusal the fragment's octets were copied into the buffer and the range went unrecorded. `is_complete` compares against `peek_front`, which the unrecorded range never advances, so the datagram cannot complete: the slot is occupied until the reassembly timeout and the caller sees a silent timeout rather than an error. Affects IPv4 and 6LoWPAN too. |
| **0002** | RFC 8200 §4.5 reassembly, plus RFC 6946 atomic fragments and RFC 7112. Widens the `PacketAssemblerSet` import to the umbrella, and adds a `ci.sh` row that runs the path. |
| **0003** | RFC 5722 overlap. |
| **0004** | `Interface::set_reassembly_max_len`, and a slot's buffer is given back when it is reused. |
| **0005** | `reassembly_entries` / `reassembly_entry` / `reassembly_evict` / `reassembly_clear`. |

### Details worth a reviewer's eye

**Identity is `{src, dst, identification}` — not the next header.** `ipv4::Key` includes the protocol field; for IPv6 the upper-layer protocol lives in the Fragment header, every fragment carries its own copy, and keying on it would let a sender split one datagram across two slots by lying in a later fragment.

**Reassembly runs after the address checks**, where the IPv4 path's runs before them, because RFC 8200 reassembles at the destination — a datagram this interface is not the destination of should not take a slot. It runs before the raw-socket filter so a raw socket sees the datagram rather than its pieces.

**`Ipv6Reassembly` passes the assembler as a field borrow rather than `&mut FragmentsBuffer`.** The 6LoWPAN path decompresses into that struct's `decompress_buf` and hands the resulting slice to `process_ipv6`, so the slice holds an immutable borrow of one field for as long as the returned packet and there is no way to pass `&mut` to the whole struct alongside it. `Option<&mut …>` with `None` from 6LoWPAN was the cheaper option and would have been a silent capability gap. Happy to reshape this if you would rather split `FragmentsBuffer` instead.

**The first fragment's next header and hop limit live in `PacketAssembler`**, so the `reset` that frees a slot clears them; a side table would be one more thing to keep in step with expiry and eviction.

**Enforced:** atomic fragments bypass the assembler (RFC 6946), so they cannot collide with a genuinely fragmented datagram of the same identification; a non-final fragment that is not a multiple of 8 draws Parameter Problem code 0 at Payload Length; a fragment whose tail passes 65535 draws code 0 at its Fragment Offset; a first fragment missing its upper-layer header draws code 3 (new `ParamProblem::IncompleteHdrChain`, RFC 7112); two finals that disagree on the total reset the slot; none of these errors is sent to a multicast destination (RFC 4443 §2.4); and the reassembled header comes from the first fragment, not from whichever one completed it.

### Two judgement calls I would like your opinion on

1. **An exactly-repeated fragment counts as an overlap** (0003), so a network-duplicated fragment costs the datagram. The stricter reading: a same-offset, same-length repeat with different octets is precisely the rewrite RFC 5722 exists to stop, and telling it apart from a duplicate would mean comparing payloads. Both behaviours have a test, so relaxing it is a small change if you prefer that.
2. **RFC 7112 is checked, not walked** (0002). The check is that the upper-layer header the Fragment header *names* is present — the case the RFC was written for. Where the minimum is unknowable here (an extension header, an unknown protocol) the rule is not enforced rather than guessed at.

## Deliberately not included

**ICMPv6 Time Exceeded on reassembly timeout** (RFC 4443 §3.3). It needs an egress path — the expiring side has no transmit token — plus retention of the first fragment to quote, and it bears on neither correctness of delivery nor bounded memory. I would rather send it as its own PR, with its own case for "no first fragment was seen, so stay silent", than bundle it here.

## Testing

Nine tests over both media in `iface::interface::tests::ipv6::fragmentation`, plus assembler unit tests. The reassembly cases send a *checksummed* ICMPv6 echo request split across fragments and require the echo **reply**, so every octet has to land at its own offset for a case to pass.

- Every commit from 0002 on builds and passes under the new `ci.sh` row: 429 / 433 / 437 / 439. 0001 predates that row and passes under default features: 674, against 673 before it — the one new test.
- `default`, the new row, an `alloc` row and the 6LoWPAN row: all green (676 / 439 / 277 / 195 on the rebased branch).
- `cargo clippy --tests --examples -- -D warnings`: clean.
- `RUSTFLAGS="--cfg ci_forbid_unused --deny unused --deny irrefutable_let_patterns" cargo check --tests --examples`: clean.
- `cargo check --examples`: clean.

The overlap rule and the memory bound were each confirmed to fail their tests with the rule disabled, rather than assumed to be doing something.

MSRV 1.91.0; developed and tested on 1.91.1.

## Why I need it

We are building an IPv6 host stack on smoltcp and need reassembly under budgets the embedder owns — which is what 0004 and 0005 are for, and why they are last: the first three stand without them. Happy to split this into smaller PRs if that suits review better.


254CARBON:
Disclosure under this project's LLM policy (CONTRIBUTING → nlnetlabs.nl/llm-policy): the code in this pull request and its description were produced with an LLM (Anthropic's Claude) working for 254CARBON, and that should have been stated when it was opened. The policy does not accept LLM-generated code, so we are withdrawing it. Apologies for any review time it cost. This comment was also written with Claude, at the account owner's direction.

